PR #3, opened 2016-03-17
Restoring original behavior on Linux: use system zlib package by default

mfursov:
I found that the change I made about 1 month ago to switch Linux to use bundled zlib was not safe: UGENE has no images/icons on Fedora build when built with bundled zlib package because some incompatibilities with OS png library.

This fix reverts to the original behavior.
Checked on Linux  & Windows.


dkandrov:
Thanks!

PR #5, opened 2016-03-24
CMake build files for most of the plugins

mfursov:
CMake build files for most of the plugins
- Removed deprecated (Qt4) version of biostruct3d_view

Checked Windows & Linux builds


src/plugins/biostruct3d_view/biostruct3d_view.pri
```diff
@@ -1,20 +1,5 @@
 # include (biostruct3d_view.pri)
 
-defineTest( use_deprecated_view ) {
```
atiunov: Restore the deprecated view. It is still used for Qt versions less than 5.4. The minimum supported Qt version is 5.2.1.

atiunov:
Restore deprecated view in the biostruct3d_view plugin. It is still needed for Qt versions less than 5.4.

PR #8, opened 2016-03-30
Fixing search path for lupdate/lrelease

mfursov:
The only way to build UGENE today is to have all QT tools in path. However the good practice for QT projects is to rely on qmake environment. 

This patch fixes the way UGENE build looks for additional tools like lupdate and lrelease.

After this patch applied QT tools are detected from qmake environment and there is no use of $PATH. So a developer may have multiple QT versions in the same system and rely on qmake only to build UGENE.

Checked Windows & Linux builds


dkandrov:
Thanks :)

PR #11, opened 2016-04-13
fixed UGENE-4630 Opening of FASTQ file with invalid lines

vaskinyy:


src/corelibs/U2Formats/src/FastqFormat.cpp
```diff
@@ -357,16 +410,29 @@ Document* FastqFormat::loadDocument(IOAdapter* io, const U2DbiRef& dbiRef, const
...
+    load(io, dbiRef, _hints, io->getURL(), objects, os, gapSize, predictedSize, lockReason, skippedLines);
+    if (skippedLines.size() > 0){
+        QMapIterator<QString, QString> i(skippedLines);
+        while (i.hasNext()) {
+            i.next();
+            QString msg = i.key() + ": " + i.value();
+            if (objects.length() > 0){
+                os.addWarning(msg);
+            }else{
+                os.setError(msg);
```
ggrekhov: Only the last message is stored as an error.

src/corelibs/U2Formats/src/FastqFormat.cpp
```diff
@@ -478,30 +544,30 @@ DNASequence *FastqFormat::loadSequence(IOAdapter* io, U2OpStatus& os) {
     QString sequenceName = readSequenceName(os, io, '@');
     // check for eof while trying to read another FASTQ block
     CHECK_EXT(!io->isEof(), , NULL);
-    CHECK_OP(os, NULL);
+    CHECK_OP(os, new DNASequence());
```
ggrekhov, line 551: Why do you want to return "new DNASequence()" instead of NULL?
There is a code that expects exactly NULL (StreamSequenceReader::hasNext()). What should we do with it?

src/corelibs/U2Formats/src/FastqFormat.cpp
```diff
@@ -357,16 +410,29 @@ Document* FastqFormat::loadDocument(IOAdapter* io, const U2DbiRef& dbiRef, const
...
     QVariantMap hints = _hints;
     QList<GObject*> objects;
+    QMap<QString, QString> skippedLines;
 
     int gapSize = qBound(-1, DocumentFormatUtils::getMergeGap(_hints), 1000*1000);
     int predictedSize = qMax(100*1000, DocumentFormatUtils::getMergedSize(hints, gapSize==-1 ? 0 : io->left()));
 
     QString lockReason;
-    load(io, dbiRef, _hints, io->getURL(), objects, os, gapSize, predictedSize, lockReason);
+    load(io, dbiRef, _hints, io->getURL(), objects, os, gapSize, predictedSize, lockReason, skippedLines);
```
ggrekhov, line 419: I would wrap it into a new method (processSkippedLines(), processWarnings() etc.).

src/corelibs/U2Formats/src/FastqFormat.cpp
```diff
@@ -211,11 +212,22 @@ static void readQuality(U2OpStatus& os, IOAdapter *io, QByteArray &sequence, int
     }
 }
 
+#define SKIPPED_LINES_ERRORS_LIMIT 50
+static bool error_logging_break(U2OpStatus& os, QMap<QString, QString>& skippedLines, const QString& seqName){
```
ggrekhov: Why do you use under_score instead of camelCase here? Is there a reason?

ggrekhov:
Could you, please, fix the described code issues?

src/corelibs/U2Formats/src/FastqFormat.cpp
vaskinyy: Too much Python.
Fixed

src/corelibs/U2Formats/src/FastqFormat.cpp
vaskinyy, line 419: Looks good this way

src/corelibs/U2Formats/src/FastqFormat.cpp
vaskinyy, line 551: Alignment stops with an error if it returns NULL, and finishes correctly with new DNASequence()

src/corelibs/U2Formats/src/FastqFormat.cpp
vaskinyy: Right. Fixed

PR #19, opened 2016-09-18
Note about maximum version of Qt

AlekseyDurachenko:
I can't compile the source code with Qt >= 5.6 because QtWebKit are dropped. Read more: http://blog.qt.io/blog/2016/03/16/qt-5-6-released/

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ### Prerequisites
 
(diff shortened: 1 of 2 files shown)


epushkova:
Yes, thank you.
We are aware of that, but getting rid of Qt WebKit is not a priority task for now.

PR #22, opened 2016-10-27
MSA Editor support for high DPI displays

ggrekhov:
Before:
before

After:
after

### DIFF
--- a/src/corelibs/U2View/src/ov_msa/MSAEditor.cpp
+++ b/src/corelibs/U2View/src/ov_msa/MSAEditor.cpp
@@ -25,12 +25,14 @@
 #include <QGridLayout>
 #include <QHBoxLayout>
 #include <QLabel>
+#include <QMainWindow>
 #include <QMessageBox>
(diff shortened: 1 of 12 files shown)


dkandrov:
Thanks!

epushkova:
As far as you are ex-teammate, I am proposing some refactoring: there should be a common widget that reacts on screen change and provides a proper size - so all the other text drawing widgets (SeqView, CircularView, Assembly, etc) can be inherited from it and don't bother about the screen.

ggrekhov:
@epushkova, everything is simpler. See the pull request #23.

ggrekhov:
I would consider another refactoring: extract a basic class for graphic widgets that have a cachedView. I added processing of devicePixelRatio() to these widgets but the basic class will simplify creating of new such widgets and remove copy/paste.

epushkova:
Your suggestion and mine do not contradict.

ggrekhov:
Yes, they don't, but there is no code for processing screen changes now

PR #51, opened 2020-01-16
Ugene 6224 2

rasputinkirill:
Fix for 5 of 6 problems found on the sprint review (except MacOS delete button interception)

### DIFF
--- a/src/corelibs/U2Gui/src/util/GUIUtils.cpp
+++ b/src/corelibs/U2Gui/src/util/GUIUtils.cpp
@@ -226,10 +226,7 @@ const QColor GUIUtils::OK_COLOR = QColor(255,255,255);
 
 void GUIUtils::setWidgetWarning(QWidget *widget, bool value) {
     QColor color = value ? WARNING_COLOR : OK_COLOR;
-    widget->setStyleSheet("background: rgb(" +
-        QString::number(color.red()) + "," +
-        QString::number(color.green()) + "," +
-        QString::number(color.blue()) + ");");
+    widget->setStyleSheet("background: " + color.name());
 }
 
 void GUIUtils::showMessage(QWidget *widgetToPaintOn, QPainter& painter, const QString& message) {
--- a/src/corelibs/U2Gui/src/util/GUIUtils.cpp
+++ b/src/corelibs/U2Gui/src/util/GUIUtils.cpp
@@ -226,7 +226,7 @@ const QColor GUIUtils::OK_COLOR = QColor(255,255,255);
 
 void GUIUtils::setWidgetWarning(QWidget *widget, bool value) {
     QColor color = value ? WARNING_COLOR : OK_COLOR;
-    widget->setStyleSheet("background: " + color.name());
+    widget->setStyleSheet("background-color: " + color.name() + ";");
 }
 
 void GUIUtils::showMessage(QWidget *widgetToPaintOn, QPainter& painter, const QString& message) {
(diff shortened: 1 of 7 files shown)


src/corelibs/U2Gui/src/util/GUIUtils.cpp
```diff
@@ -226,9 +226,10 @@ const QColor GUIUtils::OK_COLOR = QColor(255,255,255);
 
 void GUIUtils::setWidgetWarning(QWidget *widget, bool value) {
     QColor color = value ? WARNING_COLOR : OK_COLOR;
```
dkandrov, line 228: Why do you use `background: rgb(r,g,b)` format of a color in style definition, but don't use `background: #RRGGBB` format? As I understand the result of the `QColor.name()` method returns a color name in such format: `#RRGGBB`. And you won't have to convert the color components separately as you do in lines below.

dkandrov: Documentation links:
- https://doc.qt.io/archives/qt-4.8/qcolor.html#name
- https://doc.qt.io/qt-5/qcolor.html#name

dkandrov (changes requested):
See my previous comments.

dkandrov (approved):
Looks good to me.

src/corelibs/U2View/src/ov_msa/find_pattern/FindPatternMsaTask.cpp
```diff
@@ -50,7 +50,7 @@ void FindPatternMsaTask::createSearchTaskForCurrentSequence() {
     FindAlgorithmTaskSettings algoSettings;
     algoSettings.searchIsCircular = false;
     algoSettings.strand = FindAlgorithmStrand_Direct;
-    algoSettings.maxResult2Find = settings.findSettings.maxResult2Find;
+    algoSettings.maxResult2Find = FindAlgorithmSettings::MAX_RESULT_TO_FIND_UNLIMITED;
```
atiunov, line 54: Create an issue and add a TODO comment with a link to the issue.

src/corelibs/U2View/src/ov_msa/find_pattern/FindPatternMsaTask.cpp
```diff
@@ -75,7 +75,7 @@ QList<Task*> FindPatternMsaTask::onSubTaskFinished(Task* subTask) {
 
     if (subTask == searchInSingleSequenceTask) {
         getResultFromTask();
-        if (currentSequenceIndex < settings.msaObj->getNumRows() && totalResultsCounter < settings.findSettings.maxResult2Find) {
+        if (currentSequenceIndex < settings.msaObj->getNumRows() && totalResultsCounter <= settings.findSettings.maxResult2Find) {
```
atiunov: There is no need to create a new search task if there are already settings.findSettings.maxResult2Find results found.

src/corelibs/U2View/src/ov_msa/find_pattern/FindPatternMsaWidgetFactory.cpp
```diff
@@ -32,7 +34,7 @@ namespace U2 {
 
 const QString FindPatternMsaWidgetFactory::GROUP_ID = "OP_MSA_FIND_PATTERN";
 const QString FindPatternMsaWidgetFactory::GROUP_ICON_STR = ":core/images/find_dialog.png";
-const QString FindPatternMsaWidgetFactory::GROUP_DOC_PAGE = "666";
+const QString FindPatternMsaWidgetFactory::GROUP_DOC_PAGE = HelpButton::INVALID_VALUE;
```
atiunov: Do not initialize static variables with another static variables if they are in different compilation units. It will cause an undefined behavior.

atiunov (approved):

PR #55, opened 2020-02-03
UGENE-6695 A videocard in the application settings dialog should be s…

DmitriiSukhomlinov:
…elected with a radio button

### DIFF
--- a/src/plugins/genome_aligner/src/GenomeAlignerSettingsWidget.cpp
+++ b/src/plugins/genome_aligner/src/GenomeAlignerSettingsWidget.cpp
@@ -52,7 +52,7 @@ GenomeAlignerSettingsWidget::GenomeAlignerSettingsWidget(QWidget* parent) : DnaA
 
     buildIndexFileButton->toggle();
 #ifdef OPENCL_SUPPORT
-    if (AppContext::getOpenCLGpuRegistry()->getEnabledGpus().empty()) {
+    if (AppContext::getOpenCLGpuRegistry()->getEnabledGpu() != nullptr) {
 #endif
         gpuBox->setEnabled(false);
 #ifdef OPENCL_SUPPORT
--- a/src/plugins/genome_aligner/src/GenomeAlignerSettingsWidget.cpp
+++ b/src/plugins/genome_aligner/src/GenomeAlignerSettingsWidget.cpp
@@ -52,7 +52,7 @@ GenomeAlignerSettingsWidget::GenomeAlignerSettingsWidget(QWidget* parent) : DnaA
 
     buildIndexFileButton->toggle();
 #ifdef OPENCL_SUPPORT
-    if (AppContext::getOpenCLGpuRegistry()->getEnabledGpu() != nullptr) {
+    if (AppContext::getOpenCLGpuRegistry()->getEnabledGpu() == nullptr) {
 #endif
         gpuBox->setEnabled(false);
 #ifdef OPENCL_SUPPORT
--- a/src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
+++ b/src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
@@ -293,19 +293,18 @@ void OpenCLSupportPlugin::unregisterAvailableGpus() {
 }
 
 void OpenCLSupportPlugin::loadGpusSettings() {
-    Settings* s = AppContext::getSettings();
-    QString enabledGpu = s->getValue(OPENCL_GPU_REGISTRY_SETTINGS_GPU_ENABLED, QVariant()).toString();
-    CHECK_EXT(!enabledGpu.isEmpty(), gpus.first()->setEnabled(true), );
-    
-    bool enabledGpuWasFound = false;
-    foreach(OpenCLGpuModel* m, gpus) {
-        CHECK_CONTINUE(m->getName() == enabledGpu);
+    CHECK(!gpus.isEmpty(), );
 
-        m->setEnabled(true);
-        enabledGpuWasFound = true;
-        break;
-    }
-    if (!enabledGpuWasFound) {
+    Settings* s = AppContext::getSettings();
+    QString enabledGpuName = s->getValue(OPENCL_GPU_REGISTRY_SETTINGS_GPU_ENABLED, QVariant()).toString();
+    CHECK_EXT(!enabledGpuName.isEmpty(), gpus.first()->setEnabled(true), );
+
+    OpenCLGpuModel *enabledGpu = AppContext::getOpenCLGpuRegistry()->getGpuByName(enabledGpuName);
+    if (nullptr != enabledGpu) {
+        SAFE_POINT(gpus.contains(enabledGpu), "The GPU is absent", );
+
+        enabledGpu->setEnabled(true);
+    } else {
         gpus.first()->setEnabled(true);
     }
 }
--- a/src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
+++ b/src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
@@ -54,7 +54,7 @@ extern "C" Q_DECL_EXPORT QString * U2_PLUGIN_FAIL_MASSAGE_FUNC() {
                                                <a href=\"%1\">%1</a>").arg("http://ugene.net/using-video-cards.html"));
 }
 
-const static char * RESOURCE_OPENCL_GPU_NAME = "OpenCLGpu";
+const char *OpenCLSupportPlugin::RESOURCE_OPENCL_GPU_NAME = "OpenCLGpu";
 
 OpenCLSupportPlugin::OpenCLSupportPlugin() : Plugin(tr("OpenCL Support"),
                                                     tr("Plugin provides support for OpenCL-enabled GPUs.") ) {
@@ -89,6 +89,7 @@ OpenCLSupportPlugin::OpenCLSupportPlugin() : Plugin(tr("OpenCL Support"),
 OpenCLSupportPlugin::~OpenCLSupportPlugin() {
     OpenCLGpuRegistry* registry = AppContext::getOpenCLGpuRegistry();
     CHECK(NULL != registry, );
+    registry->saveGpusSettings();
     unregisterAvailableGpus();
     AppResourcePool::instance()->unregisterResource(RESOURCE_OPENCL_GPU);
     registry->setOpenCLHelper(NULL);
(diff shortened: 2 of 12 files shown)


src/plugins/genome_aligner/src/GenomeAlignerSettingsWidget.cpp
```diff
@@ -52,7 +52,7 @@ GenomeAlignerSettingsWidget::GenomeAlignerSettingsWidget(QWidget* parent) : DnaA
 
     buildIndexFileButton->toggle();
 #ifdef OPENCL_SUPPORT
-    if (AppContext::getOpenCLGpuRegistry()->getEnabledGpus().empty()) {
+    if (AppContext::getOpenCLGpuRegistry()->getEnabledGpu() != nullptr) {
```
atiunov: An incorrect condition.

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -292,16 +293,20 @@ void OpenCLSupportPlugin::unregisterAvailableGpus() {
...
-            QString::number(m->getId()) + OPENCL_GPU_SETTINGS_ENABLED;
-        QVariant enabled_v = s->getValue( key );
-        if( !enabled_v.isNull() ) {
-            m->setEnabled( enabled_v.toBool() );
-        } else {
-            m->setEnabled( true );
-        }
+    Settings* s = AppContext::getSettings();
+    QString enabledGpu = s->getValue(OPENCL_GPU_REGISTRY_SETTINGS_GPU_ENABLED, QVariant()).toString();
+    CHECK_EXT(!enabledGpu.isEmpty(), gpus.first()->setEnabled(true), );
```
atiunov: "gpus" list could be empty. Do not call "first" method without check.

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -292,16 +293,20 @@ void OpenCLSupportPlugin::unregisterAvailableGpus() {
...
+    bool enabledGpuWasFound = false;
+    foreach(OpenCLGpuModel* m, gpus) {
+        CHECK_CONTINUE(m->getName() == enabledGpu);
+
+        m->setEnabled(true);
+        enabledGpuWasFound = true;
+        break;
+    }
+    if (!enabledGpuWasFound) {
+        gpus.first()->setEnabled(true);
```
atiunov, line 308: Check that "gpus" list in not empty.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,155 @@
...
+        if (enabledGpu.startsWith(gpu->getName())) {
+            gpu->setEnabled(true);
+            enabledGpuWasFound = true;
+        } else {
+            gpu->setEnabled(false);
+        }
+    }
+    if (!enabledGpuWasFound) {
+        registeredGpus.first()->setEnabled(true);
     }
```
atiunov: Use method OpenCLGpuRegistry::getGpuByName.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,155 @@
...
-    OpenCLSupportSettingsPageState * s = new OpenCLSupportSettingsPageState( registeredGpus.size() );
-    for( int i = 0, end = s->enabledGpus.size(); i < end; ++i ) {
-        s->enabledGpus[i] = registeredGpus.at(i)->isEnabled();
+    QString nameOfEnabledGpu;
+    foreach (OpenCLGpuModel *gpu, registeredGpus) {
+        CHECK_CONTINUE(gpu->isEnabled());
+
+        nameOfEnabledGpu = gpu->getName();
+        break;
     }
```
atiunov: Use method OpenCLGpuRegistry::getEnabledGpu.

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -292,16 +293,20 @@ void OpenCLSupportPlugin::unregisterAvailableGpus() {
...
     }
```
atiunov: Use method OpenCLGpuRegistry::getGpuByName.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,155 @@
...
     w->setState(state);
     return w;
 }
 
 const QString OpenCLSupportSettingsPageController::helpPageId = QString("24742346");
 
-OpenCLSupportSettingsPageState::OpenCLSupportSettingsPageState( int num_gpus ) {
-    assert( num_gpus >= 0 );
-    enabledGpus.resize( num_gpus );
+OpenCLSupportSettingsPageState::OpenCLSupportSettingsPageState(const QString name)
```
atiunov: const QString &name

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,155 @@
...
+    : enabledGpuName(name) {
+}
+
+QString OpenCLSupportSettingsPageState::getEnabledGpuName() const {
```
atiunov: const QString &

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,155 @@
...
+        if (rb->text().startsWith(enbledGpuName)) {
+            rb->setChecked(true);
+            checkedRadioButtonwasFound = true;
+        } else {
+            rb->setChecked(false);
+        }
+    }
+    if (!checkedRadioButtonwasFound) {
+        gpuRadioButtons.first()->setChecked(true);
     }
```
atiunov: Create a QMap<QString, QRadioButton *> to get the buttons. There is no need to uncheck radio buttons manually, if they are grouped. They are grouped automatically if they have one parent.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,155 @@
...
 }
 
-AppSettingsGUIPageState* OpenCLSupportSettingsPageWidget::getState( QString & /*err*/ ) const {
-    OpenCLSupportSettingsPageState * state = new OpenCLSupportSettingsPageState( gpuEnableChecks.size() );
+AppSettingsGUIPageState *OpenCLSupportSettingsPageWidget::getState(QString & /*err*/) const {
+    CHECK(!gpuRadioButtons.isEmpty(), nullptr);
```
atiunov: Do not return nullptr, the caller doesn't expect it. Return a state with an empty name.

atiunov (changes requested):
Please, fix the places I've commented.

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
DmitriiSukhomlinov: Technically - yes, but literally - no, so, I think, SAFE_POINT is required.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
DmitriiSukhomlinov: I'm not quite sure, that I can change a status of the enabled GPU only and everything will work well, it may cause an unexpected behaviour. It's better to set one GPU as enabled and the rest of GPUs as disabled.

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -293,19 +293,18 @@ void OpenCLSupportPlugin::unregisterAvailableGpus() {
...
-        enabledGpuWasFound = true;
-        break;
-    }
-    if (!enabledGpuWasFound) {
+    QString enabledGpuName = s->getValue(OPENCL_GPU_REGISTRY_SETTINGS_GPU_ENABLED, QVariant()).toString();
+    CHECK_EXT(!enabledGpuName.isEmpty(), gpus.first()->setEnabled(true), );
+
+    OpenCLGpuModel *enabledGpu = AppContext::getOpenCLGpuRegistry()->getGpuByName(enabledGpuName);
+    if (nullptr != enabledGpu) {
+        SAFE_POINT(gpus.contains(enabledGpu), "The GPU is absent", );
```
atiunov, line 304: It is possible, if user changed a video card in the computer. Also it is not critical. Just put this check to as additional condition to "if" statement.

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -293,19 +293,18 @@ void OpenCLSupportPlugin::unregisterAvailableGpus() {
 }
 
 void OpenCLSupportPlugin::loadGpusSettings() {
+    SAFE_POINT(!gpus.isEmpty(), "The GPU list is empty!", );
```
atiunov: Replace with **CHECK**. I'm not sure that it absolutely impossible.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,150 @@
...
-    OpenCLSupportSettingsPageState * s = qobject_cast<OpenCLSupportSettingsPageState *>(_s);
+    OpenCLSupportSettingsPageState *s = qobject_cast<OpenCLSupportSettingsPageState *>(_s);
 
     //saving state of enabled/disabled GPUs into registry
-    for( int i = 0, end = s->enabledGpus.size(); i < end; ++i ) {
-        registeredGpus[i]->setEnabled( s->enabledGpus[i] );
+    const QString enabledGpu = s->getEnabledGpuName();
+    bool enabledGpuWasFound = false;
+    foreach (OpenCLGpuModel *gpu, registeredGpus) {
+        if (enabledGpu.startsWith(gpu->getName())) {
```
atiunov: They should be equal, shouldn't they?

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
atiunov: OK, as you wish.

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
```diff
@@ -19,125 +19,150 @@
...
+    foreach (QRadioButton *rb, gpuRadioButtons.values()) {
+        if (rb->text().startsWith(enbledGpuName)) {
+            rb->setChecked(true);
+            checkedRadioButtonwasFound = true;
+            break;
+        }
+    }
+    if (!checkedRadioButtonwasFound) {
+        gpuRadioButtons.values().first()->setChecked(true);
     }
```
atiunov: Get the radio button by the GPU name from the map.

atiunov (changes requested):

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
DmitriiSukhomlinov: If a video card was removed, it won't be in the registry, so "getGpuByName" will return nullptr

src/plugins/opencl_support/src/OpenCLSupportSettingsController.cpp
DmitriiSukhomlinov: Since I've added QMap with the name of GPU as keys - yes, but, earlier, the "enabledGpu" was a sign near the QRadioButton and this sign also includes the information about the memory.

atiunov: But you've already added a QMap? So, it is time to replace it to equality sign.

atiunov (approved):

PR #58, opened 2020-02-04
UGENE-6684 Building a dot plot for an amino acid sequence cause SAFE_POINT

rasputinkirill:


atiunov (changes requested):
The algorithm should be able to work with the window size equal to the shortest sequence length. The offered changes should be reverted.

src/plugins/dotplot/src/DotPlotDialog.cpp
```diff
@@ -157,6 +161,25 @@ void DotPlotDialog::sl_loadedStateChanged() {
...
 }
 
+void DotPlotDialog::sl_sequenceSelectorIndexChanged() {
+    int xIdx = xAxisCombo->currentIndex();
+    int yIdx = yAxisCombo->currentIndex();
+    
+    QList<GObject *> sequenceObjects = GObjectUtils::findAllObjects(UOF_LoadedOnly, GObjectTypes::SEQUENCE);
+    
+    U2SequenceObject *objX = qobject_cast<U2SequenceObject *>(sequenceObjects[xIdx]);
+    U2SequenceObject *objY = qobject_cast<U2SequenceObject *>(sequenceObjects[yIdx]);
```
atiunov: Add SAFE_POINTs to be sure that these indexes are legal.

atiunov (changes requested):

atiunov (approved):

PR #66, opened 2020-02-20
UGENE-6691 "Find pattern in MSA" task changes selection even if a user does a drag and drop action right in this moment

rasputinkirill:


src/corelibs/U2View/src/ov_msa/find_pattern/FindPatternMsaWidget.cpp
```diff
@@ -969,6 +973,10 @@ void FindPatternMsaWidget::sl_findPatternTaskStateChanged() {
             prevPushButton->setEnabled(true);
             checkState();
             correctSearchInCombo();
+            if (setSelectionToFirstValuebleResult) {
+                setSelectionToFirstValuebleResult = true;
```
DmitriiSukhomlinov: It has no sence - you can get this line only if **setSelectionToFirstValuebleResult** is true.

DmitriiSukhomlinov (changes requested):

DmitriiSukhomlinov (approved):

PR #67, opened 2020-02-27
UGENE-6649 Invalid extracted PCR product when a primer goes through the circular sequence start coordinate

rasputinkirill:


src/plugins/pcr/src/InSilicoPcrTask.cpp
```diff
@@ -161,8 +161,10 @@ InSilicoPcrTask::PrimerBind InSilicoPcrTask::getPrimerBind(const FindAlgorithmRe
         const qint64 reverseRegionEndPos = reverse.region.endPos();
         const qint64 sequenceSize = settings.sequence.size();
         if (reverseRegionEndPos > sequenceSize) {
-            result.region = U2Region(reverse.region.startPos, sequenceSize - reverse.region.startPos);
-            if (!settings.isCircular) {
+            if (settings.isCircular) {
+                result.region = U2Region(reverse.region.startPos, sequenceSize - reverse.region.startPos);
+            } else {
+                result.region = U2Region(reverse.region.startPos, sequenceSize);
                 result.region.length -= reverse.region.startPos;
```
DmitriiSukhomlinov: If and else have the same conditions, so you can short it.

DmitriiSukhomlinov (changes requested):

src/plugins/pcr/src/InSilicoPcrTask.cpp
rasputinkirill: Done, and checked with existing tests.

PR #72, opened 2020-03-05
Fix UGENE-6722

ichebyki:
Just check registered gpu count before saving state for some gpu
If count == 0 return from function


rasputinkirill:
If you going to merge with GitHub web interface do not forget to change target branch from master to release-34.0.

PR #85, opened 2020-04-07
UGENE-6760 Two SV and adding an annotation object to sequence

rasputinkirill:


src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -914,6 +914,7 @@ QString AnnotatedDNAView::addObject(GObject *o) {
     } else if (o->getGObjectType() == GObjectTypes::ANNOTATION_TABLE) {
         AnnotationTableObject *ao = qobject_cast<AnnotationTableObject *>(o);
         SAFE_POINT(NULL != ao, "Invalid annotation table!", QString::null);
+        checkAndAddObjectToOtherViews(o);
```
atiunov: Each view should work only with itself. Other views will add the object by themselves, if it is supposed to do.

atiunov (changes requested):

rasputinkirill:
For now i fix only problem described in issue - adding annotation table to sequence view.
IMHO problem with removing annotation table should be resolved as separate issue, because current task already took too much time.

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -786,6 +786,30 @@ void AnnotatedDNAView::sl_updatePasteAction() {
...
+    CHECK(o != nullptr, );
+    QList<AnnotationTableObject *> currentAnnotations = getAnnotationObjects(false);
+    QList<GObject *> allObjs;
+    foreach (Document *d, AppContext::getProject()->getDocuments()) {
+        allObjs << d->getObjects();
+    }
+    QList<GObject *> objectsToAdd;
+    foreach (GObject *obj, allObjs) {
+        foreach (const GObjectRelation &rel, obj->findRelatedObjectsByRole(ObjectRole_Sequence)) {
+            if (rel.ref.entityRef == o->getEntityRef() && !currentAnnotations.contains(qobject_cast<AnnotationTableObject*>(obj))) {
```
atiunov: Use GObjectUtils::findObjectsRelatedToObjectByRole instead of checking of all objects.

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -786,6 +786,30 @@ void AnnotatedDNAView::sl_updatePasteAction() {
...
+                objectsToAdd << obj;
+            }
+        }
+    }
+    foreach(GObject* obj, objectsToAdd) {
+        QString error = addObject(obj);
+        if (!error.isEmpty()) {
+            QMessageBox::critical(NULL, tr("Error!"), error);
```
atiunov: If you have several views, then you'll receive several message boxes. I offer just to write the message to log without additional notification.

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -865,7 +889,7 @@ QString AnnotatedDNAView::tryAddObject(GObject* o) {
             CHECK(!d.isNull(), "");
 
             rCtx = findRelatedSequenceContexts(o);
-            if (rCtx.isEmpty()) {
+            if (rCtx.isEmpty() || d->relationIsSet) {
```
atiunov: Shouldn't there be "!d->relationIsSet"?

src/plugins/GUITestBase/src/GTUtilsMdi.cpp
```diff
@@ -264,7 +264,7 @@ void GTUtilsMdi::activateWindow(HI::GUITestOpStatus &os, const QString &windowNa
     MainWindow* mw = AppContext::getMainWindow();
     GT_CHECK(mw != NULL, "MainWindow == NULL");
 
-    CHECK(!activeWindowTitle(os).contains(windowName), );
+    CHECK(!(activeWindowTitle(os) == windowName), );
```
atiunov, line 267: Be ready to fix tests, that expect the previous behavior.

atiunov (changes requested):

src/plugins/GUITestBase/src/GTUtilsMdi.cpp
rasputinkirill: In case of too many test fails, i'll change my scenario.

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
atiunov: Ok, it's move clear.

atiunov (approved):

PR #91, opened 2020-04-22
UGENE-6781 GTF parsing errors for GTF from UCSC with additional whitespace characters

rasputinkirill:
…space characters

### DIFF
--- a/src/corelibs/U2Formats/src/GTFFormat.cpp
+++ b/src/corelibs/U2Formats/src/GTFFormat.cpp
@@ -361,7 +361,7 @@ bool parseAttributes(QString attributesStr, QMap<QString, QString>& parsedAttrVa
 
     while (pos < attributesStr.size()) {
         //skip sequence of spaces
-        while (attributesStr.indexOf(' ', pos) == pos) {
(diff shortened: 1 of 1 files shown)


src/corelibs/U2Formats/src/GTFFormat.cpp
```diff
@@ -360,6 +360,11 @@ bool parseAttributes(QString attributesStr, QMap<QString, QString>& parsedAttrVa
     int pos = 0;
 
     while (pos < attributesStr.size()) {
+        //skip sequence of spaces
+        while (attributesStr.indexOf(' ', pos) == pos) {
+            pos++;
+        }
```
atiunov: This code will skip all spaces, but the next code is supposed to find one space.

atiunov (changes requested):

src/corelibs/U2Formats/src/GTFFormat.cpp
rasputinkirill: Here we "rewind" to first non-space symbol, because string could be beginning with more thn one space.
Also this code made more simple.

atiunov (approved):

PR #95, opened 2020-05-06
Fix clang-format errors for Windows

ichebyki:
Fix clang-format errors for Windows after merging UTI-390 to master


ichebyki:
Local build was passed

PR #99, opened 2020-05-20
Fix UGENE-6768

ichebyki:
Check argc count, if 1 then exit with non-zero code

### DIFF
--- a/src/plugins_checker/src/Main.cpp
+++ b/src/plugins_checker/src/Main.cpp
@@ -58,6 +58,12 @@ static void registerCoreServices() {
 }
 
 int main(int argc, char **argv) {
+    // User lauches the program manually
+    if (argc == 1) {
+        printf("Use \"ugeneui\" to start Unipro UGENE graphical interface or \"ugenecl\" to use the command-line interface.");
+        return 1;
+    }
+
     CrashHandler::setupHandler();
     CrashHandler::setSendCrashReports(false);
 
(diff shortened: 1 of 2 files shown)


rasputinkirill (approved):

src/plugins_checker/src/Main.cpp
```diff
@@ -58,6 +58,12 @@ static void registerCoreServices() {
 }
 
 int main(int argc, char **argv) {
+    // User lauches the program manually
+    if (argc == 1) {
+        printf("Use \"ugeneui\" to start Unipro UGENE graphical interface or \"ugenecl\" to use the command-line interface.");
+        return 1;
+    }
+
```
atiunov: According the issue description, in case of GUI presented, a message box should be shown.

src/ugenem/src/main.cpp
```diff
@@ -44,6 +44,12 @@ QString loadReport(int argc, char *argv[]) {
 }    // namespace
 
 int main(int argc, char *argv[]) {
+    // User lauches the program manually
+    if (argc == 1) {
+        printf("Use \"ugeneui\" to start Unipro UGENE graphical interface or \"ugenecl\" to use the command-line interface.");
+        return 1;
+    }
+
```
atiunov: According the issue description, in case of GUI presented, a message box should be shown.

atiunov (changes requested):

PR #101, opened 2020-05-22
UGENE-6799 Task member becomes invalid during the task execution

DmitriiSukhomlinov:


src/plugins/workflow_designer/src/WorkflowSceneIOTasks.cpp
```diff
@@ -101,6 +101,7 @@ Task::ReportResult LoadWorkflowSceneTask::report() {
     if (hasError()) {
         return ReportResult_Finished;
     }
+    CHECK(scene != nullptr, ReportResult_Finished);
```
atiunov, line 104: "scene" pointer should be QPointer to be nullified on the WD closing.

src/plugins/workflow_designer/src/WorkflowViewController.cpp
```diff
@@ -304,6 +304,12 @@ WorkflowView::WorkflowView(WorkflowGObject *go)
 
 WorkflowView::~WorkflowView() {
     uiLog.trace("~WorkflowView");
+    if (!loadWorkflowSceneTask.isNull()) {
+        TaskScheduler*  ts = AppContext::getTaskScheduler();
+        SAFE_POINT(ts != nullptr, "TaskScheduler isn't found", );
+
+        ts->cancelTask(loadWorkflowSceneTask);
```
atiunov: You can just use Task::cancel() here.

atiunov (changes requested):

src/plugins/workflow_designer/src/WorkflowSceneIOTasks.cpp
DmitriiSukhomlinov: Fixed

src/plugins/workflow_designer/src/WorkflowViewController.cpp
DmitriiSukhomlinov: Fixed

atiunov (approved):

PR #103, opened 2020-05-27
UGENE-6804 Add a dialog before deleting of an object in Project View

rasputinkirill:


src/corelibs/U2Gui/src/util/project/ProjectTreeController.cpp
```diff
@@ -792,8 +795,64 @@ void ProjectTreeController::sl_onRemoveSelectedItems() {
...
+    foreach(const Folder& folder, selectedFolders) {
+        selectedItemNames.append(folder.getFolderName());
+    }
+    foreach(GObject* obj, selectedObjects) {
+        selectedItemNames.append(obj->getGObjectName());
+    }
+    
+    QString fullItemsNamesList;
+
+    QString warningMessageText = (1 == selectedItemNames.count()) ? tr("Remove selected item?") : tr("Remove selected items?");
```
yalgaer: Yoda style comparison is not necessary. gcc compiler will catch and show warnings for lines with assignments in expression.
More: you can't assign here at all (to list.count())

src/corelibs/U2Gui/src/util/project/ProjectTreeController.cpp
```diff
@@ -792,8 +795,64 @@ void ProjectTreeController::sl_onRemoveSelectedItems() {
...
+    warningMessageText += "<ul style=\"margin-top:5px;margin-bottom:0px\"><li>";
+
+    const bool tooManyItemsSelected = (MAX_DISPLAING_NAME_COUNT < selectedItemNames.count());
```
yalgaer: MAX_DISPLAING_NAME_COUNT is a constant and can't be < or >.
To make it more readable (in English) I would suggest avoiding the use of Yoda style if not necessary:
The more readable text will be:  'selectedItemNames.count() < MAX_DISPLAING_NAME_COUNT'

yalgaer: Why do you want to repeat all item names to users? There is no such requirement in the issue.

The code would be much simpler if you just show a message box with a short message and Yes/No buttons.
What is there are 100+ items in the list. It won't fit the screen.

src/corelibs/U2Gui/src/util/project/ProjectTreeController.cpp
```diff
@@ -792,8 +795,64 @@ void ProjectTreeController::sl_onRemoveSelectedItems() {
...
+        }
+
+        if (MAX_DISPLAING_NAME_COUNT >= ++itemCounter) {
+            // cut long names
+            if (NAME_DISPLAYING_SYMBOLS_COUNT < name.size()) {
+                name = name.left(NAME_DISPLAYING_SYMBOLS_COUNT);
+                name += "...";
+            }
+            warningMessageText += name;
+            warningMessageText += "</li><li>";
```
yalgaer, line 830: break from the loop here? Why do you want to add dozens of <li>...</li> elements here?

src/corelibs/U2Gui/src/util/project/ProjectTreeController.cpp
```diff
@@ -792,8 +795,64 @@ void ProjectTreeController::sl_onRemoveSelectedItems() {
...
+    warningMessageText = warningMessageText.left(
+        warningMessageText.length() - 9); // 9 is "</li><li>" length
+    warningMessageText += "</li></ul>";
+    if (tooManyItemsSelected) {
+        warningMessageText += "<pre style=\"margin-top:0px;\">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;...</pre>";
+    }
+
+    QObjectScopedPointer<QMessageBox> questionBox = new QMessageBox;
+    questionBox->setIcon(QMessageBox::Question);
+    questionBox->setWindowTitle(QObject::tr("Removing Dashboards"));
```
yalgaer: Dashboards?

yalgaer (commented):
Check inline comments

src/corelibs/U2Gui/src/util/project/ProjectTreeController.cpp
atiunov: Nobody reads compiler warnings.

atiunov: The warning message is incorrect. Check the issue details for the correct message.

atiunov (changes requested):

src/corelibs/U2Gui/src/util/project/ProjectTreeController.cpp
yalgaer: Here what we have in our build file for those who don't like to read warnings like these:
QMAKE_CXXFLAGS += -Werror=parentheses :)

yalgaer (approved):

PR #107, opened 2020-06-02
UGENE-6775 Selection in circular mode should the same behavior as in linear mode

rasputinkirill:


yalgaer (commented):
Everything works fine except 1 segment: try to select around 0-base point (3 o'clock item)
The behavior should be the same as with other points (like 12 o'clock) but it is different/

PR #110, opened 2020-06-22
UGENE-6592 Incorrect "Simple extended" consensus calculation in case of changing "Threshold"

rasputinkirill:
For now because of result consensus character '\0' we got following picture. Is it right?
Capture

### DIFF
--- a/src/corelibs/U2Algorithm/src/util_msa_consensus/MaConsensusAlgorithmSimpleExtended.cpp
+++ b/src/corelibs/U2Algorithm/src/util_msa_consensus/MaConsensusAlgorithmSimpleExtended.cpp
@@ -155,7 +155,8 @@ char MaConsensusAlgorithmSimpleExtended::getConsensusChar(const MultipleAlignmen
     QVector<QVector<char>> frequencies = getFrequences(ma, column, seqIdx);
 
     char bestCharacter = INVALID_CONS_CHAR;
-    const int thresholdCount = qRound(static_cast<double>((frequencies.size() - 1) * getThreshold()) / 100);
+    int thresholdCount = ((frequencies.size() - 1) * getThreshold()) / 100;
+    thresholdCount += (((frequencies.size() - 1) * getThreshold()) % 100) == 0 ? 0 : 1;
 
     for (int frequency = frequencies.size() - 1; frequency > 0; frequency--) {
         CHECK_CONTINUE(0 < frequencies[frequency].size());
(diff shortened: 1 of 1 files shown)


atiunov:
Consensus calculation rules are described in the comment to UGENE-5715. There cannot be an empty space, there should be some character of the extended DNA alphabet or gap.
There should be 'W' in consensus in your case. Since neither 'T' nor 'W' fits the threshold, at least two characters should be involved into the calculations. The most frequent character 'T' should be merged with a bit less frequent 'W'. The result is 'W".

src/corelibs/U2Algorithm/src/util_msa_consensus/MaConsensusAlgorithmSimpleExtended.cpp
```diff
@@ -155,7 +155,8 @@ char MaConsensusAlgorithmSimpleExtended::getConsensusChar(const MultipleAlignmen
     QVector<QVector<char>> frequencies = getFrequences(ma, column, seqIdx);
 
     char bestCharacter = INVALID_CONS_CHAR;
-    const int thresholdCount = qRound(static_cast<double>((frequencies.size() - 1) * getThreshold()) / 100);
+    int thresholdCount = ((frequencies.size() - 1) * getThreshold()) / 100;
+    thresholdCount += (((frequencies.size() - 1) * getThreshold()) % 100) == 0 ? 0 : 1;
```
atiunov: This code is ok, but I offer to use "ceil" function to round to the bigger integer.

src/corelibs/U2Algorithm/src/util_msa_consensus/MaConsensusAlgorithmSimpleExtended.cpp
```diff
@@ -171,11 +172,6 @@ char MaConsensusAlgorithmSimpleExtended::getConsensusChar(const MultipleAlignmen
             // There are no characters that fit the threshold, but we can merge a bit less popular characters
             return mergeCharacters(frequencies[frequency] << bestCharacter);
         }
-        if (frequencies[frequency].size() == 1) {
-            // There are no characters that fit the threshold and we found a single the most popular character.
-            // We need more characters to merge them with this one.
-            bestCharacter = frequencies[frequency].first();
-        }
```
atiunov: Restore this code, it is a specific case. See details in the comment to UGENE-5715.

atiunov (changes requested):

atiunov (approved):

PR #111, opened 2020-06-22
UGENE-5886 The reference selection in the Sanger Reads Editor can be out of boundaries

rasputinkirill:


src/corelibs/U2View/src/ov_msa/McaEditorReferenceArea.cpp
```diff
@@ -241,6 +241,10 @@ void McaEditorReferenceArea::setReferenceSelection(QMouseEvent *e) {
         start = pos;
         count = 1;
     }
+    int length = editor->getReferenceContext()->getSequenceLength();
```
atiunov: qint64

atiunov (approved):

PR #112, opened 2020-06-28
Fix UGENE-6369

ichebyki:
Just ignore visible state of the "Plugin Viewer" windows while starting Ugene gui
Always save false for visibility of the window, so don't restore the visibility state

### DIFF
--- a/src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
+++ b/src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
@@ -50,9 +50,8 @@ PluginViewerController::PluginViewerController() {
 PluginViewerController::~PluginViewerController() {
     AppContext::getPluginSupport()->disconnect(this);
 
-    // 'true' mode is not functional anymore after service<->plugin model refactoring
-    // so always save false for pluginview settings, also see UGENE-6369
-    AppContext::getSettings()->setValue(PLUGIN_VIEW_SETTINGS + "isVisible", false);
+    // 'true' mode is not functional anymore after some old service<->plugin model refactoring
+    // so we don't save state for pluginview settings, see UGENE-6369
 
     if (mdiWindow) {
         AppContext::getMainWindow()->getMDIManager()->closeMDIWindow(mdiWindow);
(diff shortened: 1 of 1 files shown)


src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
```diff
@@ -45,16 +45,14 @@ PluginViewerController::PluginViewerController() {
...
-    if (AppContext::getSettings()->getValue(PLUGIN_VIEW_SETTINGS + "isVisible", false).toBool()) {
-        createWindow();
-    }
 }
 
 PluginViewerController::~PluginViewerController() {
     AppContext::getPluginSupport()->disconnect(this);
 
-    AppContext::getSettings()->setValue(PLUGIN_VIEW_SETTINGS + "isVisible", mdiWindow != NULL);
+    // 'true' mode is not functional anymore after service<->plugin model refactoring
```
yalgaer: Which service/plugin model refactoring do you mention here?

src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
```diff
@@ -45,16 +45,14 @@ PluginViewerController::PluginViewerController() {
...
+    // so always save false for pluginview settings, also see UGENE-6369
+    AppContext::getSettings()->setValue(PLUGIN_VIEW_SETTINGS + "isVisible", false);
```
yalgaer: Why do we save the value? Is it used anywhere? If not, lets remove this line

+ The same for the PLUGIN_VIEW_SETTINGS constant

yalgaer (changes requested):

src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
ichebyki: This is an old comment, not my

src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
ichebyki: removed

src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
yalgaer: PLUGIN_VIEW_SETTINGS is not used/needed anymore

src/ugeneui/src/plugin_viewer/PluginViewerController.cpp
yalgaer: I think we can remove PLUGIN_VIEW_SETTINGS constant also - it is not used anymore.

yalgaer (approved):

PR #115, opened 2020-07-07
UGENE-5735 Regression: Colon in the name of an exported file results in an error

rasputinkirill:


src/corelibs/U2Core/src/util/FileAndDirectoryUtils.h
```diff
@@ -44,6 +44,7 @@ class U2CORE_EXPORT FileAndDirectoryUtils {
     static bool isFileEmpty(const QString &url);
     static void dumpStringToFile(QFile *f, QString &str);    //Be aware: string will be cleared after dumping
     static QString getAbsolutePath(const QString &filePath);
+    static bool isFilepathCorrect(const QString &filePath);
```
yalgaer, line 47: Please add docs what this method does and when it return false;

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -142,4 +142,88 @@ QString FileAndDirectoryUtils::getAbsolutePath(const QString &filePath) {
...
+        // Check for illegal characters
+        if (illegal.contains(c))
+            return false;
+    }
+
+    // Check for device names in filenames
+    static QStringList devices;
+
+    if (!devices.count())
+        devices << "CON"
```
yalgaer, line 175: Create a static list outside of this function: QStringList WINDOWS_DEVICES_LIST
Reason: 50% of your method is an initialization code for that list

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -142,4 +142,88 @@ QString FileAndDirectoryUtils::getAbsolutePath(const QString &filePath) {
...
+                << "LPT5"
+                << "LPT6"
+                << "LPT7"
+                << "LPT8"
+                << "LPT9";
+
+    const QFileInfo fi(path);
+    const QString basename = fi.baseName();
+
+    foreach (const QString &s, devices)
```
yalgaer, line 203: Rename 's' to 'deviceFileName'

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -142,4 +142,88 @@ QString FileAndDirectoryUtils::getAbsolutePath(const QString &filePath) {
...
+                << "LPT0"
+                << "LPT1"
+                << "LPT2"
+                << "LPT3"
+                << "LPT4"
+                << "LPT5"
+                << "LPT6"
+                << "LPT7"
+                << "LPT8"
+                << "LPT9";
```
yalgaer, line 198: How many LPTs to you want to expect here? Do you have other devices missed on the list?
If you do not know: it is not a good solution to list the devices here. Let's not do it at all.

yalgaer (changes requested):

yalgaer:
This problem should be solved another way without listing all device files.

PR #116, opened 2020-07-08
UGENE-6839 SAFE_POINT during cufflinks working

DmitriiSukhomlinov:


src/plugins/external_tool_support/src/cufflinks/CufflinksWorker.cpp
```diff
@@ -365,6 +366,9 @@ Task *CufflinksWorker::tick() {
 
         return cufflinksSupportTask;
     } else if (input->isEnded()) {
+        if (output->hasMessage() == 0) {
+            return new FailTask(tr("No annatations was processed by Cufflinks, abort."));
```
yalgaer, line 370: Typo: No annotations were processed by Cufflinks, aborting...

yalgaer (changes requested):

ichebyki (approved):

src/plugins/external_tool_support/src/cufflinks/CufflinksWorker.cpp
```diff
@@ -365,6 +366,9 @@ Task *CufflinksWorker::tick() {
 
         return cufflinksSupportTask;
     } else if (input->isEnded()) {
+        if (output->hasMessage() == 0) {
```
atiunov, line 369: If the worker already processed some messages, then the output bus could contain messages.

src/plugins/external_tool_support/src/cufflinks/CufflinksWorker.cpp
atiunov, line 370: Do not interrupt the workflow run because of this. If there were errors during message processing, then write an error message to the dashboard and skip the message.
I've added a comment to the issue. I almost sure that this behavior of cufflinks is correct, so the worker should interpret the tool behavior correctly. It is not an error at all.

atiunov (changes requested):
Check a comment to the issue and offer another solution.

PR #117, opened 2020-07-13
UGENE-6715 UGENE shouldn't set "Directory to save color scheme" if there is no permission to write in there

rasputinkirill:


src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -142,4 +142,9 @@ QString FileAndDirectoryUtils::getAbsolutePath(const QString &filePath) {
     return QFileInfo(result).absoluteFilePath();
 }
 
+bool FileAndDirectoryUtils::isDirectoryWritable(const QString &path) {
```
yalgaer: You actually do not need this method.

src/corelibs/U2View/src/ov_msa/ColorSchemaDialogController.cpp
```diff
@@ -374,6 +379,10 @@ void ColorSchemaSettingsPageWidget::sl_schemaChanged(int index) {
 void ColorSchemaSettingsPageWidget::sl_onColorsDirButton() {
     QString path = colorsDirEdit->text();
     QString dir = U2FileDialog::getExistingDirectory(this, tr("Choose Folder"), path, QFileDialog::ShowDirsOnly | QFileDialog::DontResolveSymlinks);
+    if (!FileAndDirectoryUtils::isDirectoryWritable(dir)) {
```
yalgaer, line 385: You already have an existing dir path: all your need is to check if it is writable FileInfo(path).isWritable()
This way you can avoid the creation of a 1-line helper methods.

src/corelibs/U2View/src/ov_msa/ColorSchemaDialogController.cpp
```diff
@@ -374,6 +379,10 @@ void ColorSchemaSettingsPageWidget::sl_schemaChanged(int index) {
...
+        QMessageBox::warning(this, L10N::warningTitle(), tr("You don't have permissions to write in selected folder."));
+        return;
+    }
     if (!dir.isEmpty()) {
```
yalgaer: How dir can be empty here if you checked if it is writable above?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5627,6 +5627,47 @@ GUI_TEST_CLASS_DEFINITION(test_6714) {
...
+    QDir().mkpath(sandBoxDir + "read_only_dir");
+    GTFile::setReadOnly(os, sandBoxDir + "read_only_dir");
+
+    //new NewColorSchemeCreator(os, "test_5268", NewColorSchemeCreator::nucl, NewColorSchemeCreator::Change)
+    
+    class Scenario : public CustomScenario {
+    public:
+        Scenario() {};
+        void run(HI::GUITestOpStatus &os) {
+            QWidget *dialog = QApplication::activeModalWidget();
```
yalgaer: With GTWidget::activeModalWidget(os) you do not need to check if the result is null (saves 1 line in the code)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5627,6 +5627,47 @@ GUI_TEST_CLASS_DEFINITION(test_6714) {
...
+            CHECK_SET_ERR(dialog, "activeModalWidget is NULL");
+
+            QTreeWidget *tree = qobject_cast<QTreeWidget *>(GTWidget::findWidget(os, "tree"));
+            CHECK_SET_ERR(tree, "tree widger not found");
+
+            QList<QTreeWidgetItem *> items = GTTreeWidget::getItems(tree->invisibleRootItem());
```
yalgaer: GTTreeWidget has helper methods to find an item by text.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5627,6 +5627,47 @@ GUI_TEST_CLASS_DEFINITION(test_6714) {
...
+        }
+    };
+    
+    // 1. Open {Settings -> Preferences -> Alignment Color Scheme}. 
+    GTUtilsDialog::waitForDialog(os, new NewColorSchemeCreator(os, new Scenario()));
+    GTMenu::clickMainMenuItem(os, QStringList() << "Settings"
+                                                << "Preferences...");
+    // 2. Choose read only folder by pressing "..." button
+    // Expected state: warning message about read only folder has appeared
+    GTFile::setReadWrite(os, sandBoxDir + "read_only_dir");
```
yalgaer: This line is not needed. If test fails it will never be executed.

yalgaer (changes requested):

rasputinkirill:
FileInfo(path).isWritable() not work properly for directories.
The only way to be sure is directory "writable" or not - create file in it.
https://forum.qt.io/topic/112928/how-to-check-if-a-folder-is-writable-using-qt
So FileAndDirectoryUtils::isDirectoryWritable was updated.

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -142,4 +143,13 @@ QString FileAndDirectoryUtils::getAbsolutePath(const QString &filePath) {
...
+    if (!QFileInfo(path).isDir()) {
+        return false;
+    }
+
+    QTemporaryFile tmp(path + "/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX");
```
yalgaer: Are you sure that the file is created on the file system with no call to file.open() ?

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -142,4 +143,13 @@ QString FileAndDirectoryUtils::getAbsolutePath(const QString &filePath) {
     return QFileInfo(result).absoluteFilePath();
 }
 
+bool FileAndDirectoryUtils::isDirectoryWritable(const QString &path) {
```
yalgaer: We already have a method like this:

bool TmpDirChecker::checkWritePermissions(const QString &dirPath) 
We can move that method's code here.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5627,6 +5627,39 @@ GUI_TEST_CLASS_DEFINITION(test_6714) {
...
+    public:
+        Scenario() {};
+        void run(HI::GUITestOpStatus &os) {
+            QWidget *dialog = GTWidget::getActiveModalWidget(os);
+
+            QTreeWidget *tree = qobject_cast<QTreeWidget *>(GTWidget::findWidget(os, "tree"));
+            CHECK_SET_ERR(tree, "tree widger not found");
+
+            QTreeWidgetItem *item = GTTreeWidget::findItem(os, tree, "  Alignment Color Scheme");
+            GTMouseDriver::moveTo(GTTreeWidget::getItemCenter(os, item));
```
yalgaer: Hint: You can use GTTreeWidget::click method.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5627,6 +5627,39 @@ GUI_TEST_CLASS_DEFINITION(test_6714) {
...
+        }
+    };
+    
+    // 1. Open {Settings -> Preferences -> Alignment Color Scheme}. 
+    GTUtilsDialog::waitForDialog(os, new NewColorSchemeCreator(os, new Scenario()));
+    GTMenu::clickMainMenuItem(os, QStringList() << "Settings"
+                                                << "Preferences...");
+    // 2. Choose read only folder by pressing "..." button
+    // Expected state: warning message about read only folder has appeared
+}
```
yalgaer, line 5658: Please add the check for the warning message here

yalgaer (changes requested):

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
rasputinkirill: Moved this method from TmpDirChecker to FileAndDirectoryUtils because there it more easy to find.

src/corelibs/U2Core/src/tasks/SaveDocumentTask.cpp
```diff
@@ -44,7 +45,7 @@ namespace U2 {
 
 bool isNoWritePermission(GUrl &url) {
     if (!QFile::exists(url.getURLString())) {
-        return (!TmpDirChecker::checkWritePermissions(url.dirPath()));
+        return (!FileAndDirectoryUtils::isDirectoryWritable(url.dirPath()));
```
yalgaer: Not a problem: braces are not needed

yalgaer (approved):

PR #119, opened 2020-07-13
UGENE-6674 Uninformative error message in "Export subaligment" dialog for incorrect file name

rasputinkirill:


src/corelibs/U2View/src/ov_msa/CreateSubalignmentDialogController.cpp
```diff
@@ -200,7 +201,7 @@ void CreateSubalignmentDialogController::accept() {
         QMessageBox::critical(this, this->windowTitle(), tr("No path specified"));
         return;
     }
-    if (fi.baseName().isEmpty()) {
+    if (fi.baseName().isEmpty() || possibleDir.exists()) {
```
yalgaer: You can use https://doc.qt.io/qt-5/qfileinfo.html#isDir to check if file-info is dir and do not create a new variable here.

yalgaer (changes requested):

yalgaer (approved):

PR #121, opened 2020-07-15
UGENE-5500

ichebyki:
return null QByteArray

### DIFF
--- a/src/corelibs/U2Core/src/globals/GUrl.cpp
+++ b/src/corelibs/U2Core/src/globals/GUrl.cpp
@@ -153,13 +153,13 @@ QByteArray GUrl::getURLStringAnsi(int codePage) const {
 
     DWORD buffSize = WideCharToMultiByte(codePage, 0, wPath.c_str(), -1, NULL, 0, NULL, NULL);
     if (!buffSize) {
-        return nullptr;
+        return QByteArray();
     }
 
     char *buffer = new char[buffSize + 1];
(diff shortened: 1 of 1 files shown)


src/corelibs/U2Core/src/globals/GUrl.cpp
```diff
@@ -153,13 +153,13 @@ QByteArray GUrl::getURLStringAnsi(int codePage) const {
 
     DWORD buffSize = WideCharToMultiByte(codePage, 0, wPath.c_str(), -1, NULL, 0, NULL, NULL);
     if (!buffSize) {
-        return nullptr;
+        return QByteArray();
```
yalgaer, line 156: Your original code was 100% correct. 
QByteArray(nullptr) is equal to the QByteArray()

Docs:
https://doc.qt.io/qt-5/qbytearray.html#QByteArray-1

Example:
    QByteArray nullArray(nullptr);
    printf("Data '%s'\n", nullArray.constData());
    fflush(stdout);

yalgaer (approved):

src/corelibs/U2Core/src/globals/GUrl.cpp
atiunov: It is much more better to use default constructor instead of implicit conversion.

atiunov (approved):

DmitriiSukhomlinov (approved):

PR #122, opened 2020-07-16
Ucr 281

DmitriiSukhomlinov:
Crash report 23432

### DIFF
--- a/src/corelibs/U2Formats/src/PDBFormat.cpp
+++ b/src/corelibs/U2Formats/src/PDBFormat.cpp
@@ -334,7 +334,7 @@ bool PDBFormat::PDBParser::seqResContains(char chainIdentifier, int residueIndex
     }
 }
 
-void PDBFormat::PDBParser::parseAtom(BioStruct3D &biostruct, U2OpStatus &) {
+void PDBFormat::PDBParser::parseAtom(BioStruct3D &biostruct, U2OpStatus &os) {
     /*
     Record Format
 
@@ -426,7 +426,11 @@ void PDBFormat::PDBParser::parseAtom(BioStruct3D &biostruct, U2OpStatus &) {
     biostruct.modelMap[currentModelIndex + 1].insert(id, a);
 
     if (atomIsInChain) {
+        SAFE_POINT_EXT(chainIndex <= biostruct.moleculeMap.size(), os.setError("Unexpected chainIndex"), );
+
         SharedMolecule &mol = biostruct.moleculeMap[chainIndex];
+        SAFE_POINT_EXT(currentModelIndex <= mol->models.size(), os.setError("Unexpected currentModelIndex"), );
+
         Molecule3DModel &model3D = mol->models[currentModelIndex];
         model3D.atoms.insert(id, a);
     }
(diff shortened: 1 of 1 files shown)


src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -426,7 +426,11 @@ void PDBFormat::PDBParser::parseAtom(BioStruct3D &biostruct, U2OpStatus &) {
     biostruct.modelMap[currentModelIndex + 1].insert(id, a);
 
     if (atomIsInChain) {
+        SAFE_POINT_EXT(0 <= chainIndex && chainIndex < biostruct.moleculeMap.size(), os.setError("Unexpected chainIndex"), );
```
yalgaer, line 429: The error is expected to be shown to a user and chainIndex/currentModelIndex is not known to the user.

src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -426,7 +426,11 @@ void PDBFormat::PDBParser::parseAtom(BioStruct3D &biostruct, U2OpStatus &) {
...
+
         SharedMolecule &mol = biostruct.moleculeMap[chainIndex];
+        SAFE_POINT_EXT(0 <= currentModelIndex && currentModelIndex < mol->models.size(), os.setError("Unexpected currentModelIndex"), );
```
yalgaer, line 432: Using Yoda-style comparisons makes code unreadable.

Our build has a special compiler flag (see ugene-globals.pri) that will make assignment in expressions a compile-time error.

src/corelibs/U2Formats/src/PDBFormat.cpp
yalgaer, line 429: tr()?

yalgaer (changes requested):
UI or XML test?

src/corelibs/U2Formats/src/PDBFormat.cpp
DmitriiSukhomlinov: Historically, I've never translated SAFE_POINTs. Should i?

DmitriiSukhomlinov: You mean, that this value should be added to the message?

src/corelibs/U2Formats/src/PDBFormat.cpp
DmitriiSukhomlinov: As far as I'm concerned, this is the most convenient way for the double comparisons visualizing. Let's remember, how it was in school:
a <= X < b
From the left to the right, the left number is the smallest one, the right number is the largest one, the variable in the middle. Unfortunately, C++ doesn't support double comparisons as one logical action, so the following code is invalid:
`0 <= currentModelIndex < mol->models.size()`
But, we can use LOGICAL AND and transfer the above-mentioned code to the most similar one:
`0 <= currentModelIndex && currentModelIndex < mol->models.size()`

The second part isn't clear for me.

DmitriiSukhomlinov:
Incorrect fix

PR #125, opened 2020-08-12
Fix UGENE-6859 vs2015 problem with qt5.12.9

ichebyki:
Add DEFINE += __STDC_LIMIT_MACROS for win32-msvc2015


yalgaer (approved):
We should have the same for VS 2019

PR #126, opened 2020-08-14
UGENE-6852: Translation for the selection is empty

yalgaer:


yalgaer:
About changes:

1) I completely rewrote drawDirectTranslation function. The code of the function is shortened 3-4 times and uses common logic for different modes (translation with & without selected regions). 
This function does not rely on helpers now. I can do the same for the drawComplementTranslation if needed (not changed in this change to simplify the review)

2) I renamed visibleRows to visibleFrames to limit the naming options. Before this patch we had 'rows', 'frames', 'lines'. Now we have only translation frames (part of the semantical data model) and lines (part of the on-screen layout).

src/corelibs/U2View/src/ov_sequence/DetView.cpp
```diff
@@ -493,27 +480,15 @@ void DetView::setSelectedTranslations() {
...
-
-void DetView::updateTranslationsState(const U2Strand::Direction direction) {
-    QVector<U2Region> selectedRegions = ctx->getSequenceSelection()->getSelectedRegions();
-    QList<bool> lineState = QList<bool>() << false << false << false;
-    foreach (const U2Region &reg, selectedRegions) {
-        int mod = direction == U2Strand::Direct ? reg.startPos % 3 : ((ctx->getSequenceLength() - reg.endPos()) % 3);
-        lineState[mod] = true;
+void DetView::updateTranslationRowsVisibilityBySelectionState() {
+    QVector<bool> frameRowVisibilityFlag(6, false);
+    for (auto region : ctx->getSequenceSelection()->getSelectedRegions()) {
```
DmitriiSukhomlinov: `for (const auto& region : ctx->getSequenceSelection()->getSelectedRegions()) {
`

src/corelibs/U2View/src/ov_sequence/DetView.cpp
```diff
@@ -716,7 +691,7 @@ void DetView::updateVerticalScrollBar() {
         verticalScrollBar->setSliderPosition(0);
     }
 
-    int maximum = 0;
+    int maximum;
```
DmitriiSukhomlinov, line 694: Why did you remove initialization? It's better to initialize variables in any case, even if these variables will be initialized further, because you or somebody may change the code, but won't initialize variable. Please, undo.

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewSingleLineRenderer.cpp
```diff
@@ -419,112 +411,96 @@ void DetViewSingleLineRenderer::drawTranslations(QPainter &p, const U2Region &vi
...
+        qint64 translationIndent = translationStartPos - visibleRange.startPos;
+        const char *sequenceToTranslate = visibleSequence + translationIndent;
+        qint64 sequenceToTranslateLength = qMin(visibleRange.endPos() + 1, ctx->getSequenceLength()) - translationStartPos;
+        QByteArray aminoSequence = translate(aminoTT, sequenceToTranslate, sequenceToTranslateLength);
+
+        QVector<bool> aminoVisibilityFlags(aminoSequence.size(), true);    // Visible amino acid positions. All positions are visible by default.
+        if (ctx->getTranslationState() == SequenceObjectContext::TS_AnnotationsOrSelection) {
+            aminoVisibilityFlags.fill(false);    // Mark as visible only acids within selected regions.
+            U2Region translatableRange(visibleRange.startPos + translationIndent, aminoSequence.length() * 3);
+            for (auto selectedRegion : ctx->getSequenceSelection()->getSelectedRegions()) {
```
DmitriiSukhomlinov: for (const auto& selectedRegion : ctx->getSequenceSelection()->getSelectedRegions()) {

src/corelibs/U2View/src/ov_sequence/DetView.cpp
```diff
@@ -120,7 +114,7 @@ DetView::DetView(QWidget *p, SequenceObjectContext *ctx)
     showTranslationAction->setChecked(hasAmino);
 
     assert(ctx->getSequenceObject() != NULL);
-    featureFlags &= !GSLV_FF_SupportsCustomRange;
+    featureFlags &= !GSLV_FF_SupportsCustomRange;    //TODO: bug?
```
DmitriiSukhomlinov, line 117: Why?

DmitriiSukhomlinov (changes requested):

src/corelibs/U2View/src/ov_sequence/DetView.cpp
yalgaer: Fixed

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewSingleLineRenderer.cpp
yalgaer: Fixed

src/corelibs/U2View/src/ov_sequence/DetView.cpp
yalgaer: ! is a boolean operation. I think the original intention was to use ~
https://www.geeksforgeeks.org/bitwise-operators-in-c-cpp/

src/corelibs/U2View/src/ov_sequence/DetView.cpp
yalgaer: Use of uninitialized value will break our build.
   QMAKE_CXXFLAGS += -Werror=uninitialized
   QMAKE_CXXFLAGS += -Werror=maybe-uninitialized

The artificial '0' value will hide real bugs: if we forget to assign a real value to it nobody will detect this issue.
Not using the artificial value at all will make a compiler to detect the problem and break Commit build.

src/corelibs/U2View/src/ov_sequence/DetView.cpp
DmitriiSukhomlinov: Yes, looks like it's true. Fix it, or  create the issue

DmitriiSukhomlinov (approved):

PR #128, opened 2020-08-24
UGENE-5842 Error on opening Sanger Reads Editor after unloading

rasputinkirill:


yalgaer (dismissed):

src/corelibs/U2Core/src/dbi/DbiDocumentFormat.cpp
```diff
@@ -80,17 +80,19 @@ Document *DbiDocumentFormat::loadDocument(IOAdapter *io, const U2DbiRef &dstDbiR
...
     objects << prepareObjects(handle, objectIds);
 
+    U2DbiRef resultDbiRef = dstDbiRef;
     if (fs.value(DEEP_COPY_OBJECT, false).toBool()) {
         QList<GObject *> clonedObjects = cloneObjects(objects, dstDbiRef, fs, os);
         qDeleteAll(objects);
         CHECK_OP_EXT(os, qDeleteAll(clonedObjects), NULL);
         objects = clonedObjects;
     } else {
+        resultDbiRef = srcDbiRef;
```
yalgaer, line 90: This is a wrong solution that breaks the method contract: the method must return document with dstDbiRef

The correct solution is to resolve the loaded object by the original reference:

MaEditorTasks.cpp:103
` GObject *obj = GObjectUtils::selectObjectByReference(unloadedReference, UOF_LoadedOnly);`

yalgaer (changes requested):

PR #129, opened 2020-08-24
UGENE-6863. 'Open file by OS' and 'Open folder with the file' buttons…

yalgaer:
The path fixes 'dashboard folder move' issue for both new and old kind of dashboards.

### DIFF
--- a/src/corelibs/U2Designer/src/dashboard/DashboardWidget.cpp
+++ b/src/corelibs/U2Designer/src/dashboard/DashboardWidget.cpp
@@ -312,14 +312,20 @@ void DashboardFileButton::sl_openFileClicked() {
     QString type = tokens[0];
     QString url = tokens[1];
     QFileInfo fileInfo(url);
-    if (type == "folder") {
+    bool isFolder = type == "folder";
+    if (isFolder) {
         fileInfo = QFileInfo(fileInfo.absolutePath());
     }
     if (!fileInfo.exists()) {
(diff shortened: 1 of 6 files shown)


src/corelibs/U2Designer/src/dashboard/DashboardWidget.cpp
```diff
@@ -280,19 +281,52 @@ void DashboardFileButton::addUrlActionsToMenu(QMenu *menu, const QString &url, b
     menu->addAction(openFileAction);
 }
 
+/**
+ * Finds a file in the given dashboard dir by path suffix of the 'fileInfo'.
+ * Returns new file info or the old one if the file detection algorithm is failed.
+ * This method is designed to find dashboard output files in moved dashboard.
+ */
+static QFileInfo findFileOpenCandidateInTheDashboardOutputDir(const QFileInfo &dashboardDirInfo, const QFileInfo &fileInfo) {
```
DmitriiSukhomlinov, line 289: Please, separate declaration to the ".h" file.

DmitriiSukhomlinov (changes requested):

src/corelibs/U2Designer/src/dashboard/DashboardWidget.cpp
yalgaer: This function is not not designed to be a public API or to be a class member (depend on a class context). 
It is designed to be a stateless local function. I prefer limiting the scope/context where possible.

If you have a strong opinion that the function should be declared as a public one in the .h file I will move it.

DmitriiSukhomlinov (approved):

PR #134, opened 2020-09-03
UGENE-6871

atiunov:
Shortcuts are missed on macOS.

### DIFF
--- a/src/ugeneui/src/main_window/ProxyStyle.cpp
+++ b/src/ugeneui/src/main_window/ProxyStyle.cpp
@@ -1,0 +1,29 @@
+#include "ProxyStyle.h"
+
+#include <QMenu>
+
+namespace U2 {
+
+ProxyStyle::ProxyStyle(QStyle *style)
+    : QProxyStyle(style)
+{
+
+}
+
+void ProxyStyle::polish(QWidget *widget) {
+    QProxyStyle::polish(widget);
+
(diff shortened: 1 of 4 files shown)


src/ugeneui/src/main_window/ProxyStyle.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+void ProxyStyle::polish(QWidget *widget) {
+    QProxyStyle::polish(widget);
+
+#if QT_VERSION_CHECK(5, 10, 0) < QT_VERSION
+    // It's a workaround for https://bugreports.qt.io/browse/QTBUG-49435:
+    // shortcuts are not shown in context menus.
+    auto *menu = qobject_cast<QMenu *>(widget);
+    if (nullptr != menu) {
```
yalgaer, line 20: no problem: you do not need to use Yoda style with modern compilers. UGENE build will break if you have assignment within expressions in your code.

or read the text above like this (mimics the way devs reads the Yoda expressions in code):

Modern compilers with no Yoda style to use. Break will UGENE it once use anyone for bad.

yalgaer (approved):

src/ugeneui/src/main_window/ProxyStyle.cpp
atiunov: We use rather old compilers and rather old C++ standard.
Assignment in the condition will not break the build. It will cause a warning, which could be ignored.

PR #136, opened 2020-09-09
Temporary disable OpenCL plugin

ichebyki:
Just return nullptr from function

### DIFF
--- a/src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
+++ b/src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
@@ -36,10 +36,8 @@
 namespace U2 {
 
 extern "C" Q_DECL_EXPORT Plugin *U2_PLUGIN_INIT_FUNC() {
-    // Temporary disable OpenCL plugin
-    //OpenCLSupportPlugin *plug = new OpenCLSupportPlugin();
-    //return plug;
-    return nullptr;
+    OpenCLSupportPlugin *plug = new OpenCLSupportPlugin();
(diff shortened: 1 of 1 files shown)


atiunov (approved):

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -36,8 +36,10 @@
 namespace U2 {
 
 extern "C" Q_DECL_EXPORT Plugin *U2_PLUGIN_INIT_FUNC() {
```
yalgaer: If you want to remove the plugin from our distribution bundles why don't you just remove in create_bundle.sh/cmd ?

yalgaer (changes requested):

src/plugins/opencl_support/src/OpenCLSupportPlugin.cpp
```diff
@@ -36,8 +36,10 @@
...
-    OpenCLSupportPlugin *plug = new OpenCLSupportPlugin();
-    return plug;
+    // Temporary disable OpenCL plugin
+    //OpenCLSupportPlugin *plug = new OpenCLSupportPlugin();
+    //return plug;
+    return nullptr;
```
DmitriiSukhomlinov: It's a bad behavior, error notification will be shown in every UGENE run. Add return to the beginning of the OpenCLSupportPlugin constructor.

DmitriiSukhomlinov (changes requested):

PR #137, opened 2020-09-09
readme.md and build.txt update

DmitriiSukhomlinov:


README.md
```diff
@@ -4,31 +4,37 @@
 
 ### Prerequisites
 
-Make sure the Qt (>= 5.3.2) development libraries are installed:
+Make sure the Qt (>= 5.3.2 and < 5.15) development libraries are installed:
```
yalgaer: 5.4.2 <= ... <= 5.15?
I use 5.15 and 5.4.2 locally on Linux, so I know that it works for these OSes.

How do you know that UGENE builds with 5.3.2?

README.md
```diff
@@ -4,31 +4,37 @@
...
 
+We highly recommend you to install Qt to any operating system with Qt online-installer. This is the main way for Windows and macOS operating systems, also available  for Linux. To install Qt, download the online-installer from Qt official site (http://www.qt.io/download/). Make sure, that you have installed the following components:
+* Qt Prebuild components
+* QtCharts
```
yalgaer: I think we do not use QtCharts anywhere in our code. Do we?

build.txt
```diff
@@ -2,16 +2,23 @@
 Prerequisites
```
yalgaer: Looks like this file contains duplicated content and should be dropped. 
If for some reason we can't remove it today we can put "read Readme.md" inside

yalgaer (changes requested):

README.md
DmitriiSukhomlinov: I've got an qt assert when I have tried to run UGENE with 5.15 on Windows. I will fix max version.
I do not know about 5.3.2, it just was here. I can change it to 5.4.2.

README.md
DmitriiSukhomlinov: I've got it installed for some reason, but maybe, accidentally. Remove the line

build.txt
DmitriiSukhomlinov: I think we can remove this file at all

yalgaer (approved):

README.md
yalgaer: Just to confirm: we do not use/need this QtCharts.

PR #138, opened 2020-09-11
UGENE-6715 + UGENE-6870

rasputinkirill:


src/corelibs/U2View/src/ov_msa/ColorSchemaDialogController.cpp
```diff
@@ -382,17 +383,35 @@ void ColorSchemaSettingsPageWidget::sl_schemaChanged(int index) {
     }
 }
 
+bool ColorSchemaSettingsPageWidget::sl_validateSchemesDir() {
+    QString dirPath = colorsDirEdit->text();
```
yalgaer, line 387: Use GUrlUtils::prepareDirLocation()

src/corelibs/U2View/src/ov_msa/ColorSchemaDialogController.cpp
```diff
@@ -340,6 +340,7 @@ ColorSchemaSettingsPageWidget::ColorSchemaSettingsPageWidget(ColorSchemaSettings
     connect(addSchemaButton, SIGNAL(clicked()), SLOT(sl_onAddColorSchema()));
     connect(deleteSchemaButton, SIGNAL(clicked()), SLOT(sl_onDeleteColorSchema()));
     connect(colorSchemas, SIGNAL(currentRowChanged(int)), SLOT(sl_schemaChanged(int)));
+    connect(colorsDirEdit, SIGNAL(editingFinished()), SLOT(sl_validateSchemesDir()));
```
yalgaer, line 343: This is not needed anymore because colorsDir is validated when OK button is pressed.

More than that this won't give you the desired effect: colorEditValue still stays incorrect after the message is shown to user.

yalgaer (changes requested):

src/corelibs/U2View/src/ov_msa/ColorSchemaDialogController.cpp
yalgaer: Ensure that prepareDirLocation() is called only once at the end and not for every path entered by user

PR #139, opened 2020-09-11
UGENE-6579 Intersect annotations does not work if an annotation group consists of digits only

rasputinkirill:
Test(s) will be added after successful code review.


src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
```diff
@@ -179,6 +181,9 @@ QList<Task *> BedtoolsIntersectAnnotationsByEntityTask::onSubTaskFinished(Task *
         Document *resultDoc = loadResultTask->getDocument();
         CHECK_EXT(resultDoc != NULL, setError(tr("Result document is NULL")), res);
         result = resultDoc->findGObjectByType(GObjectTypes::ANNOTATION_TABLE);
+        AnnotationTableObject *ato = qobject_cast<AnnotationTableObject *>(result.first());
```
yalgaer, line 185: Check that result list is not empty before calling first() ?

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
```diff
@@ -179,6 +181,9 @@ QList<Task *> BedtoolsIntersectAnnotationsByEntityTask::onSubTaskFinished(Task *
...
+        CHECK_EXT(ato != NULL, setError(tr("Failed to cast to AnnotationTableObject")), res);
```
yalgaer: Here you failed to find annotation object, not to cast. 

findGObjectByType(GObjectTypes::ANNOTATION_TABLE) guarantees that you have a valid AnnotationTableObjects in the list.

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
```diff
@@ -179,6 +181,9 @@ QList<Task *> BedtoolsIntersectAnnotationsByEntityTask::onSubTaskFinished(Task *
...
+        renameAnnotationsFromBed(ato->getRootGroup());
```
yalgaer, line 186: Process all annotation table objects from the list (result)?

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
```diff
@@ -199,11 +204,32 @@ Document *BedtoolsIntersectAnnotationsByEntityTask::createAnnotationsDocument(co
...
     return doc;
 }
 
+void BedtoolsIntersectAnnotationsByEntityTask::renameAnnotationsForBed(AnnotationGroup *group) {
+    bool isNumber = false;
+    group->getName().toInt(&isNumber);
+    if (isNumber) {
+        group->setName(group->getName() + RENAME_STRING);
+    }
+    foreach(AnnotationGroup * g, group->getSubgroups()) {
```
yalgaer: use for, not foreach

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.h
```diff
@@ -139,6 +142,8 @@ class BedtoolsIntersectAnnotationsByEntityTask : public ExternalToolSupportTask
     SaveMultipleDocuments *saveAnnotationsTask;
     BedtoolsIntersectTask *intersectTask;
     LoadDocumentTask *loadResultTask;
+
+    QMap<QString, int> renameMap;
```
yalgaer: Delete unused field

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
```diff
@@ -199,11 +204,32 @@ Document *BedtoolsIntersectAnnotationsByEntityTask::createAnnotationsDocument(co
...
+        renameAnnotationsForBed(g);
+    }
+}
+
+void BedtoolsIntersectAnnotationsByEntityTask::renameAnnotationsFromBed(AnnotationGroup *group) {
+    if (group->getName().contains(RENAME_STRING)) {
```
yalgaer: Use endsWith()

yalgaer (changes requested):

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
rasputinkirill: According to knowledge how to bedtools works there is only one annotation table in the resulting file.

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
rasputinkirill: Why? What's wrong with foreach?

yalgaer (approved):

src/plugins/external_tool_support/src/bedtools/BedtoolsIntersectTask.cpp
yalgaer: foreach (Qt only) was added when there was no for (C++)

Compare
 foreach(AnnotationGroup * g, group->getSubgroups()) {
vs
 for(AnnotationGroup * g: group->getSubgroups()) {

Tools that are not aware about QT may not understand foreach macro (refactoring, highlighting, formatting, etc...)

PR #140, opened 2020-09-15
UGENE-5579 Restriction site is not found when region contains sequence end/start

rasputinkirill:


src/plugins/enzymes/src/FindEnzymesTask.cpp
```diff
@@ -351,7 +351,15 @@ Task *FindEnzymesAutoAnnotationUpdater::createAutoAnnotationsUpdateTask(const Au
 
     U2SequenceObject *dnaObj = aa->getSeqObject();
     const U2Region wholeSequenceRegion = U2Region(0, dnaObj->getSequenceLength());
-    cfg.searchRegion = savedSearchRegion.intersect(wholeSequenceRegion);
+    if (cfg.circular) {
+        cfg.searchRegion = savedSearchRegion;
```
yalgaer, line 355: Use 1 line: 
cfg.searchRegion = U2Region(savedSearchRegion.startPos, qMin(savedSearchRegion.length, wholeSequenceRegion.length);

And add a comment: "In circular mode the region can have an overflow to handle end/start positions correctly"

yalgaer (changes requested):
Could you please rebase this fix to release-36.0 ?

yalgaer:
The fix looks good but there is a mess with branches merge. Please submit a new clean pull request for release-36.0 only

PR #142, opened 2020-09-15
6873: UGENE crashes in "Pairwise alignment" for big alignments

sgrekhov:


src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -307,7 +307,7 @@ void SmithWatermanAlgorithm::calculateMatrixForMultipleAlignmentResult() {
         }
     } while (++i <= static_cast<int>(src_n));
 
-    free(matrix);
+    delete matrix;
```
atiunov: delete [] matrix

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -414,7 +414,7 @@ void SmithWatermanAlgorithm::calculateMatrixForAnnotationsResult() {
     }
 #endif
 
-    free(matrix);
+    delete matrix;
```
atiunov: delete [] matrix

atiunov (changes requested):
Just replacing malloc calls with new is not enough. There is a solution offered in the comment to the issue.

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
sgrekhov: > Just replacing malloc calls with new is not enough. There is a solution offered in the comment to the issue.

This task is called from https://github.com/ugeneunipro/ugene/blob/master/src/corelibs/U2Private/src/TaskSchedulerImpl.cpp#L1052-L1057
```C
        try {
            ti->task->run();
            assert(ti->task->getState() == Task::State_Running);
        } catch (const std::bad_alloc &) {
            onBadAlloc(ti->task);
        }
```
Calling `new` throws `bad_alloc` in case of low memory (unlike `malloc()`) which is caught here and sets an error in `onBadAlloc`

DmitriiSukhomlinov: I agree with @sgrekhov . There is no need to write error to the operationStatus member. If there is not enough memory to allocate, "new" throws the "bad_alloc" exception and TaskScheduler catches it.

atiunov: Ok, you convinced me. But fix delete calls, pointers should be released as pointers to array, not as pointers to a value.

atiunov (changes requested):
Delete calls should be fixed.

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -173,7 +173,7 @@ void SmithWatermanAlgorithm::calculateMatrixForMultipleAlignmentResult() {
 
     n = pat_n * 2;
     int dirn = (4 + pat_n + 3) >> 2;
-    int *buf, *matrix = (int *)malloc(n * sizeof(int) + pat_n * 0x80 + matrixLength * dirn);
+    int *buf, *matrix = new int[n * sizeof(int) + pat_n * 0x80 + matrixLength * dirn];
```
DmitriiSukhomlinov: Not like this. When we use "malloc" we need to pass the exact memory value (in bytes) we need to allocate as an argument, usually it looks like:
`malloc(size_of_array * syze_of_one_element_of_array_in_bytes)`
And when we use "new" we pass just size of the array,  like:
`new[size_of_array]`
Let's figure our case out:
`malloc(n * sizeof(int) + pat_n * 0x80 + matrixLength * dirn)`
`n * sizeof(int)` - **n** is size of araray, **sizeof(int)** - sizeof is a function, which calculates size of the object (or of the type, _int_, in our case)
`pat_n * 0x80 + matrixLength * dirn` - some additional information which is stored in this array (I hate the guy who did this, he should use separate variables)
So, we need to decrease the array size, I think like this:
`new int[n + ((pat_n * 0x80 + matrixLength * dirn) / sizeof(int))];`

PS: test it!

DmitriiSukhomlinov (changes requested):

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
yalgaer: Why can't you keep the old code and just add a check: if (matrix === nullptr) throw bad-alloc... ?

+ there is another place like this with malloc() in the file.

yalgaer (changes requested):

yalgaer:
These is another improvement can be done here that will help to make behavior more predictable and reliable.

UGENE has "AppResource" framework. A task can declare its needs (like I need 1Gb of memory). This framework is used for memory/cpu consuming tasks.

Task scheduler will not run 2 tasks in parallel if their total memory usage exceeds some MAX limit.
If resources required by a single task are above the limit - the task will not be run at all: user will see a valid error message.

So the complete solution would be adding to SW task an estimator: how much resources does it need?

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
yalgaer: new qint8[original size]  looks simpler ?

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -173,7 +173,7 @@ void SmithWatermanAlgorithm::calculateMatrixForMultipleAlignmentResult() {
...
     char *score, *score1 = (char *)(matrix + n);
     unsigned char *dir, *dir2, *dir1 = (unsigned char *)score1 + pat_n * 0x80;
     memset(matrix, 0, n * sizeof(int));
```
yalgaer, line 183: While this is not directly related to the fix it it looks strange to call malloc () memset(0) when you can use a single calloc() instead.
Leaving it up to you if to fix it or not.

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
sgrekhov: Ok. Fix with minimal changes

sgrekhov:
> These is another improvement can be done here that will help to make behavior more predictable and reliable.
> 
> UGENE has "AppResource" framework. A task can declare its needs (like I need 1Gb of memory). This framework is used for memory/cpu consuming tasks.
> 
> Task scheduler will not run 2 tasks in parallel if their total memory usage exceeds some MAX limit.
> If resources required by a single task are above the limit - the task will not be run at all: user will see a valid error message.
> 
> So the complete solution would be adding to SW task an estimator: how much resources does it need?

After the fix the task can be run sucessfully. Seems that UGENE does several attempts to run the task and one of them is sucessful

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -174,6 +174,10 @@ void SmithWatermanAlgorithm::calculateMatrixForMultipleAlignmentResult() {
     n = pat_n * 2;
     int dirn = (4 + pat_n + 3) >> 2;
     int *buf, *matrix = (int *)malloc(n * sizeof(int) + pat_n * 0x80 + matrixLength * dirn);
+    if (matrix == NULL) {
+        std::bad_alloc e;
+        throw e;
```
yalgaer, line 179: not a problem: can be done in 1 line: throw std::bad_alloc();

yalgaer (approved):

atiunov (approved):

PR #144, opened 2020-09-22
UGENE-5576 Crash on building the second tree when the first tree is unloaded

rasputinkirill:


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -2042,6 +2042,29 @@ GUI_TEST_CLASS_DEFINITION(test_5562_3) {
...
+
+    //2. Build phy tree using options panel.
+    QDir().mkdir(QFileInfo(sandBoxDir + "test_5220/COI_1.nwk").dir().absolutePath());
+    GTUtilsDialog::waitForDialog(os, new BuildTreeDialogFiller(os, sandBoxDir + "test_5220/COI_1.nwk", 0, 0, true));
+    GTWidget::click(os, GTAction::button(os, "Build Tree"));
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    //3. Unload tree.
+    GTUtilsDocument::unloadDocument(os, "COI_1.nwk", false);
+    GTGlobals::sleep();
```
yalgaer, line 2059: Please never use sleep in tests. 

Sleeps make your tests slow and unpredictable: how do you know which sleep time is enough.
Wait for tasks to finish for for some UI event.

In your case the UI event is "no document with the given name in the project tree".
Check the GTUtilsProjectTreeView and add this utility method if missed. Use "for (int time = 0; time < GT_OP_WAIT_MILLIS..." pattern for waiting.

src/ugeneui/src/project_support/ProjectImpl.cpp
```diff
@@ -229,6 +230,21 @@ void ProjectImpl::sl_onObjectRelationChanged(const QList<GObjectRelation> &previ
     }
 }
 
+void ProjectImpl::sl_objectLoadedStateChanged() {
+    Document *doc = qobject_cast<Document *>(sender());
+    CHECK(doc != nullptr, )
```
yalgaer, line 235: Not a blocker:
Can be done in  1 line instead of 4:
CHECK( doc != nullptr && !doc.isLoaded())

src/ugeneui/src/project_support/ProjectImpl.cpp
```diff
@@ -229,6 +230,21 @@ void ProjectImpl::sl_onObjectRelationChanged(const QList<GObjectRelation> &previ
...
+    if (doc->isLoaded()) {
+        return;
+    }
+    foreach (GObject *obj, doc->getObjects()) {
+        UnloadedObject *uo = qobject_cast<UnloadedObject *>(obj);
```
yalgaer, line 240: How it is possible that unloaded document has loaded object?
I think the check below is not needed.

src/ugeneui/src/project_support/ProjectImpl.cpp
```diff
@@ -229,6 +230,21 @@ void ProjectImpl::sl_onObjectRelationChanged(const QList<GObjectRelation> &previ
...
+        CHECK(obj != nullptr, )
+        GObjectReference from(uo->getDocument()->getURLString(), uo->getGObjectName(), uo->getLoadedObjectType());
+        //invalidate entity ref
+        updateObjectRelations(from, from);
```
yalgaer, line 244: the call
updateObjectRelations(const GObjectReference &oldRef, const GObjectReference &newRef) 
where both old and new refs are the same looks meaningless.

If it has some 'hidden' behavior could you please be more explicit here? For example create a dedicated method in ProjectImpl to handle this behavior?

yalgaer: Next question: if you remove relation on unload, why don't you restore them back on load?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -2042,6 +2042,29 @@ GUI_TEST_CLASS_DEFINITION(test_5562_3) {
...
 }
 
+GUI_TEST_CLASS_DEFINITION(test_5576) {
+    //1. Open "data/samples/CLUSTALW/COI.aln".
+    GTFileDialog::openFile(os, dataDir + "samples/CLUSTALW/COI.aln");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    GTUtilsOptionPanelMsa::openTab(os, GTUtilsOptionPanelMsa::TreeSettings);
+
+    //2. Build phy tree using options panel.
+    QDir().mkdir(QFileInfo(sandBoxDir + "test_5220/COI_1.nwk").dir().absolutePath());
```
yalgaer, line 2052: Why don't you use a simpler QDir(sandBoxDir).mkdir("test_5220) instead?

src/corelibs/U2Core/src/models/GObject.cpp
```diff
@@ -172,18 +172,10 @@ void GObject::setRelationsInDb(QList<GObjectRelation> &list) const {
         const U2DataType refType = U2ObjectTypeUtils::toDataType(relation.ref.objType);
         const bool relatedObjectDbReferenceValid = relation.ref.entityRef.dbiRef.isValid();
 
-        if (U2Type::Unknown == refType || (relatedObjectDbReferenceValid && !(relation.ref.entityRef.dbiRef == entityRef.dbiRef))) {
+        if (U2Type::Unknown == refType || !relatedObjectDbReferenceValid) {
```
yalgaer, line 175: The removed check checks that self-relations are not allowed. Why do you want to remove it?

src/corelibs/U2Core/src/models/GObject.cpp
```diff
@@ -172,18 +172,10 @@ void GObject::setRelationsInDb(QList<GObjectRelation> &list) const {
...
             continue;
         }
 
-        if (!relatedObjectDbReferenceValid) {
```
yalgaer, line 179: This method had a contract before: if relation object has invalid dbi-ref it will search for the valid target by the name.
Why you decided to remove this behavior? It may affect other logic not related to your bug

yalgaer:
Closing the pull request. We will consider removal of 'unload' option completely.

PR #146, opened 2020-09-28
UGENE-6589 Gaps in area where there are no Sanger reads

rasputinkirill:


src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusUtils.cpp
```diff
@@ -67,14 +67,18 @@ QString MSAConsensusUtils::getConsensusPercentTip(const MultipleAlignment &ma, i
...
             freqs[idx].second = c;
         } else {
             // count gaps
+            if (ignoreLeadingTrailingGaps && ma->isGap(seq, pos, true)) {
+                nSeq--;
+                continue;
+            } 
             ++gaps;
         }
     }
```
yalgaer, line 86: Check that nSeq != 0 and avoid division by 0 below

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusUtils.cpp
```diff
@@ -67,14 +67,18 @@ QString MSAConsensusUtils::getConsensusPercentTip(const MultipleAlignment &ma, i
...
-    for (int seq = 0; seq < nSeq; seq++) {
+    for (int seq = 0; seq < ma->getNumRows(); seq++) {
         uchar c = (uchar)ma->charAt(seq, pos);
         if (c >= 'A' && c <= 'Z') {
             int idx = c - 'A';
             freqs[idx].first++;
             freqs[idx].second = c;
         } else {
             // count gaps
+            if (ignoreLeadingTrailingGaps && ma->isGap(seq, pos, true)) {
```
yalgaer: Add a method ma->isLeadingOrTrailingGap() and do not change ma->isGap()

yalgaer (changes requested):

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusUtils.cpp
yalgaer: Note: there is not such thing as 'trailingGap' in the current gap model at all: isGap returns false for any position that can be considered as 'trailingGap' here.

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
```diff
@@ -327,6 +327,25 @@ bool MsaRowUtils::isGap(int dataLength, const U2MsaRowGapModel &gapModel, int po
     return false;
 }
 
+bool MsaRowUtils::isLeadingOrTrailingGap(int dataLength, const U2MsaRowGapModel &gapModel, int position) {
+    int gapsLength = 0;
+    foreach (const U2MsaGap &gap, gapModel) {
+        if (gap.offset == 0 && gap.offset <= position && position < gap.offset + gap.gap) {
```
yalgaer: gap.offset <= position is not needed here.

Overall I would suggest not to copy-paste, but to write a simple/optimal/readable/documented code focused on the leading or trailing gap only. Here is an example:

```cpp
    CHECK(!gapModel.isEmpty(), false);
    if (gapModel[0].offset == 0 && position < gapModel[0].endPos()) {
        return true;    // leading gap.
    }
    int totalGapsLen = 0;
    for (const U2MsaGap& gap : gapModel) {
        totalGapsLen += gap.gap;
        if (position < gap.offset) {
            return false; // somethere in the middle.
        } 
    }
    return position >= dataLength + totalGapsLen; // trailing gap.
```

yalgaer (approved):

PR #151, opened 2020-11-17
UGENE-6907. Improve object window title format

yalgaer:
Since the patch is trivial I also added a cleanup to the related file.
1) Replaced NULL with nullptr
2) Removed unused public API methods and fields
3) Replaced deprecated foreach loop with native for
4) Made local variables naming convention more explicit
5) Moved some bulky methods from .h to .cpp

No public API was changed as a part of the patch.

### DIFF
--- a/src/corelibs/U2Gui/src/ObjectViewModel.cpp
+++ b/src/corelibs/U2Gui/src/ObjectViewModel.cpp
@@ -21,7 +21,6 @@
 
 #include "ObjectViewModel.h"
 
-#include <QFileInfo>
 #include <QScrollArea>
 #include <QSplitter>
 #include <QVBoxLayout>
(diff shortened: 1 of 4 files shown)


src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -364,222 +377,214 @@ void GObjectViewWindow::setupViewMenu(QMenu *m) {
...
-    QList<GObjectViewWindow *> result;
-    foreach (MWMDIWindow *w, windows) {
-        if (w->windowTitle() == name) {
-            GObjectViewWindow *v = qobject_cast<GObjectViewWindow *>(w);
-            if (v != NULL) {
-                return v;
+    QList<MWMDIWindow *> mdiWindows = AppContext::getMainWindow()->getMDIManager()->getWindows();
+    for (MWMDIWindow *mdiWindow : mdiWindows) {
+        if (mdiWindow->windowTitle() == name) {
+            GObjectViewWindow *objectViewWindow = qobject_cast<GObjectViewWindow *>(mdiWindow);
```
DmitriiSukhomlinov, line 383: `SAFE_POINT(mdiWindows != nullptr);`

yalgaer: mdiWindows is a QList that can never be nullptr here. I think you mean objectViewWindow pointer.
I would avoid SAFE_POINT here in this review because it will change the current UGENE behavior and this is what I want to avoid.

DmitriiSukhomlinov (approved):
Approved, but I'd rather prefer you don't refactor in the same commit with the issue changes

PR #152, opened 2020-11-19
UGENE-6908. SAFE_POINT during the validation Bio module on the first …

evelinabis:
…launch.

When UGENE was first launched, the Bio module from Python was added to
the tools validation list earlier than Python. As a consequence, the
program was aborted by SAFE_POINT. Also Python validation was incorrect
on newer versions of Windows.

Decision:
1. Changing the message for checking the validity of Python.
2. Splitting ExternalToolManagerImpl::addTool() into two methods.

### DIFF
--- a/src/plugins/external_tool_support/src/ExternalToolManager.cpp
+++ b/src/plugins/external_tool_support/src/ExternalToolManager.cpp
@@ -193,7 +193,19 @@ void ExternalToolManagerImpl::addTool(ExternalTool *tool) {
     }
 }
 
-QString ExternalToolManagerImpl::addToolToListsAndReturnToolPath(const ExternalTool &tool) {
+void ExternalToolManagerImpl::addToolToListsIfNecessary(const ExternalTool &tool) {
+    if (!tool.isValid()) {
+        if (tool.getDependencies().isEmpty() || dependenciesAreOk(tool.getId())) {
+            validateList << tool.getId();
+        }
+    }
+
+    if (!validateList.contains(tool.getId()) && !tool.isModule() && !tool.isValid()) {
+        searchList << tool.getId();
+    }
+}
+
+QString ExternalToolManagerImpl::findToolPath(const ExternalTool &tool) {
     QString toolPath;
     SAFE_POINT(nullptr != etRegistry, "The external tool registry is nullptr", toolPath);
 
@@ -209,17 +221,11 @@ QString ExternalToolManagerImpl::addToolToListsAndReturnToolPath(const ExternalT
             } else {
                 toolPath = tool.getPath();
             }
-            validateList << tool.getId();
+        } else {
         }
+    } else if (!tool.isValid()) {
+        toolPath = tool.getPath();
     } else {
-        if (!tool.isValid()) {
-            validateList << tool.getId();
-            toolPath = tool.getPath();
-        }
-    }
-
-    if (!validateList.contains(tool.getId()) && !tool.isModule() && !tool.isValid()) {
-        searchList << tool.getId();
     }
 
     return toolPath;
@@ -346,12 +352,13 @@ void ExternalToolManagerImpl::sl_customToolsLoaded(Task *task) {
 
     QList<ExternalTool *> toolsList = etRegistry->getAllEntries();
     StrStrMap toolPaths;
-    foreach (ExternalTool *tool, toolsList) {
+    for (auto * const tool : toolsList) {
         SAFE_POINT(tool, "Tool is NULL", );
         addTool(tool);
     }
-    foreach (ExternalTool *tool, toolsList) {
-        QString toolPath = addToolToListsAndReturnToolPath(*tool);
+    for (const auto * const tool : toolsList) {
+        addToolToListsIfNecessary(*tool);
+        const QString toolPath = findToolPath(*tool);
         if (!toolPath.isEmpty()) {
             toolPaths.insert(tool->getId(), toolPath);
         }
@@ -366,11 +373,12 @@ void ExternalToolManagerImpl::sl_customToolImported(const QString &toolId) {
 
     StrStrMap toolPaths;
     addTool(tool);
-    const QString toolPath = addToolToListsAndReturnToolPath(*tool);
+    addToolToListsIfNecessary(*tool);
+    const QString toolPath = findToolPath(*tool);
     if (!toolPath.isEmpty()) {
         toolPaths.insert(tool->getId(), toolPath);
+        validateTools(toolPaths);
     }
-    validateTools(toolPaths);
 }
 
 void ExternalToolManagerImpl::sl_customToolRemoved(const QString &toolId) {
--- a/src/plugins/external_tool_support/src/ExternalToolManager.cpp
+++ b/src/plugins/external_tool_support/src/ExternalToolManager.cpp
@@ -174,10 +174,7 @@ ExternalToolManager::ExternalToolState ExternalToolManagerImpl::getToolState(con
     return toolStates.value(toolId, NotDefined);
 }
 
-QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
-    SAFE_POINT(etRegistry, "The external tool registry is NULL", "");
-    QString toolPath;
-
+void ExternalToolManagerImpl::addTool(ExternalTool *tool) {
     if (tool->isValid()) {
         toolStates.insert(tool->getId(), Valid);
     } else {
@@ -193,28 +190,36 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
         foreach (const QString &dependency, toolDependencies) {
             dependencies.insertMulti(dependency, tool->getId());
         }
+    }
+}
 
-        if (dependenciesAreOk(tool->getId()) && !tool->isValid()) {
-            if (tool->isModule()) {
-                QString masterId = tool->getDependencies().first();
+QString ExternalToolManagerImpl::addToolToListsAndReturnToolPath(const ExternalTool &tool) {
+    QString toolPath;
+    SAFE_POINT(nullptr != etRegistry, "The external tool registry is nullptr", toolPath);
+
+    QStringList toolDependencies = tool.getDependencies();
+    if (!toolDependencies.isEmpty()) {
+        if (dependenciesAreOk(tool.getId()) && !tool.isValid()) {
+            if (tool.isModule()) {
+                QString masterId = tool.getDependencies().first();
                 ExternalTool *masterTool = etRegistry->getById(masterId);
-                SAFE_POINT(masterTool, QString("An external tool '%1' isn't found in the registry").arg(masterId), "");
+                SAFE_POINT(nullptr != masterTool, QString("An external tool '%1' isn't found in the registry").arg(masterId), toolPath);
 
                 toolPath = masterTool->getPath();
             } else {
-                toolPath = tool->getPath();
+                toolPath = tool.getPath();
             }
-            validateList << tool->getId();
+            validateList << tool.getId();
         }
     } else {
-        if (!tool->isValid()) {
-            validateList << tool->getId();
-            toolPath = tool->getPath();
+        if (!tool.isValid()) {
+            validateList << tool.getId();
+            toolPath = tool.getPath();
         }
     }
 
-    if (!validateList.contains(tool->getId()) && !tool->isModule() && !tool->isValid()) {
-        searchList << tool->getId();
+    if (!validateList.contains(tool.getId()) && !tool.isModule() && !tool.isValid()) {
+        searchList << tool.getId();
     }
 
     return toolPath;
@@ -343,7 +348,10 @@ void ExternalToolManagerImpl::sl_customToolsLoaded(Task *task) {
     StrStrMap toolPaths;
     foreach (ExternalTool *tool, toolsList) {
         SAFE_POINT(tool, "Tool is NULL", );
-        QString toolPath = addTool(tool);
+        addTool(tool);
+    }
+    foreach (ExternalTool *tool, toolsList) {
+        QString toolPath = addToolToListsAndReturnToolPath(*tool);
         if (!toolPath.isEmpty()) {
             toolPaths.insert(tool->getId(), toolPath);
         }
@@ -357,7 +365,8 @@ void ExternalToolManagerImpl::sl_customToolImported(const QString &toolId) {
     SAFE_POINT(nullptr != tool, "Tool is nullptr", );
 
     StrStrMap toolPaths;
-    const QString toolPath = addTool(tool);
+    addTool(tool);
+    const QString toolPath = addToolToListsAndReturnToolPath(*tool);
     if (!toolPath.isEmpty()) {
         toolPaths.insert(tool->getId(), toolPath);
     }
@@ -409,7 +418,7 @@ void ExternalToolManagerImpl::validateTools(const StrStrMap &toolPaths, External
             }
         }
 
-        ExternalToolValidateTask *task;
+        ExternalToolValidateTask *task = nullptr;
         QString toolName = AppContext::getExternalToolRegistry()->getToolNameById(toolId);
         if (pathSpecified) {
             task = new ExternalToolJustValidateTask(toolId, toolName, toolPath);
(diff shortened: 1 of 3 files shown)


src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -193,28 +190,36 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
...
+    SAFE_POINT(nullptr != etRegistry, "The external tool registry is nullptr", toolPath);
+
+    QStringList toolDependencies = tool.getDependencies();
+    if (!toolDependencies.isEmpty()) {
+        if (dependenciesAreOk(tool.getId()) && !tool.isValid()) {
+            if (tool.isModule()) {
+                QString masterId = tool.getDependencies().first();
                 ExternalTool *masterTool = etRegistry->getById(masterId);
-                SAFE_POINT(masterTool, QString("An external tool '%1' isn't found in the registry").arg(masterId), "");
+                SAFE_POINT(nullptr != masterTool, QString("An external tool '%1' isn't found in the registry").arg(masterId), toolPath);
```
yalgaer: Do not use Yoda style checks in UGENE.
This style makes your code less readable and does not add extra safety: assignments in expression will break UGENE build.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -193,28 +190,36 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
...
         }
+    }
+}
 
-        if (dependenciesAreOk(tool->getId()) && !tool->isValid()) {
-            if (tool->isModule()) {
-                QString masterId = tool->getDependencies().first();
+QString ExternalToolManagerImpl::addToolToListsAndReturnToolPath(const ExternalTool &tool) {
+    QString toolPath;
+    SAFE_POINT(nullptr != etRegistry, "The external tool registry is nullptr", toolPath);
```
yalgaer: Fix Yoda, see the comment below

src/plugins/external_tool_support/src/ExternalToolManager.h
```diff
@@ -80,8 +80,9 @@ private slots:
     void innerStart();
     void checkStartupTasksState();
     void markStartupCheckAsFinished();
-    QString addTool(ExternalTool *tool);
+    void addTool(ExternalTool *tool);
     bool dependenciesAreOk(const QString &toolId);
+    QString addToolToListsAndReturnToolPath(const ExternalTool &tool);
```
yalgaer: Add comments for every new method created

src/plugins/external_tool_support/src/python/PythonSupport.cpp
```diff
@@ -34,6 +34,12 @@
 
 namespace U2 {
 
+namespace {
+
+const QString PYTHON_VERSION_REGEXP = "(\\d+.\\d+.\\d+)";
```
yalgaer: 6 lines of code for a constant that can be a local variable. Since this constant is used inside of the 1 method you can move it into that method or you can use static/#define to make it 1-liner.

yalgaer (changes requested):

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -193,28 +190,36 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
...
+                toolPath = tool.getPath();
             }
-            validateList << tool->getId();
+            validateList << tool.getId();
         }
     } else {
-        if (!tool->isValid()) {
-            validateList << tool->getId();
-            toolPath = tool->getPath();
+        if (!tool.isValid()) {
```
yalgaer: what if tool.isValid == true ?

The logic of how all states are covered is not clear and may have missed states.
To make it simpler I would suggest is to extract the code that searches for the master tool path into a separate method and focus the code that all possible states (tool.isValid|dependenciesOk|...) are covered and none is missed.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -343,7 +348,10 @@ void ExternalToolManagerImpl::sl_customToolsLoaded(Task *task) {
     StrStrMap toolPaths;
     foreach (ExternalTool *tool, toolsList) {
         SAFE_POINT(tool, "Tool is NULL", );
-        QString toolPath = addTool(tool);
+        addTool(tool);
+    }
+    foreach (ExternalTool *tool, toolsList) {
```
yalgaer: foreach is deprecated. Use for-loop for the new code

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -343,7 +348,10 @@ void ExternalToolManagerImpl::sl_customToolsLoaded(Task *task) {
     StrStrMap toolPaths;
     foreach (ExternalTool *tool, toolsList) {
         SAFE_POINT(tool, "Tool is NULL", );
-        QString toolPath = addTool(tool);
+        addTool(tool);
```
yalgaer: The safe point above does not look right. Can the registry contain nullptrs?

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -343,7 +348,10 @@ void ExternalToolManagerImpl::sl_customToolsLoaded(Task *task) {
...
+    }
+    foreach (ExternalTool *tool, toolsList) {
+        QString toolPath = addToolToListsAndReturnToolPath(*tool);
```
yalgaer: Here you do not have the SAFE_POINT like in the loop above.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -357,7 +365,8 @@ void ExternalToolManagerImpl::sl_customToolImported(const QString &toolId) {
     SAFE_POINT(nullptr != tool, "Tool is nullptr", );
 
     StrStrMap toolPaths;
-    const QString toolPath = addTool(tool);
+    addTool(tool);
+    const QString toolPath = addToolToListsAndReturnToolPath(*tool);
```
yalgaer: move the validateTools call inside the if (). Otherwise, the map is empty

src/plugins/external_tool_support/src/ExternalToolManager.h
evelinabis: Do we have a codestyle? Where can I find it? Where should the private class method comment be: on the same line or on the line above? Does it have to be in a special format (like Doxygen or your custom, not usual "//")?

yalgaer: There is no strict code style. I use:
`
/** Comment text */
`

src/plugins/external_tool_support/src/ExternalToolManager.cpp
atiunov: You have a pointer. There is no need to dereference it.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
atiunov: The registry isn't supposed to store nullptrs. Actually, the application will crash if you try to add a nullptr to the registry, but it is an inner behavior of the registry and this class shouldn't rely on it.

src/plugins/external_tool_support/src/python/PythonSupport.cpp
evelinabis: `#define` for constants is bad.

src/plugins/external_tool_support/src/ExternalToolManager.h
atiunov: Our codestyle is defined with .clang-format file. If some situation is not defined, feel free to do as you want.

atiunov (changes requested):

src/plugins/external_tool_support/src/ExternalToolManager.cpp
evelinabis: Better build failure than unexpected behavior

atiunov: Assignment in expression won't cause build failure. Only a warning will be added that could be skipped easily. Yoda style adds some extra safety.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
evelinabis: 1. If `toolsList` contains `nullptr`, `SAFE_POINT` will be triggered and we will return from function
2. ` toolsList` does not change in above loop
3. Then `nullptr` cannot be inserted in the `toolsList`
4. This is why `SAFE_POINT` is redundant in the second loop

evelinabis: - `void foo(int *i);` -- `i` may be nullptr, but
- `void foo(int &i);` -- means that `i` not null. This allows me not to write `SAFE_POINT` inside the function `foo`. I prefer this type signature for my fiunction.

`*tool` doesn't make function call because it is built-in indirection operator. It's cheap (I don't know for sure, these are my assumptions)

src/plugins/external_tool_support/src/ExternalToolManager.cpp
yalgaer: I'm not sure what will happen on Windows, but the code like 
`if (a=2) `
will break Linux build and we will see it in the Teamcity
I think there is the same option for the Windows compiler.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -193,28 +190,42 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
...
-                SAFE_POINT(masterTool, QString("An external tool '%1' isn't found in the registry").arg(masterId), "");
+                SAFE_POINT(nullptr != masterTool, QString("An external tool '%1' isn't found in the registry").arg(masterId), toolPath);
 
                 toolPath = masterTool->getPath();
             } else {
-                toolPath = tool->getPath();
+                toolPath = tool.getPath();
             }
-            validateList << tool->getId();
+        } else {
```
yalgaer: Empty else?

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -193,28 +190,42 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
...
         }
+    } else if (!tool.isValid()) {
+        toolPath = tool.getPath();
     } else {
```
yalgaer, line 221: 1 more empty else.
If you want to express something with a code like this please add a comment

src/plugins/external_tool_support/src/ExternalToolManager.cpp
yalgaer: So SAFE_POINT in line 356  should be removed?

src/plugins/external_tool_support/src/ExternalToolManager.h
```diff
@@ -80,8 +80,12 @@ private slots:
     void innerStart();
     void checkStartupTasksState();
     void markStartupCheckAsFinished();
-    QString addTool(ExternalTool *tool);
+    void addTool(ExternalTool *tool);
     bool dependenciesAreOk(const QString &toolId);
+    void addToolToListsIfNecessary(const ExternalTool &tool);  // add toolId to validateList if tool is not valid and
+                                                               // its dependencies are ok, and to searchList if needed
+    QString findToolPath(const ExternalTool &tool); // return toolPath (or masterToolPath, if tool is module) if tool is
```
yalgaer: This is a very bad style of a comment. 

Please put it on top of the method and document how addTool() is different from addToolToLists.

src/plugins/external_tool_support/src/python/PythonSupport.cpp
yalgaer: #define is a tool and can't be good or bad. Its usages can be good or bad. 

There is nothing bad when you use #define for local constants for primitive types.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
yalgaer: toolsList is obtained from etRegistry->getAllEntries() which must never have nulls.
This is the reason why I asked about unnecessary safepoints in the code above.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
atiunov: It's a paranoid style. Let's leave it alone.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
atiunov: I've checked it on macOS with clang, there was a warning, but UGENE was built successfully. If you want to treat this case as error feel free to edit ugene_global.pro appropriately. Until then Yoda style is acceptable. At the same time, I won't insist on using it everywhere.
Also, I can't say that non-Yoda styled code is more readable. I just can't see any difference.

yalgaer: Ok, agreed, lets leave it. (If you use Mac for builds I would ask you to make this warning processed as an error, same as on Linux)

src/plugins/external_tool_support/src/ExternalToolManager.cpp
yalgaer: To me it is not a paranoid style, but a situation when a developer is not sure about the code behavior. 
A developer should always know what to expect from the code he changes and unnecessary SAFE_POINT is the signal of misunderstanding.

src/plugins/external_tool_support/src/ExternalToolManager.h
yalgaer: I would also suggest to differentiate the naming convention for addTool & addToolToLists
1) Rename addTool to registerTool and add a comment that the method registers external tool in the registry of tools.
2) Rename addToolToListsIfNecessary to addToolToSearchOrValidateList - the name itself can serve as a good comment of what is going on here.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
```diff
@@ -193,28 +190,42 @@ QString ExternalToolManagerImpl::addTool(ExternalTool *tool) {
...
+
+    if (!validateList.contains(tool.getId()) && !tool.isModule() && !tool.isValid()) {
+        searchList << tool.getId();
+    }
+}
 
-        if (dependenciesAreOk(tool->getId()) && !tool->isValid()) {
-            if (tool->isModule()) {
-                QString masterId = tool->getDependencies().first();
+QString ExternalToolManagerImpl::findToolPath(const ExternalTool &tool) {
```
yalgaer, line 210: The method has very unclear contract: it returns empty value for valid tools.
This is the reason I ask many questions about if () else parts: it looks like not all variants are covered.

src/plugins/external_tool_support/src/ExternalToolManager.cpp
evelinabis: I don't understand, how can I show that all states are covered and have no missed states (https://github.com/ugeneunipro/ugene/pull/152#pullrequestreview-535011600). Can I remove empty else branches? And you and other developers will believe that the block is missing for a reason.

src/plugins/external_tool_support/src/ExternalToolManager.h
evelinabis: But `addTool` does not register the external tool to `etRegistry`. It adds the tool to `toolStates` and `dependencies`. Maybe I need to rename `addTool` to `addToToolStatesAndDependencies`?

yalgaer: Yes this level of details in the method is better than a plain addTool

src/plugins/external_tool_support/src/ExternalToolManager.cpp
evelinabis: Yes, it returns an empty string if tool is valid, and it should be. And all variants are covered (see empty else).

src/plugins/external_tool_support/src/ExternalToolManager.cpp
yalgaer: Once you will fix the method contract or name for 'findToolPath' it will be much easier to understand why in some cases you return the tool path and in some reason an empty string  (empty if).

yalgaer (approved):

src/plugins/external_tool_support/src/ExternalToolManager.cpp
evelinabis: Now I understand why it was pointless

PR #154, opened 2020-11-24
UGENE-6916. Keep MSA sequence names unique on copy/paste and add-from…

yalgaer:


src/corelibs/U2Core/src/util/MSAUtils.cpp
```diff
@@ -507,4 +507,14 @@ void MSAUtils::removeColumnsWithGaps(MultipleSequenceAlignment &msa, int require
     }
 }
 
+QString MSAUtils::rollMsaRowName(const QString &rowName, const QSet<QString> &usedRowNamesSet, const QString &suffixSeparator) {
```
DmitriiSukhomlinov, line 510: The similar function `GUrlUtils::rollFileName` already exists. This one could be removed

DmitriiSukhomlinov (changes requested):
Fix the rolling func or tell me why I'm wrong

src/corelibs/U2Core/src/util/MSAUtils.cpp
yalgaer: These 2 functions are very different semantically and internally. 

* rollFileName: checks that the file with the result name does not exist and the result name is a valid file name.
* rollMsaRowName: has nothing to do with files but must provide a valid MSA row name.

DmitriiSukhomlinov (approved):

PR #156, opened 2020-11-30
UGENE-6904: UGENE hangs on start on macOS Big Sur

atiunov:
Details of the issue reasons are here: [Qt application gets stuck trying to open main window under Big Sur](https://bugreports.qt.io/browse/QTBUG-87014)

### DIFF
--- a/src/ugeneui/src/Main.cpp
+++ b/src/ugeneui/src/Main.cpp
@@ -420,6 +420,11 @@ int main(int argc, char **argv) {
     fixMacFonts();
 #endif
 
+ #ifdef Q_OS_MACOS
+    // A workaround for https://bugreports.qt.io/browse/QTBUG-87014: "Qt application gets stuck trying to open main window under Big Sur"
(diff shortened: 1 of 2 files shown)


src/ugeneui/src/Main.cpp
```diff
@@ -420,6 +420,11 @@ int main(int argc, char **argv) {
     fixMacFonts();
 #endif
 
+ #ifdef Q_OS_MACOS
```
DmitriiSukhomlinov, line 423: Replace with `Q_OS_MACX`

DmitriiSukhomlinov (approved):

src/ugeneui/src/Main.cpp
atiunov: No, Q_OS_MACX is already removed from Qt, Q_OS_MACOS is a correct macro.

atiunov:
I'll merge it manually.

PR #160, opened 2020-12-22
UGENE-4419. "Simple format" and "GenBank/EMBL format" can't apply some values from each other.

evelinabis:
1. Annotations with a location like 200..100 in a Simple format are
   available for creation in the "Create Annotation" widget. This
   location is equivalent to (200..len,1..100) in Genbank format.
   Changes to the Simple format are instantly displayed in the Genbank
   format, changes in the Genbank LineEdit are displayed in the Simple
   format when the focus is lost (for example, pressing a Tab). The
   Complement is supported correctly, the circularity of the sequence is
   not taken into account.
2. Corrected GUI-tests taking into account the new behavior of the
   "Create Annotation" widget.
3. (Not by task) Fixed memory leaks in
   CreateAnnotationWidgetController and RangeSelector.


src/corelibs/U2Gui/src/util/CreateAnnotationFullWidget.cpp
```diff
@@ -33,9 +33,10 @@
 
 namespace U2 {
 
-CreateAnnotationFullWidget::CreateAnnotationFullWidget(QWidget *parent)
+CreateAnnotationFullWidget::CreateAnnotationFullWidget(const qint64 seqLen, QWidget *parent)
     : CreateAnnotationWidget(parent),
-      formatType(Simple) {
+      formatType(Simple),
+      seqLen {seqLen} {
```
yalgaer: nit: we usually use a simpler form of initializers across the codebase: seqLen(seqLen)

src/corelibs/U2Gui/src/util/CreateAnnotationWidgetController.cpp
```diff
@@ -128,6 +128,10 @@ CreateAnnotationWidgetController::CreateAnnotationWidgetController(const CreateA
     connect(occ, SIGNAL(si_comboBoxChanged()), SLOT(sl_documentsComboUpdated()));
 }
 
+CreateAnnotationWidgetController::~CreateAnnotationWidgetController() {
+    delete w;
```
yalgaer: nit: if there is a way to handle resources deallocation with some kind of smart pointer it is better to use the smart-pointer, not a manual 'delete'

src/corelibs/U2Gui/src/util/RangeSelector.cpp
```diff
@@ -244,6 +244,7 @@ MultipleRangeSelector::MultipleRangeSelector(QWidget *_parent, const QVector<U2R
 }
 
 MultipleRangeSelector::~MultipleRangeSelector() {
+    delete ui->multipleRegionEdit->validator();
```
yalgaer, line 247: Do we need to delete validators for other line-edits too?

src/plugins/GUITestBase/src/tests/common_scenarios/annotations/GTTestsCreateAnnotationWidget.cpp
```diff
@@ -1638,13 +1638,13 @@ GUI_TEST_CLASS_DEFINITION(test_0027) {
 
     GTUtilsAnnotationsTreeView::deleteItem(os, "Misc. Feature  (0, 1)");
 
-    //    5. Call "Create new annotation" dialog.
+    //    6. Call "Create new annotation" dialog.
```
yalgaer, line 1641: nit: numbers like these do not scale: you have to always update them to keep the order. 
I would suggest not to use them at all. 
It is only important to identify fail message location when the test fails - fail messages should be unique and numbers (even random) help to make them unique.

src/corelibs/U2Gui/src/util/CreateAnnotationFullWidget.cpp
```diff
@@ -153,19 +154,31 @@ void CreateAnnotationFullWidget::setAnnotationName(const QString &name) {
 }
 
 void CreateAnnotationFullWidget::setLocation(const U2Location &location) {
+    // Example: (200..len,1..100)=(200..100), lambda returns true
+    auto locationSplitted = [](const U2Location &location, qint64 seqLen) {
+        return location->regions[0].endPos() == seqLen && location->regions[1].startPos == 0;
```
yalgaer, line 163: nit: add location.regions.size() == 2 to make this function safe regardless of a caller context.
I would also change the name to isSimpleSplitLocation()

req: change 'splitted'  to 'split'

yalgaer (commented):
I found that the "Complement" switch in the dialog works incorrectly
Please check the video

https://user-images.githubusercontent.com/17791039/102909775-0d1cff80-44ac-11eb-9c9c-202435c80330.mp4

src/corelibs/U2Gui/src/util/RangeSelector.cpp
evelinabis: No. This validator has no parent and is not always deleted https://github.com/ugeneunipro/ugene/blob/050a6d88b515c5b3fbe25b3cc93c7d5d0a13b9b0/src/corelibs/U2Gui/src/util/RangeSelector.cpp#L347

src/corelibs/U2Gui/src/util/CreateAnnotationFullWidget.cpp
```diff
@@ -153,19 +154,38 @@ void CreateAnnotationFullWidget::setAnnotationName(const QString &name) {
 }
 
 void CreateAnnotationFullWidget::setLocation(const U2Location &location) {
+    // Example: (200..len,1..100)=(200..100), lambda returns true
+    auto isSimpleSplitLocation = [](const U2Location &location, qint64 seqLen) {
```
DmitriiSukhomlinov: You've captured variables by value (cuz this is the default behavior), but you have `const U2Location &location` - const reference, which make no sense in this case. It's better to capture variables by reference.

src/corelibs/U2Gui/src/util/CreateAnnotationFullWidget.h
```diff
@@ -30,7 +30,7 @@ namespace U2 {
 class CreateAnnotationFullWidget : public CreateAnnotationWidget, private Ui_CreateAnnotationFullWidget {
     Q_OBJECT
 public:
-    CreateAnnotationFullWidget(QWidget *parent = NULL);
+    CreateAnnotationFullWidget(qint64 seqLen, QWidget *parent = NULL);
```
DmitriiSukhomlinov: `nullptr`, as far as you are here

src/corelibs/U2Gui/src/util/CreateAnnotationWidgetController.cpp
```diff
@@ -120,14 +120,17 @@ CreateAnnotationWidgetController::CreateAnnotationWidgetController(const CreateA
...
-    connect(w, SIGNAL(si_usePatternNamesStateChanged()), SLOT(sl_usePatternNamesStateChanged()));
+    connect(w.data(), SIGNAL(si_selectExistingTableRequest()), SLOT(sl_onLoadObjectsClicked()));
+    connect(w.data(), SIGNAL(si_selectGroupNameMenuRequest()), SLOT(sl_groupName()));
+    connect(w.data(), SIGNAL(si_groupNameEdited()), SLOT(sl_groupNameEdited()));
+    connect(w.data(), SIGNAL(si_annotationNameEdited()), SLOT(sl_annotationNameEdited()));
+    connect(w.data(), SIGNAL(si_usePatternNamesStateChanged()), SLOT(sl_usePatternNamesStateChanged()));
     connect(occ, SIGNAL(si_comboBoxChanged()), SLOT(sl_documentsComboUpdated()));
 }
 
+CreateAnnotationWidgetController::~CreateAnnotationWidgetController() {
```
DmitriiSukhomlinov, line 131: Remove the empty destructor

src/corelibs/U2Gui/src/util/CreateAnnotationWidgetController.h
```diff
@@ -149,12 +150,14 @@ private slots:
...
     GObjectComboBoxController *occ;
-    CreateAnnotationWidget *w;
+    QScopedPointer<CreateAnnotationWidget> w;
     SaveDocumentController *saveController;
 
     QString GROUP_NAME_AUTO;
     static const QString DESCRIPTION_QUALIFIER_KEY;
     static const QString SETTINGS_LASTDIR;
+
+    Q_DISABLE_COPY(CreateAnnotationWidgetController)
```
DmitriiSukhomlinov, line 162: OK, nice, you've forbidden the "copy" behavior, but what's about "move"? There's a good macro `Q_DISABLE_COPY_MOVE`, but it's available starting with Qt 5.13.0 and we use 5.12.9. I suggest you deleting the "move" constructor and the "move" `operator=`.

yalgaer: nit: We are still in Qt 5.4 compatibility mode.

We can escalate this question on the meeting and increase the version of QT to a newer one, but historically it should be the version supported by the Debian stable, which is as far as I remember is 5.7 today.

yalgaer (changes requested):
I see 80+ tests failed in the Nightly build. Will review later once the code is stabilized.

src/corelibs/U2Gui/src/util/CreateAnnotationWidgetController.cpp
evelinabis: If I remove the empty destructor, the build fails because the default destructor is inline, but to declare a QScopedPointer to a forward declared class, it must be non-inline https://doc.qt.io/qt-5/qscopedpointer.html#forward-declared-pointers

I can leave a comment with a link that this empty destructor cannot be deleted

yalgaer (changes requested):
Please remove from the ignore list test 4419.
As the result I expect to have 10 ignored test in the Nightly build (today we have  11)

For some reasons I see 13 ignored tests in the Nightly build.

yalgaer (changes requested):
All looks good but I found 1 more issue. See the screenshot


If the start position is in the "overflow" range - just do not add it. In this example the correct Genbank string should be 1..50

Please add a test for this case too.

yalgaer (approved):

PR #165, opened 2020-12-23
UGENE-6968. FindEnzymes task creates too many results on large sequen…

yalgaer:


src/corelibs/U2Core/src/globals/AutoAnnotationsSupport.h
```diff
@@ -98,6 +98,12 @@ private slots:
 
 #define AUTO_ANNOTATION_SETTINGS "auto-annotations/"
 
+/**
+ * Maximum number of annotations UGENE can handle safely.
+ * Auto-annotation tasks are recommended to avoid producing more annotation than given number to avoid UI freeze.
+ */
+#define AUTO_ANNOTATION_MAX_ANNOTATIONS_ADV_CAN_HANDLE (150 * 1000)
```
DmitriiSukhomlinov, line 106: 1. Please, use `static constexpr int AUTO_ANNOTATION_MAX_ANNOTATIONS_ADV_CAN_HANDLE  = 150 * 1000` instead of the ancient `#define`
2. Where did you get this number from? I mean, why this is the maximum?

src/plugins/enzymes/src/FindEnzymesTask.cpp
```diff
@@ -321,6 +321,11 @@ void FindSingleEnzymeTask::cleanup() {
     resultList.clear();
 }
 
+qint64 FindSingleEnzymeTask::estimateNumberOfEnzymesInSequence(qint64 sequenceLength, int variants) {
+    // The rough estimation from the tests: 1 enzymes can be found 5 times per 1000bp in both direct and complement strands.
```
DmitriiSukhomlinov: And again, where did you get this estimation from? What tests?

src/plugins/enzymes/src/FindEnzymesTask.h
```diff
@@ -59,19 +59,34 @@ class FindEnzymesTask;
 
 struct FindEnzymesTaskConfig {
     FindEnzymesTaskConfig()
-        : maxResults(0x7FFFFFFF),
+        : maxResults(INT_MAX),
```
DmitriiSukhomlinov, line 62: Use `std::numeric_limits<int>::max()` instead of this ancient stuff

src/plugins/enzymes/src/FindEnzymesTask.h
```diff
@@ -59,19 +59,34 @@ class FindEnzymesTask;
...
           minHitCount(1),
           maxHitCount(INT_MAX),
```
DmitriiSukhomlinov, line 64: And here

DmitriiSukhomlinov (changes requested):
Fix concerns and answer questions, please

src/plugins/enzymes/src/FindEnzymesTask.h
yalgaer: Same here

src/plugins/enzymes/src/FindEnzymesTask.h
yalgaer: I would like to keep standard INT_MAX for easier readability when possible.
limits<int>::max() return INT_MAX internally and both are a part of the standard.

PR #166, opened 2020-12-24
UGENE-6971. De novo workflows are invalid

yalgaer:


src/plugins/external_tool_support/src/spades/SpadesSupport.cpp
```diff
@@ -61,19 +61,7 @@ SpadesSupport::SpadesSupport()
...
-/** SPAdes is available only on Linux and Mac. */
-static bool isSupported() {
-#if defined(Q_OS_LINUX) || defined(Q_OS_MAC)
-    return true;
-#else
-    return false;
-#endif
-}
-
 void SpadesSupport::checkIn() {
```
DmitriiSukhomlinov, line 64: It's not "checkIn" now, it's "registerTool"

DmitriiSukhomlinov (approved):
А minor comment, approved

PR #167, opened 2020-12-24
UGENE-6826. The task scheduler crashes when it's running out of threads

yalgaer:


yalgaer:
Won't be applied. Will re-create from scratch.

PR #168, opened 2020-12-24
UGENE-6826. The task scheduler crashs when it's running out of threads

yalgaer:


src/corelibs/U2Private/src/TaskSchedulerImpl.cpp
```diff
@@ -210,7 +210,7 @@ bool TaskSchedulerImpl::processFinishedTasks() {
 
         Task *task = ti->task;
         priorityQueue.removeAt(i);
-        if (task->hasFlags(TaskFlag_RunMessageLoopOnly)) {
+        if (task->hasFlags(TaskFlag_RunMessageLoopOnly) && ti->thread != nullptr) {
```
DmitriiSukhomlinov, line 215: Don't like the solution. `ti->thread` shouldn't be equal to `nullptr` in any case, if it happens that means, that something goes wrong. The point of the issue was to figure out why `ti->thread` becomes `nullptr` and fix this place. You just have inserted the stopper here to avoid the problem. 
@atiunov what do you think about it?

DmitriiSukhomlinov (changes requested):
A major question

src/corelibs/U2Private/src/TaskSchedulerImpl.cpp
yalgaer: Any task that requires a thread may not get one if it has an error or is canceled before the thread allocation stage.

PR #170, opened 2020-12-27
UGENE-4209. Align to reference worker crashes

yalgaer:


src/corelibs/U2Lang/src/support/WorkflowUtils.cpp
```diff
@@ -1211,12 +1174,11 @@ bool WorkflowUtils::validateOutputFile(const QString &url, NotificationsList &no
         fi.setFile(QDir(WorkflowSettings::getWorkflowOutputDirectory()), url);
     }
 
-    if (canWriteToPath(fi.absolutePath())) {
+    if (FileAndDirectoryUtils::isDirectoryWritable(fi.absolutePath())) {
```
atiunov: "FileAndDirectoryUtils::isDirectoryWritable" checks if a dir already exists, but "canWriteToPath" checks that all necessary dirs could be created. There will be troubles if "WorkflowUtils::validateOutputFile" is called before all dirs are created.

src/corelibs/U2Lang/src/support/WorkflowUtils.cpp
```diff
@@ -1229,16 +1191,15 @@ bool WorkflowUtils::validateOutputDir(const QString &url, NotificationsList &not
         fi.setFile(QDir(WorkflowSettings::getWorkflowOutputDirectory()), url);
     }
 
-    if (canWriteToPath(fi.absoluteFilePath())) {
+    if (FileAndDirectoryUtils::isDirectoryWritable(fi.absoluteFilePath())) {
```
atiunov: The same: a new check differs from the old check.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_4001_5000.cpp
```diff
@@ -2056,17 +2056,50 @@ GUI_TEST_CLASS_DEFINITION(test_4194) {
...
+
     GTUtilsWorkflowDesigner::runWorkflow(os);
     GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    // OMG!
+    // The only error we have today is the error about missed chromatogram.
+    // The error is not correct: "Align to Reference" with BLAST does not need/use a chromatogram at all.
+    // The error was introduced during the time the test was suppressed and made impossible to run "Align to Reference" for reads with no chromatograms.
+    // See: UGENE-5423: Use ChromObject in Sanger algorithm.
+    // This problem will be addressed in the separate bug and logTracer will be checked for no errors.
```
atiunov, line 2100: When this test was in ignored state, "Sanger Reads Editor" aka "MCA Editor" was implemented, and "Align to Reference" worker was changed to generate mca objects. "MCA Editor" can't visualize reads without chromatograms now, so the worker was made strict about chromatograms.
I can't find an issue about "MCA Editor" impossibility to operate reads without chromatograms, so I offer you to create a new one and add a relation to [this one](https://ugene.dev/tracker/browse/UGENE-5582).

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.cpp
```diff
@@ -56,21 +56,40 @@ BlastReadsSubTask::BlastReadsSubTask(const QString &dbPath,
...
+    int firstBucketSize = qMin(getMaxParallelSubtasks(), reads.size());
+    for (readIndex = 0; readIndex < firstBucketSize; readIndex++) {
+        const SharedDbiDataHandler &read = reads[readIndex];
+        addSubTask(new BlastAndSwReadTask(dbPath, read, reference, minIdentityPercent, readsNames[read], storage));
+    }
+}
+
+QList<Task *> BlastReadsSubTask::onSubTaskFinished(Task *task) {
+    QList<Task *> newSubtasks;
+    CHECK(!isCanceled(), newSubtasks);
```
atiunov: Check hasError as well. This method doesn't contain any code that should be executed in case of errors.

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.cpp
```diff
@@ -56,21 +56,40 @@ BlastReadsSubTask::BlastReadsSubTask(const QString &dbPath,
...
+
+    BlastAndSwReadTask *blastTask = qobject_cast<BlastAndSwReadTask *>(task);
+    SAFE_POINT(blastTask != nullptr, "Must be a BlastAndSwReadTask", newSubtasks);
```
atiunov: SAFE_POINT must call setError on fail.

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.cpp
```diff
@@ -56,21 +56,40 @@ BlastReadsSubTask::BlastReadsSubTask(const QString &dbPath,
...
+    blastSubTasks << blastTask;
```
atiunov: If a blastTask added to the blustSubTasks list on the blastTask finish, then there could be tasks order violation. It is possible in the following case:
1. The first two tasks are created for the corresponding reads and begin execution.
2. The task for the second read finished before the task for the first read and added to the list with index 0.
3. The task for the first read finished and added to the list with index 1.
So these two tasks have incorrect order in the list.

Tasks should be added to blastSubTasks list right after their creation both in prepare and in onSubTaskFinished methods. Of course, progress calculation should be changed properly.

atiunov (changes requested):

src/corelibs/U2Lang/src/support/WorkflowUtils.cpp
yalgaer: This is correct. Will move the 'canWriteToPath' as an independent function in FileAndDirectoryUtils

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.cpp
yalgaer: Good catch. Will update.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_4001_5000.cpp
yalgaer: AFAIU the main goal of "Align to Reference" WD element is to align short reads to reference using BLAST. In this case MCA Editor is a secondary use-case. 
I will create a new bug as you suggested that will make this scenario work with no errors and MCA editor accepts reads with no chromatograms

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.cpp
yalgaer: Done

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.cpp
yalgaer: Updated the code to preserve the original order.

atiunov (approved):

PR #171, opened 2020-12-28
UGENE-6951. GUI tests for "Extract consensus from assemblies..." main menu item

evelinabis:
4 tests have been written to extract consensus from
1. .bam assembly
2. .bam.ugenedb assembly
3. multiple assemblies
4. invalid input

### DIFF
--- a/src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
+++ b/src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
@@ -26,74 +26,42 @@
 #include <QRegularExpression>
 #include <QStringBuilder>
 #include <QTabWidget>
+#include <QXmlStreamReader>
 
 #include <U2Gui/HoverQLabel.h>
 
 #include "GTUtilsDashboard.h"
-
-namespace {
-
-// A substring with the name of the notification type is searched for. If the |str| contains exactly one type of
-// notification, then this type is returned, otherwise a null QVariant is returned.
-QVariant getNotificationTypeFromString(const QString &str) {
-    using Notification = U2::GTUtilsDashboard::Notification;
-
-    QVariant returnable;
-    constexpr int typesNumber = sizeof(Notification::typesNames) / sizeof(QString);
-    int counter = 0;
-    Notification::Type type;
-
-    for (int typeInd = 0; typeInd < typesNumber; ++typeInd) {
-        if (str.contains(Notification::typesNames[typeInd], Qt::CaseInsensitive)) {
-            ++counter;
-            type = static_cast<Notification::Type>(typeInd);
-        }
-    }
-
-    if (counter == 1) {
-        returnable = static_cast<int>(type);
-    }
-    return returnable;
-};
-
-#define GT_CLASS_NAME "GTUtilsDashboard"
-
-#define GT_METHOD_NAME "getCellText"
-// Returns text from the (row, column) table cell. The (row, column) cell must exist
-QString getNotificationCellText(HI::GUITestOpStatus &os, const QGridLayout &tableLayout, const int row, const int col) {
-    QString text;
-    const QWidget *const cellWidget = tableLayout.itemAtPosition(row, col)->widget();
-
-    if (cellWidget != nullptr && cellWidget->objectName() == "tableCell") {
-        if (const QLayout *const cellLayout = cellWidget->layout()) {
-            for (int labelInd = 0; labelInd < cellLayout->count(); ++labelInd) {
-                if (const auto *const label = qobject_cast<QLabel *>(cellLayout->itemAt(labelInd)->widget())) {
-                    text = label->text();
-                }
-            }
-        }
-    }
-    return text;
-}
-#undef GT_METHOD_NAME
-
-#undef GT_CLASS_NAME
-
-}    // namespace
 
 namespace U2 {
 using namespace HI;
 
 const QString GTUtilsDashboard::TREE_ROOT_ID = ExternalToolsDashboardWidget::TREE_ID;
 
-const QString GTUtilsDashboard::Notification::typesNames[] = {QStringLiteral("INFO"),
-                                                              QStringLiteral("WARNING"),
-                                                              QStringLiteral("ERROR")};
+const QString GTUtilsDashboard::Notification::typeNames[] = {"INFO",
+                                                             "WARNING",
+                                                             "ERROR"};
 
-GTUtilsDashboard::Notification::operator QString() const {
-    const QString lBracket = QStringLiteral("[");
-    const QString rBracket = QStringLiteral("]");
-    return lBracket % typeToQString(type) % rBracket % lBracket % element % rBracket % QStringLiteral(" ") % message;
+QString GTUtilsDashboard::Notification::typeToString(Type t) {
+    return typeNames[static_cast<int>(t)];
+}
+
+QVariant GTUtilsDashboard::Notification::stringToType(const QString &str) {
+    QVariant type;
+    if (str.isEmpty()) {
+        return type;
+    }
+
+    constexpr int typesNumber = sizeof(Notification::typeNames) / sizeof(QString);
+    for (int typeInd = 0; typeInd < typesNumber; typeInd++) {
+        if (QString::compare(str, Notification::typeNames[typeInd], Qt::CaseInsensitive) == 0) {
+            return type = typeInd;
+        }
+    }
+    return type;
+}
+
+QString GTUtilsDashboard::Notification::toString() const {
+    return "[" % typeToString(type) % "][" % element % "] " % message;
 }
 
 #define GT_CLASS_NAME "GTUtilsDashboard"
@@ -207,31 +175,66 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
     return notificationsWidget->isVisible();
 }
 
+#define GT_METHOD_NAME "getNotificationTypeFromHtml"
+GTUtilsDashboard::Notification::Type GTUtilsDashboard::getNotificationTypeFromHtml(HI::GUITestOpStatus &os,
+    const QString &html) {
+    using Notification = U2::GTUtilsDashboard::Notification;
+    const QString errMsg = "Dashboard notification type not found";
+    GT_CHECK_RESULT(!html.isEmpty(), errMsg, Notification::Type::ERR)
+
+    QXmlStreamReader reader(html);
+    while (reader.readNextStartElement()) {
+        if (reader.name() == "img") {
+            const QVariant type = Notification::stringToType(reader.attributes().value("class").toString());
+            if (type.canConvert<int>()) {
+                return static_cast<Notification::Type>(type.toInt());
+            }
+        }
+    }
+    GT_CHECK_RESULT(false, errMsg, Notification::Type::ERR)
+}
+#undef GT_METHOD_NAME
+
+#define GT_METHOD_NAME "getNotificationCellText"
+QString GTUtilsDashboard::getNotificationCellText(HI::GUITestOpStatus &os, const QGridLayout &tableLayout,
+    const int row, const int col) {
+    const QWidget *cellWidget = tableLayout.itemAtPosition(row, col)->widget();
+    QString text;
+
+    if (cellWidget != nullptr && cellWidget->objectName() == "tableCell") {
+        if (const QLayout *const cellLayout = cellWidget->layout()) {
+            for (int labelInd = 0; labelInd < cellLayout->count(); ++labelInd) {
+                if (const auto *const label = qobject_cast<QLabel *>(cellLayout->itemAt(labelInd)->widget())) {
+                    text = label->text();
+                }
+            }
+        }
+    }
+    GT_CHECK_RESULT(!text.isEmpty(),
+        QString("Error getting (%1,%2) cell of dashboard notification table").arg(row).arg(col), text)
+    return text;
+}
+#undef GT_METHOD_NAME
+
 #define GT_METHOD_NAME "getNotifications"
 QList<GTUtilsDashboard::Notification> GTUtilsDashboard::getNotifications(GUITestOpStatus &os) {
-    using Type = Notification::Type;
-
-    const QString notificationsWidgetName = QStringLiteral("NotificationsDashboardWidget");
+    const QString notificationsWidgetName = "NotificationsDashboardWidget";
     const auto tableLayout = qobject_cast<QGridLayout *>(
         GTWidget::findWidget(os, notificationsWidgetName, GTUtilsDashboard::getDashboard(os))->layout());
     QList<Notification> notifications;
 
     GT_CHECK_RESULT(tableLayout != nullptr && tableLayout->columnCount() == 3,
-                    notificationsWidgetName % QStringLiteral(" was found, but cannot be used in a test"),
-                    notifications);
+                    notificationsWidgetName % " was found, but cannot be used in a test",
+                    notifications)
 
     const int notificationsNumber = tableLayout->rowCount();
     notifications.reserve(notificationsNumber);
-    for (int row = 0; row < notificationsNumber; ++row) {
-        const QString typeStr = getNotificationCellText(os, *tableLayout, row, 0);
+    for (int row = 1; row < notificationsNumber; row++) {
+        const Notification::Type type = getNotificationTypeFromHtml(os,
+            getNotificationCellText(os, *tableLayout, row, 0));
         const QString element = getNotificationCellText(os, *tableLayout, row, 1);
         const QString message = getNotificationCellText(os, *tableLayout, row, 2);
-        if (!typeStr.isNull() && !element.isNull() && !message.isNull()) {
-            const QVariant type = getNotificationTypeFromString(typeStr);
-            if (!type.isNull()) {
-                notifications << Notification {static_cast<Type>(type.toInt()), element, message};
-            }
-        }
+        notifications << Notification {type, element, message};
     }
     return notifications;
 }
@@ -240,8 +243,8 @@ QList<GTUtilsDashboard::Notification> GTUtilsDashboard::getNotifications(GUITest
 QString GTUtilsDashboard::getJoinedNotificationsString(GUITestOpStatus &os) {
     const auto notifications = getNotifications(os);
     QStringList stringNotifications;
-    for (const auto &notification : notifications) {
-        stringNotifications << notification;
+    for (const auto &notification : qAsConst(notifications)) {
+        stringNotifications << notification.toString();
     }
     return stringNotifications.join('\n');
 }
--- a/src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
+++ b/src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
@@ -26,7 +26,6 @@
 #include <QRegularExpression>
 #include <QStringBuilder>
 #include <QTabWidget>
-#include <QXmlStreamReader>
 
 #include <U2Gui/HoverQLabel.h>
 
@@ -37,31 +36,8 @@ using namespace HI;
 
 const QString GTUtilsDashboard::TREE_ROOT_ID = ExternalToolsDashboardWidget::TREE_ID;
 
-const QString GTUtilsDashboard::Notification::typeNames[] = {"INFO",
-                                                             "WARNING",
-                                                             "ERROR"};
-
-QString GTUtilsDashboard::Notification::typeToString(Type t) {
-    return typeNames[static_cast<int>(t)];
-}
-
-QVariant GTUtilsDashboard::Notification::stringToType(const QString &str) {
-    QVariant type;
-    if (str.isEmpty()) {
-        return type;
-    }
-
-    constexpr int typesNumber = sizeof(Notification::typeNames) / sizeof(QString);
-    for (int typeInd = 0; typeInd < typesNumber; typeInd++) {
-        if (QString::compare(str, Notification::typeNames[typeInd], Qt::CaseInsensitive) == 0) {
-            return type = typeInd;
-        }
-    }
-    return type;
-}
-
 QString GTUtilsDashboard::Notification::toString() const {
-    return "[" % typeToString(type) % "][" % element % "] " % message;
+    return "[" % type.toUpper() % "][" % element % "] " % message;
 }
 
 #define GT_CLASS_NAME "GTUtilsDashboard"
@@ -176,22 +152,17 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
 }
 
 #define GT_METHOD_NAME "getNotificationTypeFromHtml"
-GTUtilsDashboard::Notification::Type GTUtilsDashboard::getNotificationTypeFromHtml(HI::GUITestOpStatus &os,
-    const QString &html) {
-    using Notification = U2::GTUtilsDashboard::Notification;
-    const QString errMsg = "Dashboard notification type not found";
-    GT_CHECK_RESULT(!html.isEmpty(), errMsg, Notification::Type::ERR)
+QString GTUtilsDashboard::getNotificationTypeFromHtml(HI::GUITestOpStatus &os, const QString &html) {
+    QString type;
 
-    QXmlStreamReader reader(html);
-    while (reader.readNextStartElement()) {
-        if (reader.name() == "img") {
-            const QVariant type = Notification::stringToType(reader.attributes().value("class").toString());
-            if (type.canConvert<int>()) {
-                return static_cast<Notification::Type>(type.toInt());
-            }
-        }
-    }
-    GT_CHECK_RESULT(false, errMsg, Notification::Type::ERR)
+    int start = html.indexOf("<img class=\"");
+    int end = html.indexOf("\"", start + 12); // 12 = length of "<img class=\""
+    GT_CHECK_RESULT(start >= 0 && end >= 0, "Dashboard notification type not found", type)
+
+    start += 12;
+    end -= start;
+    type = html.mid(start, end);
+    return type;
 }
 #undef GT_METHOD_NAME
 
@@ -219,8 +190,9 @@ QString GTUtilsDashboard::getNotificationCellText(HI::GUITestOpStatus &os, const
 #define GT_METHOD_NAME "getNotifications"
 QList<GTUtilsDashboard::Notification> GTUtilsDashboard::getNotifications(GUITestOpStatus &os) {
     const QString notificationsWidgetName = "NotificationsDashboardWidget";
-    const auto tableLayout = qobject_cast<QGridLayout *>(
-        GTWidget::findWidget(os, notificationsWidgetName, GTUtilsDashboard::getDashboard(os))->layout());
+    QWidget *const notificationsWidget = GTWidget::findWidget(os, notificationsWidgetName,
+        GTUtilsDashboard::getDashboard(os));
+    const auto tableLayout = qobject_cast<QGridLayout *>(notificationsWidget->layout());
     QList<Notification> notifications;
 
     GT_CHECK_RESULT(tableLayout != nullptr && tableLayout->columnCount() == 3,
@@ -228,10 +200,8 @@ QList<GTUtilsDashboard::Notification> GTUtilsDashboard::getNotifications(GUITest
                     notifications)
 
     const int notificationsNumber = tableLayout->rowCount();
-    notifications.reserve(notificationsNumber);
     for (int row = 1; row < notificationsNumber; row++) {
-        const Notification::Type type = getNotificationTypeFromHtml(os,
-            getNotificationCellText(os, *tableLayout, row, 0));
+        const QString type = getNotificationTypeFromHtml(os, getNotificationCellText(os, *tableLayout, row, 0));
         const QString element = getNotificationCellText(os, *tableLayout, row, 1);
         const QString message = getNotificationCellText(os, *tableLayout, row, 2);
         notifications << Notification {type, element, message};
--- a/src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
+++ b/src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
@@ -24,16 +24,77 @@
 #include <primitives/GTWidget.h>
 
 #include <QRegularExpression>
+#include <QStringBuilder>
 #include <QTabWidget>
 
 #include <U2Gui/HoverQLabel.h>
 
 #include "GTUtilsDashboard.h"
 
+namespace {
+
+// A substring with the name of the notification type is searched for. If the |str| contains exactly one type of
+// notification, then this type is returned, otherwise a null QVariant is returned.
+QVariant getNotificationTypeFromString(const QString &str) {
+    using Notification = U2::GTUtilsDashboard::Notification;
+
+    QVariant returnable;
+    constexpr int typesNumber = sizeof(Notification::typesNames) / sizeof(QString);
+    int counter = 0;
+    Notification::Type type;
+
+    for (int typeInd = 0; typeInd < typesNumber; ++typeInd) {
+        if (str.contains(Notification::typesNames[typeInd], Qt::CaseInsensitive)) {
+            ++counter;
+            type = static_cast<Notification::Type>(typeInd);
+        }
+    }
+
+    if (counter == 1) {
+        returnable = static_cast<int>(type);
+    }
+    return returnable;
+};
+
+#define GT_CLASS_NAME "GTUtilsDashboard"
+
+#define GT_METHOD_NAME "getCellText"
+// Returns text from the (row, column) table cell. The (row, column) cell must exist
+QString getNotificationCellText(HI::GUITestOpStatus &os, const QGridLayout &tableLayout, const int row, const int col) {
+    QString text;
+    const QWidget *const cellWidget = tableLayout.itemAtPosition(row, col)->widget();
+
+    if (cellWidget != nullptr && cellWidget->objectName() == "tableCell") {
+        if (const QLayout *const cellLayout = cellWidget->layout()) {
+            for (int labelInd = 0; labelInd < cellLayout->count(); ++labelInd) {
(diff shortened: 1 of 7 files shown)


src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+#include "primitives/GTToolbar.h"
+#include "primitives/GTWidget.h"
+#include "runnables/ugene/plugins/workflow_designer/WizardFiller.h"
+#include "utils/GTUtilsDialog.h"
+
+namespace U2 {
+
+namespace GUITest_assembly_extract_consensus {
+
+struct ExtractConsensusWizardScenario : CustomScenario {
```
yalgaer: struct extends class: avoid this mix unless you have a reason for that.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+    enum class Algorithm {
+        DEFAULT,
+        SAMTOOLS
+    };
+
+    QStringList inpPaths;
```
yalgaer: nit: use the full name. Can't guess what "inp" means during the first read (write once, read 100 times)

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+    QString outFile;
+
+    ExtractConsensusWizardScenario(const QStringList &assembliesPath = QStringList(),
+                                   const Algorithm algo = Algorithm::DEFAULT,
+                                   const bool keepGaps = true,
+                                   const QString &outputFileName = "consensus.fa")
+        : inpPaths(assembliesPath), algo(algo), keepGaps(keepGaps), outFile(outputFileName) {
+    }
+
+    void run(HI::GUITestOpStatus &os) {
```
yalgaer: nit: add override when you override.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+        QWidget *const dialog = QApplication::activeModalWidget();
+        CHECK_SET_ERR(dialog != nullptr, "activeModalWidget is nullptr");
+        {
+            auto *const urlleInp = GTWidget::findExactWidget<QLineEdit *>(os, "Assembly widget", dialog);
+            CHECK_SET_ERR(urlleInp != nullptr, "Assembly widget is nullptr");
```
yalgaer: nit: useless CHECK_SER_ERR: findExactWidget asserts for you unless you pass custom FindOptions that allow returning nullptr.
Same issue in many places below.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+            QString inpText;
+            foreach (const QString &inp, inpPaths) {
```
yalgaer: nit: foreach -> for. Same issue in other places below.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+                inpText += QFileInfo(inp).absoluteFilePath();
+                inpText += ";";
+            }
+            if (inpText.length() > 0) {
+                inpText.remove(inpText.length() - 1, 1);
```
yalgaer: nip: add a comment what is going on here. Maybe a more readable solution would be is to collect all paths into QStringList and to join then in the end: list.join(";"); (up to you)

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+    //  3. Click "Run"
+    GTUtilsDialog::waitForDialog(os, new WizardFiller(
+        os, "Extract Consensus Wizard", new ExtractConsensusWizardScenario(QStringList() << path)
+    ));
+    GTMenu::clickMainMenuItem(os, QStringList() << "Tools"
+                                                << "NGS data analysis"
+                                                << "Extract consensus from assemblies...");
+
+    //  4. Wait for workflow finished
+    GTUtilsTaskTreeView::waitTaskFinished(os, 100000);
```
yalgaer: nit: Default is 180*1000. May be you do not need to make an accent on 100*1000

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+        new ExtractConsensusWizardScenario(QStringList() << path, Algo::SAMTOOLS, false, "")
+    ));
+    GTToolbar::clickButtonByTooltipOnToolbar(os, MWTOOLBAR_ACTIVEMDI, "Show wizard");
+
+    //  7. Wait for workflow finished
+    GTUtilsTaskTreeView::waitTaskFinished(os, 100000);
+    //  Expected state: There should be no errors in the log
+    CHECK_SET_ERR(!lt.hasErrors(), "Errors in log: " + lt.getJoinedErrorString());
+    //  There should be no notifications in the dashboard
+    CHECK_SET_ERR(!GTUtilsDashboard::hasNotifications(os), "Unexpected notification");
```
yalgaer: nit: include the notification text in the assertion message to simplify analysis when the error will happen.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+        ExtractConsensusWizardWithAddScenario(const QStringList &assembliesPath,
+                                              const Algorithm algo = Algorithm::DEFAULT,
+                                              const bool keepGaps = true,
+                                              const QString &outputFileName = "consensus.fa")
+            : inpPaths(assembliesPath), algo(algo), keepGaps(keepGaps), outFile(outputFileName) {
+        }
+
+        void run(HI::GUITestOpStatus &os) {
+            QWidget *const dialog = QApplication::activeModalWidget();
```
yalgaer: nit: 
Use GTWidget::getActiveModalWidget.
this method 1) Waits 2) Asserts that the model widget exists.

yalgaer (dismissed):

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,327 @@
...
+                                   const bool keepGaps = true,
+                                   const QString &outputFileName = "consensus.fa")
+        : inpPaths(assembliesPath), algo(algo), keepGaps(keepGaps), outFile(outputFileName) {
+    }
+
+    void run(HI::GUITestOpStatus &os) {
+        QWidget *const dialog = QApplication::activeModalWidget();
+        CHECK_SET_ERR(dialog != nullptr, "activeModalWidget is nullptr");
+        {
+            auto *const urlleInp = GTWidget::findExactWidget<QLineEdit *>(os, "Assembly widget", dialog);
```
DmitriiSukhomlinov: You have no need to write `auto *`, cuz `auto` could contain `*` "inside" itself. And, I know that it's OK, but we have used to text `const` before type

DmitriiSukhomlinov: The same in other places

DmitriiSukhomlinov (dismissed):

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
evelinabis: At the beginning, I thought that I would need to change the fields of the class/structure, in this regard, I did not find a reason to make unnecessary getters and setters, so I needed a structure with its public fields. However, later I realized that I could not reuse the same created object, because it is deleted, and therefore the need for the structure is gone. I'll fix that. Thank you for your comment.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
 
 #include <QRegularExpression>
+#include <QStringBuilder>
 #include <QTabWidget>
 
 #include <U2Gui/HoverQLabel.h>
 
 #include "GTUtilsDashboard.h"
 
+namespace {
```
yalgaer: nit: you will save 4-5 vertical lines of code if you will use static instead of an anonymous namespace. However this is not a blocker, but only a suggestion. You can follow your style if you find it is better

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+
+// A substring with the name of the notification type is searched for. If the |str| contains exactly one type of
+// notification, then this type is returned, otherwise a null QVariant is returned.
+QVariant getNotificationTypeFromString(const QString &str) {
+    using Notification = U2::GTUtilsDashboard::Notification;
+
+    QVariant returnable;
```
yalgaer: nit: normally you should use 'result' or 'notificationType'. It is the first time I see 'returnable' as the variable name. Usually, such names are used for types names.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+    constexpr int typesNumber = sizeof(Notification::typesNames) / sizeof(QString);
+    int counter = 0;
+    Notification::Type type;
+
+    for (int typeInd = 0; typeInd < typesNumber; ++typeInd) {
```
yalgaer: nit: every time you prefer ++i instead of i++ you want to emphasize that the returned value is important. Here it is not: I would suggest using more classic i++;

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+        if (str.contains(Notification::typesNames[typeInd], Qt::CaseInsensitive)) {
```
yalgaer: Use startsWith not contains. Reason: a user text may contain the same text as the notification type name.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+namespace {
+
+// A substring with the name of the notification type is searched for. If the |str| contains exactly one type of
+// notification, then this type is returned, otherwise a null QVariant is returned.
+QVariant getNotificationTypeFromString(const QString &str) {
+    using Notification = U2::GTUtilsDashboard::Notification;
+
+    QVariant returnable;
+    constexpr int typesNumber = sizeof(Notification::typesNames) / sizeof(QString);
+    int counter = 0;
```
yalgaer: if you use startsWith instead of contains you do not need the counter at all

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+        returnable = static_cast<int>(type);
+    }
+    return returnable;
+};
+
+#define GT_CLASS_NAME "GTUtilsDashboard"
+
+#define GT_METHOD_NAME "getCellText"
+// Returns text from the (row, column) table cell. The (row, column) cell must exist
+QString getNotificationCellText(HI::GUITestOpStatus &os, const QGridLayout &tableLayout, const int row, const int col) {
```
yalgaer: class name is missed for the method. QString GTUtilsDashboard::getNotificationCellText

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+    QString text;
+    const QWidget *const cellWidget = tableLayout.itemAtPosition(row, col)->widget();
```
yalgaer: nit: too many local consts make the code less readable.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
+    if (cellWidget != nullptr && cellWidget->objectName() == "tableCell") {
+        if (const QLayout *const cellLayout = cellWidget->layout()) {
+            for (int labelInd = 0; labelInd < cellLayout->count(); ++labelInd) {
+                if (const auto *const label = qobject_cast<QLabel *>(cellLayout->itemAt(labelInd)->widget())) {
+                    text = label->text();
+                }
+            }
+        }
+    }
+    return text;
```
yalgaer: Normally you should fail in the test method if the result is not found/available by default.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -24,17 +24,78 @@
...
 using namespace HI;
 
 const QString GTUtilsDashboard::TREE_ROOT_ID = ExternalToolsDashboardWidget::TREE_ID;
 
+const QString GTUtilsDashboard::Notification::typesNames[] = {QStringLiteral("INFO"),
+                                                              QStringLiteral("WARNING"),
+                                                              QStringLiteral("ERROR")};
+
+GTUtilsDashboard::Notification::operator QString() const {
+    const QString lBracket = QStringLiteral("[");
```
yalgaer: QStringLiterals are normally used to optimize the code. What kind of optimization you want to make here and why?
Use "["  or "]" to make the code more readable.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +207,45 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
     return notificationsWidget->isVisible();
 }
 
+#define GT_METHOD_NAME "getNotifications"
+QList<GTUtilsDashboard::Notification> GTUtilsDashboard::getNotifications(GUITestOpStatus &os) {
+    using Type = Notification::Type;
+
+    const QString notificationsWidgetName = QStringLiteral("NotificationsDashboardWidget");
```
yalgaer: Same here. Do not use string literals without special reason. Use QString notificationWidgetName = "";

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +207,45 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
...
+    const auto tableLayout = qobject_cast<QGridLayout *>(
```
yalgaer: Use findExactWidget so you do not need to check for nullptr.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +207,45 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
...
+                    notificationsWidgetName % QStringLiteral(" was found, but cannot be used in a test"),
+                    notifications);
+
+    const int notificationsNumber = tableLayout->rowCount();
+    notifications.reserve(notificationsNumber);
+    for (int row = 0; row < notificationsNumber; ++row) {
+        const QString typeStr = getNotificationCellText(os, *tableLayout, row, 0);
+        const QString element = getNotificationCellText(os, *tableLayout, row, 1);
+        const QString message = getNotificationCellText(os, *tableLayout, row, 2);
+        if (!typeStr.isNull() && !element.isNull() && !message.isNull()) {
```
yalgaer: use isEmpty instead of isNull for QStrings.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +207,45 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
...
+        }
+    }
+    return notifications;
+}
+#undef GT_METHOD_NAME
+
+QString GTUtilsDashboard::getJoinedNotificationsString(GUITestOpStatus &os) {
+    const auto notifications = getNotifications(os);
+    QStringList stringNotifications;
+    for (const auto &notification : notifications) {
```
yalgaer: use qAsConst in for loops explicitly

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +207,45 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
...
+        stringNotifications << notification;
```
yalgaer: I would avoid operator overloading completely and called toString() here to make the code more readable and less dependable from the external context.

src/plugins/GUITestBase/src/GTUtilsDashboard.h
```diff
@@ -59,7 +81,14 @@ class GTUtilsDashboard {
     static void clickOutputFile(HI::GUITestOpStatus &os, const QString &outputFileName);
 
     static QString getTabObjectName(Tabs tab);
+
     static bool hasNotifications(HI::GUITestOpStatus &os);
+
+    static QList<Notification> getNotifications(HI::GUITestOpStatus &os);
```
yalgaer, line 79: For all new methods always add at least some docs.

src/plugins/GUITestBase/src/GTUtilsDashboard.h
```diff
@@ -59,7 +81,14 @@ class GTUtilsDashboard {
...
+
+    // Returns a QString containing all dashboard notifications, splitted by '\n'
+    static QString getJoinedNotificationsString(HI::GUITestOpStatus &os);
```
yalgaer, line 82: This method is not needed at all: getNotificationList(os).join("\n") can be used instead

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,318 @@
...
+#include "primitives/GTComboBox.h"
+#include "primitives/GTLineEdit.h"
+#include "primitives/GTMenu.h"
+#include "primitives/GTToolbar.h"
+#include "primitives/GTWidget.h"
+#include "runnables/ugene/plugins/workflow_designer/WizardFiller.h"
+#include "utils/GTUtilsDialog.h"
+
+namespace {
+enum class Algo {
```
yalgaer: Add docs for all new methods and types

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,318 @@
...
+                                   const bool keepGaps = true,
+                                   const QString &outputFileName = "consensus.fa")
+        : inputPaths(assembliesPath), algo(algo), keepGaps(keepGaps), outFile(outputFileName) {
+    }
+
+    void run(HI::GUITestOpStatus &os) override {
+        QWidget *const dialog = GTWidget::getActiveModalWidget(os);
+
+        // Dialog filling
+        std::transform(inputPaths.begin(), inputPaths.end(), inputPaths.begin(), [](const QString &path) {
```
yalgaer: Use a simple indexed 'for'  loop - it will also take 3 lines but will be more readable/have fewer characters

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,318 @@
...
+    //     And click "Run"
+    GTUtilsDialog::waitForDialog(os, new WizardFiller(
+        os,
+        "Extract Consensus Wizard",
+        new ExtractConsensusWizardScenario(QStringList() << path, Algo::SAMTOOLS, false, "")
+    ));
+    GTToolbar::clickButtonByTooltipOnToolbar(os, MWTOOLBAR_ACTIVEMDI, "Show wizard");
+
+    //  7. Wait for workflow finished
+    GTUtilsTaskTreeView::waitTaskFinished(os, 180000);
```
yalgaer: 2 timeouts by 3 minutes in the same test.
Nightly tests have timeout equal to 4 minutes in total.
Consider moving this test to the Metagenomics suite or reduce timeouts or test data to make the test faster.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,318 @@
...
```
yalgaer: Same here: 2 timeouts by 3 minutes

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,318 @@
...
+}
+
+GUI_TEST_CLASS_DEFINITION(wrong_input_test_0001) {
+    const GTLogTracer lt;
+    const QString path1 = dataDir + "samples/Assembly/chrM.fa";
+    const QString path2 = dataDir + "samples/Assembly/chrM.sorted.bam";
+    const QString dashboardErrMsg = "Unsupported document format: ";
+
+    const auto hasDashboardNotification = [](HI::GUITestOpStatus &os, const QString &errMsg) {
+        return !GTWidget::findLabelByText(os, errMsg, GTWidget::findWidget(
```
yalgaer: If you will cache results into the local variables you still will have only 3 lines. Split these lines into simpler independent statements.

yalgaer (changes requested):

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
evelinabis: str is string like `"<center></center>"`, where a type is written in the png file name

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
yalgaer: It's just a style used across other code in the UGENE. You will find i++ to ++i as 100/1

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
yalgaer: I see. In this case add 'class="warning|error|info"' marker to the notification HTML or use the existing class from the style-sheet (if this this available). 
In either case notification widget should have 100% info about notification type - no need to guess.

src/plugins/GUITestBase/src/GTUtilsDashboard.h
```diff
@@ -39,6 +39,28 @@ class GTUtilsDashboard {
...
                 ExternalTools };
 
+    // Dashboard notification class. Contains the type of notification (info/warning/error), the name of the workflow
+    // element to which the notification relates and the notification message
+    struct Notification {
+        enum class Type { INFO,
+                          WARN,
+                          ERR };
+
+        static const QString typesNames[3];
```
yalgaer: Rename to typeNames or typeNameList. typesNames (when both words are plural) is not correct.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
evelinabis: It was a static function, not a class method. I will make this function a static method of the class

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
yalgaer: Same here: you introduce very verbose style that is new and unusual to UGENE.
When it comes to local variables in a simple 10-line method 2 consts in variable type declaration makes the code less readable.

I see the following simple solutions:
- Want to avoid such comments: follow the common style found across other files in UGENE.
- Want to make this style to be a standard: write a email with a suggestion to the team alias before you use it.

In either case this comment is a 'nit' - not a blocker for the final approval

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
EvelinaBiserova: I have changed my mind. I'll just remove the macros

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
evelinabis: It won't work here

EvelinaBiserova: It won't work here

src/plugins/GUITestBase/src/GTUtilsDashboard.h
evelinabis: What is `getNotificationList()`?

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
evelinabis: I don't understand what you are talking about

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -35,6 +37,33 @@ using namespace HI;
...
+
+QVariant GTUtilsDashboard::Notification::stringToType(const QString &str) {
+    QVariant type;
+    if (str.isEmpty()) {
+        return type;
+    }
+
+    constexpr int typesNumber = sizeof(Notification::typeNames) / sizeof(QString);
+    for (int typeInd = 0; typeInd < typesNumber; typeInd++) {
+        if (QString::compare(str, Notification::typeNames[typeInd], Qt::CaseInsensitive) == 0) {
```
yalgaer: What is the reason to define typeName values in the upper-case and next compare them case insensitive with notification types which are lower-case?

I would suggest to use lowercase map and replace typeNames array with QStringList.
In this case your method can be as simple as 1-line of code.
 1. typeNames.indexOf(str).

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +175,80 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
...
 }
 
+#define GT_METHOD_NAME "getNotificationTypeFromHtml"
+GTUtilsDashboard::Notification::Type GTUtilsDashboard::getNotificationTypeFromHtml(HI::GUITestOpStatus &os,
+    const QString &html) {
+    using Notification = U2::GTUtilsDashboard::Notification;
+    const QString errMsg = "Dashboard notification type not found";
+    GT_CHECK_RESULT(!html.isEmpty(), errMsg, Notification::Type::ERR)
+
+    QXmlStreamReader reader(html);
```
yalgaer: I see big overdesign and non-needed complexity. Instead of using SAX parser (or DOM model) just locate the info type in the string like: html.indexOf(""<center><img class=\""")...
As the result this method may be 1-2 lines long.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
```diff
@@ -146,6 +175,80 @@ bool GTUtilsDashboard::hasNotifications(HI::GUITestOpStatus &os) {
...
+    const auto tableLayout = qobject_cast<QGridLayout *>(
+        GTWidget::findWidget(os, notificationsWidgetName, GTUtilsDashboard::getDashboard(os))->layout());
+    QList<Notification> notifications;
+
+    GT_CHECK_RESULT(tableLayout != nullptr && tableLayout->columnCount() == 3,
+                    notificationsWidgetName % " was found, but cannot be used in a test",
+                    notifications)
+
+    const int notificationsNumber = tableLayout->rowCount();
+    notifications.reserve(notificationsNumber);
```
yalgaer: please remove optimization with reserve() unless you have a proof that this optimization is needed

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
yalgaer: Ok, could you provide more details please why findExact won't work here?

src/plugins/GUITestBase/src/GTUtilsDashboard.h
```diff
@@ -39,6 +39,32 @@ class GTUtilsDashboard {
                 Input,
                 ExternalTools };
 
+    // Dashboard notification class. Contains the type of notification (info/warning/error), the name of the workflow
+    // element to which the notification relates and the notification message
+    struct Notification {
+        // Type of notification
+        enum class Type { INFO,
```
yalgaer: Declaring Notification.Type as enum in the test package + writing a framework how to convert it to string & back looks like an over-design.

More than that the "type" used in the U2Designer code as a pure QString.

I would suggest to remove this enum and use plain old QString for the "type" in tests.
For the known set of types we can have a static named constants: static QString Notification::warning = "warning".

src/plugins/GUITestBase/src/GTUtilsDashboard.h
```diff
@@ -39,6 +39,32 @@ class GTUtilsDashboard {
...
+
+        static const QString typeNames[3];
+
+        // Returns the notification type |t| as a string
+        static QString typeToString(Type t);
+        // Returns null QVariant if the string |str| isn't the name of the notification type, and the notification type
+        // if it is
+        static QVariant stringToType(const QString &str);
+
+        Notification() = delete;
```
yalgaer: Please document.

src/plugins/GUITestBase/src/GTUtilsDashboard.h
```diff
@@ -39,6 +39,32 @@ class GTUtilsDashboard {
...
+
+        Type type;
+        QString element;
```
yalgaer: Please document all new methods and fields.

src/plugins/GUITestBase/src/GTUtilsDashboard.h
yalgaer: I meant getNotifications() method. But you are right, there is no join(",") method there.

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
yalgaer: Please cache GTWidget::findWidget() result into a separate variable to make the lines 171-173 more readable. I won't increase number of lines you have.

yalgaer (changes requested):
Please schedule a "Nightly" run for this branch to check the new test results before approval.

src/plugins/GUITestBase/src/GTUtilsDashboard.cpp
evelinabis: QLayout is not an inheritor of the QWidget class

I realized my mistake. I will make the code more readable

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/extract_consensus/GTTestsAssemblyExtractConsensus.cpp
```diff
@@ -0,0 +1,218 @@
...
+    }
+};
+
+GUI_TEST_CLASS_DEFINITION(test_0001_single_input) {
+    const GTLogTracer lt;
+
+    //  1. Select "Tools->NGS data analysis->Extract consensus from assemblies..."
+    //  2. Set "samples/Assembly/chrM.sorted.bam" as an input
+    //  3. Click "Run"
+    auto *const bamScenario =
```
DmitriiSukhomlinov: `auto` can define pointer by itself

DmitriiSukhomlinov (dismissed):

yalgaer (commented):
The code looks OK, I have no other issues to fix. Please run Linux Nightly and when it shows now regressions I will approve.

yalgaer (approved):

PR #181, opened 2021-01-01
UGENE-6984. Remove deprecated 'qint64 coordToPos(int x)' method in GS…

yalgaer:


src/corelibs/U2View/src/ov_sequence/Overview.cpp
```diff
@@ -48,7 +48,7 @@ const QString Overview::ANNOTATION_GRAPH_STATE = "sequenceViewSettings/annotatio
 Overview::Overview(ADVSingleSequenceWidget *p, ADVSequenceObjectContext *ctx)
     : GSequenceLineView(p, ctx),
       seqWidget(p) {
-    renderArea = new OverviewRenderArea(this);
+    renderArea = overviewRenderArea = new OverviewRenderArea(this);
```
atiunov: Please, do not do several assignments on one line. It is all about style, so I don't have good arguments, but I don't like it.

yalgaer: Ok, will fix in the next commit inside this PR.
BTW I would prefer to have both these fields *const ptrs but can't find a good way to write it - it can only be done in the constructor initializers.

src/plugins/circular_view/src/CircularView.cpp
```diff
@@ -67,7 +67,7 @@ CircularView::CircularView(QWidget *p, ADVSequenceObjectContext *ctx, CircularVi
         registerAnnotations(obj->getAnnotations());
     }
 
-    renderArea = new CircularViewRenderArea(this);
+    renderArea = circularViewRenderArea = new CircularViewRenderArea(this);
```
atiunov: The same.

src/corelibs/U2View/src/ov_sequence/Overview.cpp
yalgaer: Done

src/plugins/circular_view/src/CircularView.cpp
yalgaer: Done

src/plugins/circular_view/src/CircularView.h
```diff
@@ -45,30 +45,31 @@
...
+    void mouseReleaseEvent(QMouseEvent *e) override;
+    void resizeEvent(QResizeEvent *e) override;
+
+    void keyPressEvent(QKeyEvent *e) override;
+    void keyReleaseEvent(QKeyEvent *e) override;
 
-    void wheelEvent(QWheelEvent *we);
-    virtual QSize sizeHint() const;
+    void wheelEvent(QWheelEvent *we) override;
+    virtual QSize sizeHint() const override;
```
atiunov: Usually you remove "virtual" word in that cases.

yalgaer: Done

atiunov (approved):

PR #183, opened 2021-01-01
UGENE-6985. Support exclude-by-label option for GUI Test Launcher

yalgaer:


src/corelibs/U2Test/src/gui_tests/UGUITestBase.cpp
```diff
@@ -78,15 +78,36 @@ QMap<QString, GUITest *> &UGUITestBase::getMap(TestType testType) {
...
+ */
+static bool intersects(const QSet<QString> &set1, const QSet<QString> &set2) {
+    for (const QString &value1 : set1) {
+        if (set2.contains(value1)) {
+            return true;
+        }
+    }
+    return false;
+}
+
```
atiunov, line 93: I'm not sure what minimum Qt version is supposed to be supported now. In any case, even in Qt5.4 you can use ```set1.intersect(set2).isEmpty()``` to check if there is an intersection.

yalgaer: QSet::intersect modifies the current set and is not suitable here.
https://doc.qt.io/qt-5/qset.html#intersect

atiunov (approved):

PR #184, opened 2021-01-02
UGENE-6987. DetViewMultiLineRenderer::getAnnotationYRange does not co…

yalgaer:
Notes: 
In the next CR, I plan to remove y-range related methods from the virtual interface: they won't be required from all views anymore.
The DetView method will be kept but maybe re-located to the tests package because we do not need it in runtime.

This CR is to unblock tests and to review the internal method code.

### DIFF
--- a/src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
+++ b/src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
@@ -60,7 +60,7 @@ float DetViewMultiLineRenderer::posToXCoordF(const qint64 p, const QSize &canvas
 }
 
 U2Region DetViewMultiLineRenderer::getAnnotationYRange(Annotation *annotation, int locationRegionIndex, const AnnotationSettings *annotationSettings, const QSize &canvasSize, const U2Region &visibleRange) const {
-    SAFE_POINT(locationRegionIndex < annotation->getRegions().length(), "Invalid locationRegionIndex", U2Region());
+    SAFE_POINT(locationRegionIndex >= 0 && locationRegionIndex < annotation->getRegions().length(), "Invalid locationRegionIndex", U2Region());
 
     // Compute region offset within a single line.
     U2Region yRegion = singleLineRenderer->getAnnotationYRange(annotation, locationRegionIndex, annotationSettings, canvasSize, visibleRange);
--- a/src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
+++ b/src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
@@ -37,17 +37,17 @@ const int DetViewMultiLineRenderer::INDENT_BETWEEN_LINES = 30;
 DetViewMultiLineRenderer::DetViewMultiLineRenderer(DetView *detView, SequenceObjectContext *ctx)
     : DetViewRenderer(detView, ctx),
       extraIndent(0) {
-    singleLinePainter = new DetViewSingleLineRenderer(detView, ctx);
+    singleLineRenderer = new DetViewSingleLineRenderer(detView, ctx);
 }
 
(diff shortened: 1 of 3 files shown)


src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
```diff
@@ -59,13 +59,25 @@ float DetViewMultiLineRenderer::posToXCoordF(const qint64 p, const QSize &canvas
...
-U2Region DetViewMultiLineRenderer::getAnnotationYRange(Annotation *a, int r, const AnnotationSettings *as, const QSize &canvasSize, const U2Region &visibleRange) const {
-    if (qgetenv(ENV_GUI_TEST).toInt() == 1) {
-        U2Region res = singleLinePainter->getAnnotationYRange(a, r, as, QSize(canvasSize.width(), getOneLineHeight()), visibleRange);
-        res.startPos += INDENT_BETWEEN_LINES / 2;
-        return res;
+U2Region DetViewMultiLineRenderer::getAnnotationYRange(Annotation *annotation, int locationRegionIndex, const AnnotationSettings *annotationSettings, const QSize &canvasSize, const U2Region &visibleRange) const {
+    SAFE_POINT(locationRegionIndex < annotation->getRegions().length(), "Invalid locationRegionIndex", U2Region());
+
+    // Compute region offset within a single line.
+    U2Region yRegion = singleLineRenderer->getAnnotationYRange(annotation, locationRegionIndex, annotationSettings, canvasSize, visibleRange);
```
atiunov, line 66: Single line canvasSize should be passed to the method. It's height is less.

atiunov, line 66: Single line canvasSize should be passed to the method. It's height is less.

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
```diff
@@ -59,13 +59,25 @@ float DetViewMultiLineRenderer::posToXCoordF(const qint64 p, const QSize &canvas
     return commonMetrics.charWidth * (p % symbolsPerLine);
 }
 
-U2Region DetViewMultiLineRenderer::getAnnotationYRange(Annotation *a, int r, const AnnotationSettings *as, const QSize &canvasSize, const U2Region &visibleRange) const {
-    if (qgetenv(ENV_GUI_TEST).toInt() == 1) {
-        U2Region res = singleLinePainter->getAnnotationYRange(a, r, as, QSize(canvasSize.width(), getOneLineHeight()), visibleRange);
-        res.startPos += INDENT_BETWEEN_LINES / 2;
-        return res;
+U2Region DetViewMultiLineRenderer::getAnnotationYRange(Annotation *annotation, int locationRegionIndex, const AnnotationSettings *annotationSettings, const QSize &canvasSize, const U2Region &visibleRange) const {
+    SAFE_POINT(locationRegionIndex < annotation->getRegions().length(), "Invalid locationRegionIndex", U2Region());
```
atiunov: Check that locationRegionIndex is greater than 0.

atiunov (changes requested):

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
yalgaer: I agree and I plan to remove this method completely in the next CR (see bug description)
Right now I can't change its signature because this is a virtual method used in multiple places.

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewMultiLineRenderer.cpp
yalgaer: Fixed.

atiunov (approved):

PR #187, opened 2021-01-04
UGENE-6989. Remove getAnnotationYRange from the common base

yalgaer:


src/corelibs/U2View/src/ov_sequence/GSequenceLineViewAnnotated.cpp
```diff
@@ -164,52 +164,9 @@ bool GSequenceLineViewAnnotated::isAnnotationVisible(const Annotation *a) const
 }
 
 QList<Annotation *> GSequenceLineViewAnnotated::findAnnotationsByCoord(const QPoint &coord) const {
-    QList<Annotation *> resultAnnotationList;
```
yalgaer, line 167: This method was moved to the render-area with no changes.

src/plugins/circular_view/src/CircularView.cpp
```diff
@@ -188,33 +188,6 @@ void CircularView::sl_onDNASelectionChanged(LRegionsSelection *thiz, const QVect
     renderArea->update();
 }
 
-QList<Annotation *> CircularView::findAnnotationsByCoord(const QPoint &coord) const {
```
yalgaer, line 191: This method was moved to the render area with no changes.

atiunov (approved):

PR #191, opened 2021-01-07
UGENE-6993. Do not rely on UI elements order in combobox in MsaEditor…

yalgaer:


src/corelibs/U2View/src/ov_msa/sort/MsaEditorSortSequencesWidget.cpp
```diff
@@ -71,10 +71,9 @@ MsaEditorSortSequencesWidget::MsaEditorSortSequencesWidget(QWidget *parent, MSAE
 }
 
 void MsaEditorSortSequencesWidget::sl_sortClicked() {
-    int sortType = sortByCombo->currentData().toInt();
-    SAFE_POINT(sortType >= MultipleAlignment::SortByName && sortType <= MultipleAlignment::SortByLeadingGap, "Invalid sort type: " + QString::number(sortType), );
-    MultipleAlignment::Order sortOrder = sortOrderCombo->currentIndex() == 0 ? MultipleAlignment::Ascending : MultipleAlignment::Descending;
-    msaEditor->sortSequences((MultipleAlignment::SortType)sortType, sortOrder);
+    auto sortType = (MultipleAlignment::SortType)sortByCombo->currentData().toInt();
+    auto sortOrder = (MultipleAlignment::Order)sortOrderCombo->currentData().toInt();
```
atiunov, line 75: I would prefer QVariant::value<enum_type>() instead of C-style types conversion.

atiunov (changes requested):

src/corelibs/U2View/src/ov_msa/sort/MsaEditorSortSequencesWidget.cpp
yalgaer: auto sortType = sortByCombo->currentData().value<MultipleAlignment::SortType>(); 
requires meta-type registration in the application level global scope which is overkill for such a local change.

Which meta types to register and when can be a scope of a separate global CL which will review every usage of QVariant in UGENE.

atiunov (approved):

PR #196, opened 2021-01-13
UGENE-6948 GUI tests

DmitriiSukhomlinov:


src/corelibs/U2Test/src/gui_tests/UGUITest.cpp
```diff
@@ -159,6 +159,7 @@ const QString UGUITest::screenshotDir = getScreenshotDir();
 
 const QString UGUITestLabels::Nightly = "Nightly";
 const QString UGUITestLabels::Metagenomics = "Metagenonics";
+const QString UGUITestLabels::Niaid = "NIAID";
```
atiunov: I offer to name it "Cistrome". These workflows were implemented in co-op with NIAID, but  nobody remembers it.

atiunov (changes requested):

src/corelibs/U2Test/src/gui_tests/UGUITest.cpp
DmitriiSukhomlinov: Done

atiunov (approved):

PR #197, opened 2021-01-15
UGENE-7001. Add keyboard shortcut to Go To Selected Read action in MC…

yalgaer:


src/corelibs/U2View/src/ov_msa/McaEditor.h
```diff
@@ -72,6 +72,10 @@ class U2VIEW_EXPORT McaEditor : public MaEditor {
 
     SequenceObjectContext *getReferenceContext() const;
 
+    QAction *getGotoSelectedReadAction() const {
```
DmitriiSukhomlinov, line 75: Call me picky, but I think, it's better either keep all functions in ".cpp" files, either have some kind of functions (for example, short ones, or "getters" and "setters"), which are all placed in ".h" files. In this case, all getters above are in ".cpp", but this one is here, so it's a bit confused.

src/corelibs/U2View/src/ov_msa/McaEditorNameList.cpp
```diff
@@ -126,4 +128,18 @@ int McaEditorNameList::getIconColumnWidth() const {
     return iconColumnWidth;
 }
 
+void McaEditorNameList::keyPressEvent(QKeyEvent *e) {
+    const int key = e->key();
+    const Qt::KeyboardModifiers modifiers = e->modifiers();
+    if (modifiers == Qt::NoModifier && (key == Qt::Key_Space || key == Qt::Key_Enter || key == Qt::Key_Return)) {
```
DmitriiSukhomlinov, line 134: There is a special function for this - `QAction::setShortcuts`. I think, it's better to use it.

src/corelibs/U2View/src/ov_msa/McaEditor.h
yalgaer: Trivial getters can be placed either to the header or source files: we have no strict code style about it.

src/corelibs/U2View/src/ov_msa/McaEditorNameList.cpp
yalgaer: This is not what I need: the shortcut is active in the name-list only. 

When the focus is inside another element this shortcut does not work: keyboard navigation inside other MCA components is not defined today at all and is out of the scope of current PR.

DmitriiSukhomlinov: And I still don't see any contradictions. You may use `QAction::setShortcuts` and `QAction::setShortcutContext(Qt::WidgetShortcut)` to set the shortcut available for the certain widget only.

atiunov: There won't be a shortcut hint in context menu in this case. Nobody will know about the shortcut.
But I can't offer anything that would be definitely better.

yalgaer: This can work but I see 2 problems:
1) The action is defined in the MCA editor globally, while the shortcut I want to have is valid only for MCA name list. It means that it is logically incorrect to define this shortcut globally.

2) I use 3 different keys to do the same: Space, Enter, Return. Can your solution handle all of them?

src/plugins/GUITestBase/src/tests/common_scenarios/mca_editor/GTTestsMcaEditor.cpp
```diff
@@ -3770,6 +3771,31 @@ GUI_TEST_CLASS_DEFINITION(test_0041) {
     CHECK_SET_ERR("1036" == readLengthString, QString("15. Unexpected read length label: expected '%1', got '%2'").arg("1036").arg(readLengthString));
 }
 
+GUI_TEST_CLASS_DEFINITION(test_0042) {
+    // Open an MCA object.
+    GTFileDialog::openFile(os, testDir + "_common_data/sanger", "alignment_short.ugenedb");
```
atiunov, line 3776: The file will be modified if you open it. It is not critical, if the test doesn't do any significant modifications, but it is better to make the test isolated. Copy the file to a sandbox and work with the copy.

atiunov (changes requested):

src/corelibs/U2View/src/ov_msa/McaEditorNameList.cpp
DmitriiSukhomlinov: 1. Why not? If you want this action for MCANameList only, it has to be created in MCANameList. Otherwise, the behavior is correct.
2. `QAction::setShortcuts(const QList<QKeySequence> &shortcuts)` allows you to set several shortcuts.

src/plugins/GUITestBase/src/tests/common_scenarios/mca_editor/GTTestsMcaEditor.cpp
yalgaer: Our tests are safe for modifications. It is a part of the test suite logic: it runs rsync before each test to reset all data dirs.
Otherwise, any crashed test could affect all later ones.

BTW how re-positioning in MCA viewer can affect MCA data?

src/corelibs/U2View/src/ov_msa/McaEditorNameList.cpp
yalgaer: 1. This action is defined in MCA editor globally. Custom name-list shortcut is a customization of the action by the MCA name list widget. The action can be added to the context/tools menu, toolbar, activated using another user action (mouse click) and in this case, there is no need for a global shortcut.

2. Got it. Thanks

DmitriiSukhomlinov: I don't have more arguments about "1". I have presented my point of view, you may follow it or not, I won't insist.

src/plugins/GUITestBase/src/tests/common_scenarios/mca_editor/GTTestsMcaEditor.cpp
atiunov: Ok, rsync is enough.
Currently re-positioning can't affect MCA data, so I wrote that it is not critical. But for example we could add position saving to the database to open on the same position next time.

src/corelibs/U2View/src/ov_msa/McaEditorNameList.cpp
yalgaer: I think your suggestion is 100% valid but requires global shortcuts model design for MCA editor.

While this model is not ready any trivial keyboard shortcut like 'Space' or 'Enter' which has a high collision chance with other behaviors (like "edit sequence" where Space adds gaps) should be used in a very limited way.

atiunov (approved):

PR #199, opened 2021-01-17
UGENE-7005. Make MA editor minimum font size be aware about current U…

yalgaer:


atiunov (dismissed):

atiunov (changes requested):
Add a test for MCA Editor.

yalgaer:
> Add a test for MCA Editor.

Added

atiunov (approved):

PR #204, opened 2021-01-23
UGENE-7016. Tree branch expand/collapse state is not reflected in the…

yalgaer:
I also renamed/added detailed docs for the collapsing mode actions.

The goal of the renaming is to emphasize that this mode may be used by different use-cases, not only grouping by sequence content. There are no changes in the renamed code. The only changes are on the Tree-MSA boundary.


DmitriiSukhomlinov:
I really don't like refactoring here, it'll be difficult to split meaningful and useful (for this exact issue) code in the future. It's better to create separate issues for refactoring purposes.

DmitriiSukhomlinov (approved):

PR #206, opened 2021-01-25
UGENE-7010

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7010


src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/ExportMSA2MSADialogFiller.cpp
```diff
@@ -54,6 +56,9 @@ void ExportMSA2MSADialogFiller::commonScenario() {
         QComboBox *formatCombo = dialog->findChild<QComboBox *>("formatCombo");
```
yalgaer: Use GTWidget::findExactWidget. It will assert if the child element is not found.

src/plugins/dna_export/src/ExportTasks.h
```diff
@@ -90,6 +96,7 @@ class ExportMSA2MSATask : public DocumentProviderTask {
     QString url;
     QString format;
     QList<DNATranslation *> aminoTranslations;    // amino translation for a sequences in alignment. If not NULL -> sequence is translated
+    bool trimGaps;
```
yalgaer: Please document any new methods/fields added.

src/plugins/GUITestBase/src/tests/common_scenarios/msa_editor/GTTestsMsaEditor.cpp
```diff
@@ -4615,6 +4615,32 @@ GUI_TEST_CLASS_DEFINITION(test_0090) {
...
+    // 4. Copy to clipboard
+    GTUtilsMSAEditorSequenceArea::selectArea(os);
+    GTUtilsDialog::waitForDialog(os, new PopupChooser(os, QStringList() << "MSAE_MENU_COPY"
+                                                                        << "copy_selection"));
+    GTMouseDriver::click(Qt::RightButton);
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    // Expected: TAVS\nXXVS
+    const QString clipboardText = GTClipboard::sequences(os);
+    const QString expectedMSA = "TAVS\nXXVS";
```
yalgaer, line 4640: Why have you selected 'X' option instead of '-' variant? 
I do not know which is better here and unless you have a strong opinion that 'X' is better than '-' I would leave this decision to a user: ask user if to translate incomplete triplets to 'X' or to '-'.

One of the sources for a 'strong opinion' would be a reference to some existing tool that prefers one over another.

src/plugins/GUITestBase/src/tests/common_scenarios/msa_editor/GTTestsMsaEditor.cpp
```diff
@@ -4615,6 +4615,32 @@ GUI_TEST_CLASS_DEFINITION(test_0090) {
...
+
+    // 2. Do document context menu {Export->Export aligniment to amino format}
+    // 3. Translate with "Include gaps"
+    GTUtilsDialog::waitForDialog(os, new ExportMSA2MSADialogFiller(os, -1, sandBoxDir + "GUITest_common_scenarios_msa_editor_test_0091.aln", true));
+    GTUtilsDialog::waitForDialog(os, new PopupChooser(os, QStringList() << MSAE_MENU_EXPORT << "amino_translation_of_alignment_rows"));
+    GTWidget::click(os, GTUtilsMdi::activeWindow(os), Qt::RightButton);
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    // 4. Copy to clipboard
+    GTUtilsMSAEditorSequenceArea::selectArea(os);
```
yalgaer, line 4632: nit: this normally should be a 1-liner, like getWholeData()..

yalgaer (changes requested):
Everything looks good, except the final solution remains unclear: why 'X' over  '-' or the opposite.

If we have no good examples in other tools I would suggest to stay non opinionated and allow a user to make the decision.

yalgaer:
++ Could you please add everyone from the team to the list of the reviewers by default.

src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/ExportMSA2MSADialogFiller.cpp
DmitriiSukhomlinov: Done

src/plugins/GUITestBase/src/tests/common_scenarios/msa_editor/GTTestsMsaEditor.cpp
DmitriiSukhomlinov: Sorry, don't get the idea

src/plugins/GUITestBase/src/tests/common_scenarios/msa_editor/GTTestsMsaEditor.cpp
DmitriiSukhomlinov: I have added a possibility to choose

DmitriiSukhomlinov:
Have fixed all comments, please, review again.

src/plugins/GUITestBase/src/tests/common_scenarios/msa_editor/GTTestsMsaEditor.cpp
yalgaer: If you want to check the sequence area content you can use a single line to access it: getWholeData. No need to copy/check clipboard for that. That will keep the test code minimal and focused in the key subject.
This is not a blocker anyway - only a suggestion.

src/plugins/dna_export/src/ExportMSA2MSADialog.h
```diff
@@ -35,9 +35,16 @@ class ExportMSA2MSADialog : public QDialog, Ui_ExportMSA2MSADialog {
 public:
     ExportMSA2MSADialog(const QString &defaultFileName, const DocumentFormatId &f, bool wholeAlignmentOnly, QWidget *p);
 
+    enum class UnknownAmino {
```
yalgaer, line 42: Please document any new types/methods/fields.

I would suggest to use "above the subject" doc style like /** some text */  because it will never go to the line below and automatically shown by IDEs. 

The suffix-style comments like 
int fieldX; // fieldX is lorem ipsum...
are good only if they are short (single line) and are never shown by IDE as contextual docs. I think they are good for the inner comments, but not for the top-level once.

src/plugins/dna_export/src/ExportMSA2MSADialog.ui
```diff
@@ -61,6 +61,51 @@
...
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QGroupBox" name="gbTranslateAs">
+     <property name="enabled">
+      <bool>false</bool>
+     </property>
+     <property name="title">
+      <string>Translate unknown amino acid as:</string>
```
yalgaer: I would not say "translate amino acid" because you translate "nucleic acids" here.
I would rephrase it to "Translate triplets with gaps as:"

src/plugins/dna_export/src/ExportTasks.cpp
```diff
@@ -179,6 +193,9 @@ void ExportMSA2MSATask::run() {
             aminoTT->translate(seq.constData(), seq.length(), resseq.data(), resseq.length());
 
             resseq.replace("*", "X");
+            if (!trimGaps && convertUnknownToGap) {
+                resseq.replace("X", "-");
```
yalgaer, line 196: This is still not correct. 

Here is 2 reasons why:
1) We can't replace all X with gaps, because you replace stop codons (valid nucleic triplets, non-gaps) with gaps.
The simplest solution I see here is to move "resseq.replace("*", "X");" below your replacement code

2) There is a chance that some incomplete DNA triplets are always converted to the same amino. 
For example: Proline, Serine, etc... may have "any" base in the end position and still be translated into a valid protein.
https://miro.medium.com/max/1440/0*YVhxcZ55DqPBDdGF.png

This way if in the original DNA alignment 'gap' means 'unknown', the translation still should contain a valid amino acid, not a gap. 

I'm OK if the item 2) is fixed as a separate bug/PR.

src/plugins/dna_export/src/ExportTasks.h
```diff
@@ -90,6 +97,8 @@ class ExportMSA2MSATask : public DocumentProviderTask {
     QString url;
     QString format;
     QList<DNATranslation *> aminoTranslations;    // amino translation for a sequences in alignment. If not NULL -> sequence is translated
+    bool trimGaps;                                // trim gaps before translation of not
```
yalgaer: nit: if something is not expected to be changed inside of the class it can be declared as a 'const'

yalgaer (changes requested):
The last, but not the least: could you please run Linux Nighly for the PR, so we know that it does not break any existing behavior and the tests from the PR are green?

src/plugins/dna_export/src/ExportMSA2MSADialog.h
DmitriiSukhomlinov: Done

src/plugins/dna_export/src/ExportMSA2MSADialog.ui
DmitriiSukhomlinov: Done

src/plugins/dna_export/src/ExportTasks.cpp
DmitriiSukhomlinov: 1. Fixed
2. I understand the idea, but, actually, don't get a sequence, which could be a part of alignment and affected by this problem. Could you help me?

src/plugins/dna_export/src/ExportTasks.h
DmitriiSukhomlinov: Done

src/plugins/GUITestBase/src/tests/common_scenarios/msa_editor/GTTestsMsaEditor.cpp
DmitriiSukhomlinov: I took a note, but didn't fix it this time

DmitriiSukhomlinov:
@yalgaer fixed, check again

src/plugins/dna_export/src/ExportTasks.cpp
yalgaer: Ok, I will create a separate issue.

yalgaer:
The PR is ready for the approval. No code related issues left. Just run the Linux/Nightly in the Teamcity for your branch to show no regressions added by this commit.
You can put your build in front of the automatically scheduled builds.

DmitriiSukhomlinov:
Have passed [here](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewType.html?buildTypeId=bt153&branch_UgeneNightlyBuilds_GuiTests=UGENE-7010&tab=buildTypeStatusDiv). @yalgaer, approve, please.

yalgaer (approved):

PR #207, opened 2021-01-26
UGENE-6970. Assert when extract consensus from chrM.sam in debug mode

yalgaer:


src/corelibs/U2View/src/ov_assembly/AssemblyModel.cpp
```diff
@@ -175,40 +169,47 @@ U2Region AssemblyModel::getGlobalRegion() {
...
-            }
-            // ignore incorrect attribute value and remove corrupted attribute (auto-fix incorrectly converted ugenedb)
-            if (cachedModelLength == 0) {
+    if (cachedModelLength != NO_VAL) {
+        return cachedModelLength;
+    }
+    // Try to set 'cachedModelLength' from attributes.
+    QSet<U2DbiFeature> dbiFeatureSet = dbiHandle.dbi->getFeatures();
+    bool canReadAttributes = dbiFeatureSet.contains(U2DbiFeature_ReadAttributes);
+    bool canWriteAttributes = dbiFeatureSet.contains(U2DbiFeature_WriteAttributes);
```
DmitriiSukhomlinov, line 178: I can offer texting something in the log if we don't have write permission.

src/corelibs/U2View/src/ov_assembly/AssemblyModel.cpp
```diff
@@ -175,40 +169,47 @@ U2Region AssemblyModel::getGlobalRegion() {
...
+    U2AttributeDbi *attributeDbi = canReadAttributes ? dbiHandle.dbi->getAttributeDbi() : nullptr;
+    if (attributeDbi != nullptr) {
+        U2IntegerAttribute attr = U2AttributeUtils::findIntegerAttribute(attributeDbi, assembly.id, U2BaseAttributeName::reference_length, os);
+        LOG_OP(os);
+        if (attr.hasValidId()) {
+            cachedModelLength = attr.value;
+            // Ignore incorrect attribute value and remove the corrupted attribute: auto-fix incorrectly converted ugenedb.
+            if (cachedModelLength <= 0) {
                 coreLog.details(QString("ignored incorrect value of attribute %1: should be > 0, got %2. Bad attribute removed!").arg(QString(U2BaseAttributeName::reference_length)).arg(cachedModelLength));
```
DmitriiSukhomlinov: Please, translate it as soon as we are here

DmitriiSukhomlinov (dismissed):

src/corelibs/U2View/src/ov_assembly/AssemblyModel.cpp
yalgaer: This should be a valid case. For example, the assembly can be in the remote (public) database.

src/corelibs/U2View/src/ov_assembly/AssemblyModel.cpp
yalgaer: Fixed

PR #208, opened 2021-01-26
UGENE-6958. 3D struct is not rendered on Windows

evelinabis:
Some Windows systems have OpenGL 1.1 installed by default (this is a very old version of OpenGL, currently not supported) and supported DirectX. On such computers, it is possible to render 3D graphics using Qt, but at the moment, 3D models are not rendered in UGENE for unknown reasons.
Added a check that disables the plugin if there are rendering errors. This check is added to deprecated and non-deprecated code

### DIFF
--- a/src/plugins/biostruct3d_view/src/deprecated/BioStruct3DGLWidget.cpp
+++ b/src/plugins/biostruct3d_view/src/deprecated/BioStruct3DGLWidget.cpp
@@ -82,9 +82,12 @@ void BioStruct3DGLWidget::tryGL() {
 bool BioStruct3DGLWidget::canRender() {
     QGLWidget w;
     w.makeCurrent();
-    bool canRender = glGetError() == GL_NO_ERROR;
+
+    GLenum error = glGetError();
+    bool canRender = error == GL_NO_ERROR;
     if (!canRender) {
-        coreLog.error(tr("BioStruct3DView plugin has been added, but cannot be used"));
+        coreLog.error(tr("The \"3D Structure Viewer\" was disabled, because OpenGL has error ") +
+            QString("(%1): %2").arg(error).arg(reinterpret_cast<const char *>(gluErrorString(error))));
     }
     return canRender;
 }
--- a/src/plugins/biostruct3d_view/src/deprecated/BioStruct3DGLWidget.cpp
+++ b/src/plugins/biostruct3d_view/src/deprecated/BioStruct3DGLWidget.cpp
@@ -82,7 +82,11 @@ void BioStruct3DGLWidget::tryGL() {
 bool BioStruct3DGLWidget::canRender() {
     QGLWidget w;
     w.makeCurrent();
-    return glGetError() == GL_NO_ERROR;
+    bool canRender = glGetError() == GL_NO_ERROR;
+    if (!canRender) {
+        coreLog.error(tr("BioStruct3DView plugin has been added, but cannot be used"));
+    }
+    return canRender;
 }
 
 static QColor DEFAULT_BACKGROUND_COLOR = Qt::black;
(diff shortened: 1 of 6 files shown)


src/plugins/biostruct3d_view/src/deprecated/BioStruct3DViewPlugin.cpp
```diff
@@ -93,8 +93,12 @@ extern "C" Q_DECL_EXPORT QString *U2_PLUGIN_FAIL_MASSAGE_FUNC() {
 BioStruct3DViewPlugin::BioStruct3DViewPlugin()
     : Plugin(tr("3D Structure Viewer"), tr("Visualizes 3D structures of biological molecules.")) {
     // Init plugin view context
-    viewContext = new BioStruct3DViewContext(this);
-    viewContext->init();
+    if (BioStruct3DGLWidget::canRender()) {
```
evelinabis, line 96: For reviewers: should this check be enabled only on Windows systems and not on Mac or Linux?

evelinabis: It is worth testing this code on Mac and Linux

yalgaer: Could you please run Linux Nighly for your branch to see that non of the existing 3D-related tests are broken? (you will need to do it anyway to submit any CL)

src/plugins/biostruct3d_view/src/deprecated/BioStruct3DGLWidget.cpp
```diff
@@ -79,6 +79,19 @@ void BioStruct3DGLWidget::tryGL() {
     Q_UNUSED(wgt);
 }
 
+bool BioStruct3DGLWidget::canRender() {
+    QGLWidget w;
```
yalgaer, line 83: Have you tested current and deprecated variants of the method?

yalgaer (commented):
Are there any details why UGENE does not work in your environment correctly?

The code in this patch is good, but fix of the original issue is far more important than redirect of multiple 'trace' messages into a single 'error' message.

src/plugins/biostruct3d_view/src/deprecated/BioStruct3DGLWidget.cpp
evelinabis: > Are there any details why UGENE does not work in your environment correctly?

I don't know why it is not drawn. On Windows with OpenGL 1.1, [this example](https://doc.qt.io/qt-5/qtgui-openglwindow-example.html) works. However, 3D models are not rendered in UGENE (for both deprecated and non-deprecated code). On another computer with Windows and OpenGL 3.x, 3D models are rendered

> Have you tested current and deprecated variants of the method?

Yes

yalgaer: If you have a Windows installation that can run QT OpenGL examples with no issues, but UGENE's version of OpenGL visualization fails it is becomes very important to find the cause of the problem and provide a fix.

In this patch you try to use very basic features of OpenGL and it fails. Looks like the problem is with OpenGL libraries lookup, build mode flags, etc..

I believe that by comparing the working & non-working build and runtime environments you will be able to find the delta that can make UGENE work and create a real fix for the issue.

evelinabis: > >Have you tested current and deprecated variants of the method?
> 
> Yes

But locally in Windows. Don't know how non-deprecated version works on Linux

evelinabis: I hope the UTI-259 will be done and, perhaps, will solve this problem too. I see no point in dealing with deprecated code if the UTI-259 will possibly be done

yalgaer (approved):
I approved this patch. 
Please do not close/resolve the original JIRA issue because the complete solution is to fix the issue: find out why QT demo works and UGENE does not on the same environment.

PR #209, opened 2021-01-27
UGENE-7021. Edit mode cursor animation

yalgaer:


src/corelibs/U2View/src/ov_sequence/DetViewSequenceEditor.cpp
```diff
@@ -343,8 +342,10 @@ void DetViewSequenceEditor::sl_editMode(bool active) {
     }
 }
 
-void DetViewSequenceEditor::sl_changeCursorColor() {
-    cursorColor = (cursorColor == QColor(Qt::black)) ? Qt::darkGray : Qt::black;
+void DetViewSequenceEditor::sl_cursorAnimationTimerCallback() {
+    // Reproduce 'blink' effect: change the cursor color periodically.
+    // Show 'edit-cursor' only for the focused view and hide it (use transparent color) otherwise.
+    cursorColor = view->hasFocus() ? (cursorColor == Qt::black ? Qt::darkGray : Qt::black) : Qt::transparent;
```
DmitriiSukhomlinov: I think, there is no need to update view on each callback if color `cursorColor` is black and the view doesn't have focus. You can add something like this in the beginning:
```
if (!view->hasFocus() && cursorColor == Qt::black) {
    return;
}
```

src/corelibs/U2View/src/ov_sequence/DetViewSequenceEditor.h
```diff
@@ -73,7 +73,10 @@ class DetViewSequenceEditor : public QObject {
 
 private slots:
     void sl_editMode(bool active);
-    void sl_changeCursorColor();
+
+    /** Called periodically as a timer callback to animate blinking edit-mode cursor. */
```
DmitriiSukhomlinov, line 77: If we use [C++ documentation style](https://developer.lsst.io/cpp/api-docs.html) we will be able to generate code documentation one day (just in case, who knows, maybe it will be useful)

DmitriiSukhomlinov (dismissed):
Minor comments, approved

src/corelibs/U2View/src/ov_sequence/DetViewSequenceEditor.cpp
yalgaer: Agree, this is a valid optimization. Fixed.

src/corelibs/U2View/src/ov_sequence/DetViewSequenceEditor.h
yalgaer: I think that IDE (MSVC, QtCreator, CLion) are much better code browsers than HTML browser. 

In the documentation style I use I usually avoid mandatory tags and trivial statements and only put the additional information to the one that is already clear from the C++ method/field declaration.

PR #211, opened 2021-01-29
UGENE-6940. GUI tests for "Filter short scaffolds..." main menu action

evelinabis:


src/plugins/GUITestBase/GUITestBase.pro
```diff
@@ -259,6 +259,7 @@ HEADERS +=  src/GUITestBasePlugin.h \
             src/tests/common_scenarios/workflow_designer/estimating/GTTestsWorkflowEstimating.h \
             src/tests/common_scenarios/workflow_designer/name_filter/GTTestsWorkflowNameFilter.h \
             src/tests/common_scenarios/workflow_designer/parameters_validation/GTTestsWorkflowParameterValidation.h \
+            src/tests/common_scenarios/workflow_designer/samples/GTTestsWorkflowSamples.h \
```
DmitriiSukhomlinov, line 262: What kind of tests this file for? I'm not sure, that there is no way to add this test to an existing one, which is similar by purposes

DmitriiSukhomlinov (dismissed):

src/plugins/GUITestBase/GUITestBase.pro
evelinabis: I don't know. I can move the test wherever you say
Probably this folder exists for tests 6940--6952

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
```diff
@@ -3094,6 +3095,11 @@ void GUITestBasePlugin::registerTests(UGUITestBase *guiTestBase) {
     REGISTER_TEST(GUITest_common_scenarios_workflow_name_filter::test_0002);
     REGISTER_TEST(GUITest_common_scenarios_workflow_name_filter::test_0003);
 
+    /////////////////////////////////////////////////////////////////////////
+    // Common scenarios/Workflow designer/Samples
+    /////////////////////////////////////////////////////////////////////////
+    REGISTER_TEST_TL(GUITest_common_scenarios_workflow_samples::test_0001, minutes(7), labels({Cistrome, Linux, Windows, MacOS}));
```
yalgaer: The test itself looks simple, but you want to add to the special 'long-runing'  suite.
Unless we have a reason we should add tests to the Nightly (default) suite.
What is the execution time of the test?

yalgaer (changes requested):

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
evelinabis: 1 minute

yalgaer (approved):

PR #214, opened 2021-02-01
UGENE-6960. The project is automatically saved to a folder for which we have no permission to write

evelinabis:
Changed default save project path to UGENE_Data


yalgaer (changes requested):
All is good but we need a simple GUI test that will validate projectFilePathEdit content
The expected path can be taken directly from the current UserAppSettings 
Or you can set UserAppSettings value first and check that the new value is used in the profileFilePathEdit.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6605,6 +6607,52 @@ GUI_TEST_CLASS_DEFINITION(test_6959) {
     CHECK_SET_ERR(nameList3[5] == "Conocephalus_discolor", "The 6 sequence is incorrect");
 }
 
+GUI_TEST_CLASS_DEFINITION(test_6960) {
+    class ProjectPathValidationScenario : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus &os) override {
+            GTGlobals::sleep();
```
yalgaer: Please never use sleep() in the test unless this is a part of the logic of the test: "like wait for 30 seconds for a long-run task and cancel it"

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6605,6 +6607,52 @@ GUI_TEST_CLASS_DEFINITION(test_6959) {
...
+        }
+    };
+
+    // 1. Select "File->New project...". The "Create New Project" dialog appears
+    //    Expected state: In this dialog "Save project to file" field contains "~/Documents/UGENE_Data/project.uprj"
+    // 2. Close this dialog
+    GTUtilsDialog::waitForDialog(os, new DefaultDialogFiller(os, "CreateNewProjectDialog", QDialogButtonBox::Cancel, new ProjectPathValidationScenario()));
+    GTMenu::clickMainMenuItem(os, QStringList() << "File"
+                                                << "New project...");
+    GTGlobals::sleep();
```
yalgaer: Same here and below: remove sleeps

src/ugeneui/src/project_support/ProjectLoaderImpl.cpp
```diff
@@ -901,7 +902,7 @@ void ProjectDialogController::sl_projectNameEdited(const QString &text) {
 
 void ProjectDialogController::setupDefaults() {
     projectNameEdit->setText(ProjectLoaderImpl::tr("New Project"));
-    projectFilePathEdit->setText(QFileInfo("project" + PROJECTFILE_EXT).absoluteFilePath());
+    projectFilePathEdit->setText(QFileInfo(GUrlUtils::getDefaultDataPath(), "project" + PROJECTFILE_EXT).absoluteFilePath());
```
yalgaer: nit: extract the path value into a separate named string variable.

yalgaer (changes requested):

yalgaer (approved):

PR #233, opened 2021-02-11
UGENE-7047. Rework Tree & MSA sync mode to never show blocking dialogs

yalgaer:
In this PR I implemented Variant 2 discussed in the email thread:
* Sync mode is continuous in time
* Sync mode breaks when there is incompatible change in MSA
* There is no UI-blocking modal confirmation dialog about sync-mode anymore.
 
This patch also made possible removing of 2 GUI tests from the ignore list.


src/corelibs/U2Gui/U2Gui.qrc
```diff
@@ -162,6 +162,8 @@
         <file>images/sort_ascending.png</file>
         <file>images/ssp_logo.png</file>
         <file>images/sum.png</file>
+        <file>images/sync-msa-on.png</file>
```
DmitriiSukhomlinov, line 165: These icons have too small resolution. There was a letter from a user several days ago about small icons on Windows if the system "Scale" parameter is too large - this problem has appeared because of the fact, that almost all icons are 16x16 (some icons we have are 512x512 and this problem isn't appear to them). So, could you improve the resolution of the icon and make it 512x512?

DmitriiSukhomlinov (dismissed):
Minor comment, you either do it or not, approved

src/corelibs/U2Gui/U2Gui.qrc
yalgaer: I agree about the icon size: we should re-design all our icons. The icons in this PR are the same as all other icons. 

I would fix all icons separately and made them SVG (vector) if possible

PR #237, opened 2021-02-12
UGENE-7043. Replace the broken GLWidget with an warning text widget

evelinabis:
https://ugene.dev/tracker/browse/UGENE-7043
Windows systems that support DirectX and do not support OpenGL display a black screen instead of a 3D structure when opening a .pdb file. Instead of a QGLWidget with a black screen, display a QLabel with the text "Failed to initialize OpenGL"


src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.cpp
```diff
@@ -245,6 +247,9 @@ void BioStruct3DGLWidget::initializeGL() {
 
 void BioStruct3DGLWidget::resizeGL(int width, int height) {
     glFrame->updateViewPort(width, height);
+    if (lblGlError != nullptr) {
```
yalgaer, line 250: nit: for the future PRs: we use naming like glFrame (structure type is in the suffix). So lblGlError should be glErrorLabel.

src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.h
```diff
@@ -269,6 +271,10 @@ class BioStruct3DGLWidget : public QOpenGLWidget {
     static int getWidgetCount(QString objectName);
     bool isSyncModeOn();
 
+    // Checks if this widget can render and creates a label with a text error if it cannot
```
yalgaer, line 274: nit: put . at the end of sequences.

yalgaer (approved):
All is good. The "nit" comments can be taken into the account in next PRs.

DmitriiSukhomlinov (approved):

PR #238, opened 2021-02-12
UGENE-7052. Copied MSA region should be pasted with no sequence name

yalgaer:


src/corelibs/U2Core/src/globals/ClipboardController.cpp
```diff
@@ -26,6 +26,8 @@
 
 namespace U2 {
 
+const QString U2Clipboard::ugeneMimeType = "text/ugene";
```
DmitriiSukhomlinov: `static const QString U2Clipboard::UGENE_MIME_TYPE = "text/ugene";`

DmitriiSukhomlinov (dismissed):

src/corelibs/U2Core/src/globals/ClipboardController.cpp
yalgaer: Fixed.

evelinabis (dismissed):

src/corelibs/U2Gui/src/PasteController.cpp
```diff
@@ -106,18 +106,19 @@ PasteTask *PasteFactoryImpl::createPasteTask(bool isAddToProject) {
...
         return new PasteUrlsTask(mimeData->urls(), isAddToProject);
     }
-    QString clipboardText;
     try {
-        clipboardText = clipboard->text();
+        QString clipboardText = mimeData->hasFormat(U2Clipboard::UGENE_MIME_TYPE)
+                                    ? QString::fromUtf8(mimeData->data(U2Clipboard::UGENE_MIME_TYPE))
+                                    : clipboard->text();
+        if (clipboardText.isEmpty()) {
+            coreLog.error("UGENE can not recognize current clipboard content as one of the supported formats.");
```
DmitriiSukhomlinov: tr

src/corelibs/U2Gui/src/PasteController.cpp
```diff
@@ -106,18 +106,19 @@ PasteTask *PasteFactoryImpl::createPasteTask(bool isAddToProject) {
...
+            return nullptr;
```
DmitriiSukhomlinov, line 115: I don't like the structure here. The natural way out from `try` is `throw`, not `return`. Also I don't see any objects which are able to throw the exception at all, so the reason of the `try` block here is dubious. I offer you whether throw an exception here and catch it in the corresponding block or get rid of "try-catch" at all.

src/corelibs/U2Gui/src/PasteController.cpp
```diff
@@ -106,18 +106,19 @@ PasteTask *PasteFactoryImpl::createPasteTask(bool isAddToProject) {
...
+        }
+        return new PasteTextTask(clipboardText, isAddToProject);
```
DmitriiSukhomlinov, line 117: OK, not at all, `new` are able to throw `bad_alloc`, but I think it's OK if we don't take it an account.

DmitriiSukhomlinov (changes requested):
Some comments are need to be resolved

src/corelibs/U2Gui/src/PasteController.cpp
yalgaer: Fixed

src/corelibs/U2Gui/src/PasteController.cpp
yalgaer: This behavior: ("Data in clipboard is too large") was not changed in this PR at all.
We can create another issue to investigate why it was added here originally as a part of separate bug.

DmitriiSukhomlinov: OK, [done](https://ugene.dev/tracker/browse/UGENE-7060).

DmitriiSukhomlinov (approved):

PR #256, opened 2021-02-24
UGENE-7078 Set test directory from UGENE_TESTS_PATH if possible for C…

DmitriiSukhomlinov:
…InterfaceSasTests

https://ugene.dev/tracker/browse/UGENE-7078

PR #259, opened 2021-02-25
UTI-479 Revert 1 thread for XML tests

DmitriiSukhomlinov:
I want to get this feature back. I'd been setting up nightly XML tests on ugene-cuda and faced the problem - the execution fails because of the timeout in the single thread (e.g. [here](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewLog.html?buildId=213325&tab=buildResultsDiv&buildTypeId=UgeneNightlyBuilds_XmlTests_NightlyBuildUbuntu200464bitLinux)), but works good on several threads (I checked for 2 threads, [here](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewLog.html?buildId=212955&tab=buildResultsDiv&buildTypeId=UgeneNightlyBuilds_XmlTests_NightlyBuildUbuntu200464bitLinux), the only failed test will be fixed after #256).

I think, there is no problem about having this feature - if you want run tests in the single thread, you can just set the "TEST_THREADS" environment variable to 1.

### DIFF
--- a/src/plugins/test_runner/src/TestRunnerPlugin.cpp
+++ b/src/plugins/test_runner/src/TestRunnerPlugin.cpp
@@ -38,7 +38,7 @@
 
 #define SETTINGS_ROOT QString("test_runner/")
 /** Default number of threads to run XML tests. */
-#define NUM_THREADS_VAR_VALUE "1"
+#define NUM_THREADS_VAR_VALUE "10"
 #define TIME_OUT_VAR_VALUE "0"
 
 namespace U2 {
@@ -66,7 +66,6 @@ void TestRunnerPlugin::sl_startTestRunner() {
     TestRunnerService *srv = new TestRunnerService();
     srv->setEnvironment();
 
-    /* Disabling to check if it fixes slow commit/nightly builds
     CMDLineRegistry *cmdReg = AppContext::getCMDLineRegistry();
     if (cmdReg->hasParameter(CMDLineCoreOptions::TEST_THREADS)) {
         QString val = cmdReg->getParameterValue(CMDLineCoreOptions::TEST_THREADS);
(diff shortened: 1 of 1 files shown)


evelinabis (dismissed):

src/plugins/test_runner/src/TestRunnerPlugin.cpp
```diff
@@ -38,7 +38,7 @@
 
 #define SETTINGS_ROOT QString("test_runner/")
 /** Default number of threads to run XML tests. */
-#define NUM_THREADS_VAR_VALUE "1"
+#define NUM_THREADS_VAR_VALUE "10"
```
yalgaer: This magic number is not good: some (virtual) hosts may have 1 CPU, others (real) hosts may have 16 real CPUs and 10 threads is not good for either of them.

Auto-detection won't help too, because you do not know a user intention.

A better and a simpler solution would be keeping "1" by default and allowing override with any user value using environment variable.

yalgaer (changes requested):

src/plugins/test_runner/src/TestRunnerPlugin.cpp
DmitriiSukhomlinov: Fixed

evelinabis (approved):

DmitriiSukhomlinov:
Tests have passed ([Windows](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewLog.html?buildId=213636&tab=buildResultsDiv&buildTypeId=bt117), [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewLog.html?buildId=213634&tab=buildResultsDiv&buildTypeId=bt50), [macOS](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewLog.html?buildId=213671&tab=buildResultsDiv&buildTypeId=UgeneNightlyBuilds_NightlyMacOsX1075macX64)).

yalgaer (approved):

PR #263, opened 2021-03-01
UGENE-5045 Configuring "strand" alias for "find-orfs" workflow doesn't work

rasputinkirill:
XML test already added to repository cmdline\common\find-orfs\test_0009.xml

### DIFF
--- a/data/cmdline/find-orfs.uwl
+++ b/data/cmdline/find-orfs.uwl
@@ -72,6 +72,10 @@ workflow find-orfs{
                 alias:name;
                 description:"Name of the result annotations marking found ORFs (using 'ORF' by default)";
             }
+            orf-search.strand {
+                alias:strand;
+                description:"Strand to search in";
+            }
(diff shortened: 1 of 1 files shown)


yalgaer (commented):
Why did you change alignment.ugenedb in the SVN?

data/cmdline/find-orfs.uwl
```diff
@@ -72,6 +72,10 @@ workflow find-orfs{
                 alias:name;
                 description:"Name of the result annotations marking found ORFs (using 'ORF' by default)";
             }
+            orf-search.strand {
+                alias:strand;
+                description:"Strand to search in";
```
yalgaer: Please add 
1) Name of the options, so user knows which options should be passed in the command line.
2) Default option info

yalgaer (changes requested):

yalgaer (approved):

PR #272, opened 2021-03-06
UGENE-7100. Remove all usages of sleep from GUITest_common_scenarios_document_from_text suite

yalgaer:


src/ugene_version.pri
```diff
@@ -14,7 +14,7 @@ U2_DISTRIBUTION_INFO=sources
 # int version levels for executables
 UGENE_VER_MAJOR=38
 UGENE_VER_MINOR=0
-UGENE_VER_SUFFIX=-dev
+UGENE_VER_SUFFIX=
```
DmitriiSukhomlinov: Why do you want to delete this? We have to hate it in "master"

DmitriiSukhomlinov (changes requested):
Why do you wanna delete "-dev"?

src/ugene_version.pri
yalgaer: I used the wrong target branch.

DmitriiSukhomlinov (approved):

PR #276, opened 2021-03-08
UGENE-7104. Fix for tree_settings_test_0004 test

yalgaer:


src/corelibs/U2View/src/ov_phyltree/TreeViewer.cpp
```diff
@@ -1256,7 +1257,7 @@ void TreeViewerUI::sl_onBranchCollapsed(GraphicsRectangularBranchItem *) {
     qreal curScale = getScale();
     redrawRectangularLayout();
     setScale(curScale);
-    updateTreeSettings(false);
+    updateScene(false);
```
evelinabis, line 1260: Everywhere "true" is replaced by "false", and "false" is replaced by "true", except this place

evelinabis (approved):

src/corelibs/U2View/src/ov_phyltree/TreeViewer.cpp
yalgaer: This is not true. Nothing actually was changed from the method semantic perspective:

1) I removed default value from the virtual function. The default value was 'true'
2) Everywhere in TreeViewer (standalone view) I kept the original value
3) Everywhere in MsaEditorTreeViewer it does not matter what value is passed because MsaEditorTreeViewer enforces 'false' value. So the the change from the 'true' to 'false' in MsaEditorTreeViewer is also no-op, but 'false' looks more correct.

If you find in the code the inconsistency with what I wrote above, please comment the place in the code.

evelinabis (approved):

PR #277, opened 2021-03-09
UGENE-7106. Dragging in MSA with trees does not work

yalgaer:


evelinabis (dismissed):

src/corelibs/U2View/src/ov_msa/MSAEditorSequenceArea.cpp
```diff
@@ -850,10 +852,19 @@ void MSAEditorSequenceArea::sl_enableFreeRowOrderMode(const QList<QStringList> &
...
+    freeModeMasterMarkersSet.insert(marker);
     updateRowOrderActionsState();
     ui->getCollapseModel()->update(collapsibleGroupList);
 }
 
+void MSAEditorSequenceArea::disableFreeRowOrderMode(QObject *marker) {
+    freeModeMasterMarkersSet.remove(marker);
+    if (freeModeMasterMarkersSet.isEmpty() && editor->getRowOrderMode() == MaEditorRowOrderMode::Free) {
+        // Switch back to the Original ordering.
+        sl_toggleSequenceRowOrder(false);
```
DmitriiSukhomlinov, line 864: Original ordering is the same as in the file. But what if sequences were reordered during the `MaEditorRowOrderMode::Free` mode? The file will be updated? Or how it works like?

src/corelibs/U2View/src/ov_msa/MSAEditorSequenceArea.h
```diff
@@ -233,6 +242,17 @@ private slots:
...
+    /**
+     * Set of 'marker' objects from the 'master' components that requested Free ordering mode to be ON are responsible for the 'free' mode ordering.
+     * Free mode can be active only if there is at least one 'marker' in the set.
+     *
+     * When the last marker object is removed from the set the ordering automatically switches to the 'Original'.
+     * Example of master components: multiple synchronized phy-tree views that manage the order of MSA.
+     *
+     * MSAEditor can any time reset this set and switch to 'Original' or 'Sequence' mode.
+     */
+    QSet<QObject *> freeModeMasterMarkersSet;
```
DmitriiSukhomlinov, line 255: I think, we can specify this pointer more tough, at least as a pointer to `GObjectView`. `QObject` too much vague.

src/corelibs/U2View/src/ov_msa/PhyTrees/MSAEditorTreeViewer.cpp
```diff
@@ -50,6 +50,12 @@ MSAEditorTreeViewer::MSAEditorTreeViewer(const QString &viewName, GObject *obj,
       msaTreeViewerUi(nullptr) {
 }
 
+MSAEditorTreeViewer::~MSAEditorTreeViewer() {
+    if (editor != nullptr && isSyncModeEnabled()) {
```
DmitriiSukhomlinov, line 54: Don't like it. Destructors are supposed to clear the captured memory, not to do some post-actions. I think, it's better to make the slot `objectIsAbout2BeDestroyed`, connect it to [QObject::destroyed](https://doc.qt.io/qt-5/qobject.html#destroyed) and move this part of code to this slot.

src/corelibs/U2View/src/ov_msa/MSAEditorSequenceArea.cpp
yalgaer: 'Free' is a virtual ordering managed by Tree viewer.
When MSA Editor is in Free mode the file is not updated (it is virtual like Order-By-Sequence).

src/corelibs/U2View/src/ov_msa/PhyTrees/MSAEditorTreeViewer.cpp
DmitriiSukhomlinov, line 54: `editor != nullptr` why do you need this check? I can't understand, why the `editor` in destructor could be removed or not removed yet?

DmitriiSukhomlinov (approved):

src/corelibs/U2View/src/ov_msa/MSAEditorSequenceArea.h
yalgaer: I have no strong opinion today about the type for the 'marker'. For the purpose we have today it can even be void*.

More than that in the next release it may completely be gone: all depends which of the scenarios described in the bug we will select.

DmitriiSukhomlinov:
Also - are you sure, that this commit should go to the "release-38.0" branch?  As far as I understand, we have feature-freeze now, so this changes shouldn't be released in 38.0.

DmitriiSukhomlinov (changes requested):
Approved by mistake, a lot of major comments.

src/corelibs/U2View/src/ov_msa/PhyTrees/MSAEditorTreeViewer.cpp
yalgaer: Sorry didn't get the question. Anyway this code is removed.

yalgaer: Ok, will try to replace it with destroyed() signal.

yalgaer:
> Also - are you sure, that this commit should go to the "release-38.0" branch? As far as I understand, we have feature-freeze now, so this changes shouldn't be released in 38.0.

Absolutely. This one is not a feature.

src/corelibs/U2View/src/ov_msa/PhyTrees/MSAEditorTreeViewer.cpp
yalgaer: I prepared a solution with destroy() signal handling, checked that it worked, but later decided to roll it back (see the history of commits) 

The way to solve the problem with destroy() adds a new state management logic (subscribe/unsubscribe/do-not subscribe twice...) and adds a lot of boilerplate code.

For example the original solution with the destructor requires 3 lines against 10+ in the new code and technically works absolutely the same. But the original is much simpler and is less error prone. 

I think the final cleanup is out of scope today because the code will be reworked in the next release: moved away from MsaSequenceArea to MsaEditor, etc. So we will have a second chance to discuss it.

DmitriiSukhomlinov: I disagree and think, that the solution with `destroy()` was good enough. I won't argue about your solution cuz don't wan't to.

DmitriiSukhomlinov: Why do you need this check? I mean, `editor != nullptr`. In what case `editor` equals `nullptr`?

yalgaer: I agree with your point that 'destructors are evil'. 

But the problem is that the alternative with connect/disconnect is not better at all: it is still a manual management of life-range callbacks using very verbose QT signal/slots. 

I accept the problem you mentioned and will try to address it in a better way during the first spring of the next release. For now I'm keeping the smaller/simpler solution.

yalgaer: There are 2 cases when editor may be null:

1) When tree view is created but editor is not yet assigned. We have a TODO for it in the code to avoid this state.

2) When MsaEditor is destroyed. The tree-view is be destroyed after the editor instance. 
This order looks wrong to me: the tree view must be destroyed after the MsaEditor, but the release sprint is not the best time to make such big refactoring.

PR #278, opened 2021-03-10
UTI-484

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UTI-484

The problem has appeared in Windows API functions`shlobj.h` (1) and `shlguid.h` (2) . They both have inners inside of the  `WINAPI_PARTITION_DESKTOP` partition. The problem is that (2) uses the macro defined in the (1). So, they should be included in the following order, otherwise - compile errors:
```
#include <shlobj.h>
#include <shlguid.h>

```


yalgaer (approved):
Could you please also surround the fixed code with
// clang-format on
// clang-format off

Otherwise clang-format will change it back.

PR #292, opened 2021-03-18
UGENE-7131. Make unused-parameter compiler warning be a compile-time …

yalgaer:
1. This patch does not introduce any changes to the code and is safe to be merged both into release-38 or master branches.

2. The UI file has warnings too and was opened and saved back with no changes: QtDesigner automatically removed the bad code.

3. Some files include "disable-warnings" header and some have a warning fix in place. The ones that have the fix contain our own code. The ones that have no fix but include "disable-warnings" header are 3rd party code we do not want to modify.


evelinabis (approved):

DmitriiSukhomlinov (approved):
lgtm

PR #295, opened 2021-03-21
UGENE-7134. Unicode support for text document formats

yalgaer:
A feature for the version 39.

### DIFF
--- a/src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
+++ b/src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
@@ -123,7 +123,8 @@ int IOAdapterReader::read(U2OpStatus &os, QString &result, int maxLength, const 
     textForUndo.clear();
     bool isReadingTerminatorSequence = false;
     while (!stream.atEnd() && result.length() != maxLength) {
-        QChar unicodeChar = get();
+        QChar unicodeChar = get(os);
+        CHECK_OP(os, 0);
         uchar latin1Char = unicodeChar.toLatin1();
         bool isTerminatorChar = terminators.at(latin1Char);
         if (isTerminatorChar) {
@@ -151,11 +152,12 @@ int IOAdapterReader::read(U2OpStatus &os, QString &result, int maxLength, const 
     return result.length();
 }
 
-QChar IOAdapterReader::get() {
+QChar IOAdapterReader::get(U2OpStatus &os) {
     QChar ch;
     if (unreadCharsBuffer.isEmpty()) {
         stream >> ch;
     } else {
+        SAFE_POINT_EXT(unreadCharsBufferPos < unreadCharsBuffer.size(), os.setError(L10N::internalError()), 0);
         ch = unreadCharsBuffer[unreadCharsBufferPos];
         unreadCharsBufferPos++;
         if (unreadCharsBufferPos == unreadCharsBuffer.length()) {
--- a/src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
+++ b/src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
@@ -19,7 +19,7 @@
  * MA 02110-1301, USA.
  */
 
-#include "TextStream.h"
+#include "IOAdapterTextStream.h"
 
 #include <QTextCodec>
 
@@ -79,9 +79,9 @@ bool IOAdapterDevice::seek(qint64 seekPos) {
 }
 
 ///////////////////////////////////////////////
-////      TextStreamBase
+////      IOAdapterReaderAndWriterBase
 ///////////////////////////////////////////////
-TextStreamBase::TextStreamBase(IOAdapter *_ioAdapter, QTextCodec *codec)
+IOAdapterReaderAndWriterBase::IOAdapterReaderAndWriterBase(IOAdapter *_ioAdapter, QTextCodec *codec)
     : ioAdapter(_ioAdapter) {
     ioDevice.reset(new IOAdapterDevice(ioAdapter));
     stream.setDevice(ioDevice.data());
@@ -92,32 +92,32 @@ TextStreamBase::TextStreamBase(IOAdapter *_ioAdapter, QTextCodec *codec)
     }
 }
 
-GUrl TextStreamBase::getURL() const {
+GUrl IOAdapterReaderAndWriterBase::getURL() const {
     return ioAdapter->getURL();
 }
 
-IOAdapterFactory *TextStreamBase::getFactory() const {
+IOAdapterFactory *IOAdapterReaderAndWriterBase::getFactory() const {
     return ioAdapter->getFactory();
 }
 
 ///////////////////////////////////////////////
-////      TextStreamReader
+////      IOAdapterReader
 ///////////////////////////////////////////////
-TextStreamReader::TextStreamReader(IOAdapter *ioAdapter)
-    : TextStreamBase(ioAdapter), unreadCharsBufferPos(0) {
+IOAdapterReader::IOAdapterReader(IOAdapter *ioAdapter)
+    : IOAdapterReaderAndWriterBase(ioAdapter), unreadCharsBufferPos(0) {
 }
 
-TextStreamReader::~TextStreamReader() {
+IOAdapterReader::~IOAdapterReader() {
     if (unreadCharsBufferPos < unreadCharsBuffer.length() && ioAdapter->isOpen()) {
         // Roll IOAdapter back to the size of the data left in the unreadCharsBuffer.
-        // This is needed to correctly support streaming (reading multiple documents from the same file using TextStreamReader)
+        // This is needed to correctly support streaming (reading multiple documents from the same file using IOAdapterReader)
         int nCharsToReturn = unreadCharsBuffer.length() - unreadCharsBufferPos;
         int nBytesToReturn = stream.codec()->fromUnicode(unreadCharsBuffer.right(nCharsToReturn)).length();
         stream.seek(stream.pos() - nBytesToReturn);
     }
 }
 
-int TextStreamReader::read(U2OpStatus &os, QString &result, int maxLength, const QBitArray &terminators, IOAdapter::TerminatorHandling terminatorMode, bool *terminatorFound) {
+int IOAdapterReader::read(U2OpStatus &os, QString &result, int maxLength, const QBitArray &terminators, IOAdapter::TerminatorHandling terminatorMode, bool *terminatorFound) {
     CHECK_OP(os, 0);
     result.clear();
     textForUndo.clear();
@@ -151,7 +151,7 @@ int TextStreamReader::read(U2OpStatus &os, QString &result, int maxLength, const
     return result.length();
 }
 
-QChar TextStreamReader::get() {
+QChar IOAdapterReader::get() {
     QChar ch;
     if (unreadCharsBuffer.isEmpty()) {
         stream >> ch;
@@ -167,13 +167,13 @@ QChar TextStreamReader::get() {
     return ch;
 }
 
-void TextStreamReader::unget() {
+void IOAdapterReader::unget() {
     SAFE_POINT(!textForUndo.isEmpty(), "Can't unget(), textForUndo is empty!", );
     SAFE_POINT(unreadCharsBuffer.isEmpty(), "unreadCharsBuffer must be empty during unget()", );
     unreadCharsBuffer.append(textForUndo[textForUndo.length() - 1]);
 }
 
-void TextStreamReader::undo() {
+void IOAdapterReader::undo() {
     SAFE_POINT(!textForUndo.isEmpty(), "Only 1 undo() per read() is supported!", );
     SAFE_POINT(unreadCharsBuffer.length() <= 1, "unreadCharsBuffer size must be <= 1!", );
     unreadCharsBuffer.clear();
@@ -182,18 +182,18 @@ void TextStreamReader::undo() {
     textForUndo.clear();
 }
 
-int TextStreamReader::getProgress() const {
+int IOAdapterReader::getProgress() const {
     return ioAdapter->getProgress();
 }
 
 ///////////////////////////////////////////////
-////      TextStreamWriter
+////      IOAdapterWriter
 ///////////////////////////////////////////////
-TextStreamWriter::TextStreamWriter(IOAdapter *ioAdapter, QTextCodec *codec)
-    : TextStreamBase(ioAdapter, codec) {
+IOAdapterWriter::IOAdapterWriter(IOAdapter *ioAdapter, QTextCodec *codec)
+    : IOAdapterReaderAndWriterBase(ioAdapter, codec) {
 }
 
-void TextStreamWriter::writeBlock(U2OpStatus &os, const QString &text) {
+void IOAdapterWriter::write(U2OpStatus &os, const QString &text) {
     CHECK_OP(os, );
     stream << text;
     if (ioAdapter->hasError()) {
--- a/src/corelibs/U2Core/src/io/IOAdapterTextStream.h
+++ b/src/corelibs/U2Core/src/io/IOAdapterTextStream.h
@@ -64,11 +64,11 @@ private:
     qint64 currentPos;
 };
 
-/** Base class for reader & writer streams. */
-class U2CORE_EXPORT TextStreamBase {
+/** Base class for reader & writer io-adapter text streams. */
+class U2CORE_EXPORT IOAdapterReaderAndWriterBase {
 public:
     /** Initializes underlying qt stream & io-device. If 'codec' is null uses default locale codec. */
-    TextStreamBase(IOAdapter *ioAdapter, QTextCodec *codec = nullptr);
+    IOAdapterReaderAndWriterBase(IOAdapter *ioAdapter, QTextCodec *codec = nullptr);
 
     /** Returns url from the ioAdapter. */
     GUrl getURL() const;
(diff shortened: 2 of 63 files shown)


src/corelibs/U2Core/src/io/TextStream.h
```diff
@@ -0,0 +1,155 @@
...
+    /** Returns url from the ioAdapter. */
+    GUrl getURL() const;
+
+    /** Returns factory from the ioAdapter. */
+    IOAdapterFactory *getFactory() const;
+
+protected:
+    IOAdapter *ioAdapter;
+    QScopedPointer<QIODevice> ioDevice;
+    QTextStream stream;
```
evelinabis: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c133-avoid-protected-data

src/corelibs/U2Core/src/io/TextStream.cpp
```diff
@@ -0,0 +1,204 @@
...
+////      IOAdapterDevice
+///////////////////////////////////////////////
+IOAdapterDevice::IOAdapterDevice(IOAdapter *_ioAdapter, QObject *parent)
+    : QIODevice(parent), ioAdapter(_ioAdapter), currentPos(0) {
+    SAFE_POINT(ioAdapter->isOpen(), "IOAdapter must be opened", );
+    bool isReadable = ioAdapter->isIOModeSupported(IOAdapterMode_Read);
+    bool isWritable = ioAdapter->isIOModeSupported(IOAdapterMode_Write);
+    SAFE_POINT(isReadable || isWritable, "IOAdapter must support read, write or both ops", );
+
+    OpenMode deviceMode(QIODevice::Unbuffered);
```
evelinabis: Why don't you use buffering?

src/corelibs/U2Core/src/io/TextStream.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /** Pushes back all unused buffers to ioAdapter by calling 'skip' method. */
+    ~TextStreamReader();
+
+    /**
+     * Reads the stream until one of the separators is found. Saves the result to 'result' string buffer.
+     * Clears the 'result' buffer first before reading.
+     * Returns number of characters read, same as result.length().
+     */
+    int read(U2OpStatus &os, QString &result, int maxLength, const QBitArray &terminators, IOAdapter::TerminatorHandling terminatorMode, bool *terminatorFound = 0);
```
evelinabis: Replace `bool *terminatorFound = 0` with `bool *terminatorFound = nullptr`

src/corelibs/U2Formats/src/ClustalWAlnFormat.h
```diff
@@ -34,21 +34,20 @@ class IOAdapter;
...
 
-    virtual void storeEntry(IOAdapter *io, const QMap<GObjectType, QList<GObject *>> &objectsMap, U2OpStatus &os);
+    void storeTextEntry(TextStreamWriter &writer, const QMap<GObjectType, QList<GObject *>> &objectsMap, U2OpStatus &os) override;
 
-protected:
-    virtual FormatCheckResult checkRawTextData(const QByteArray &rawData, const GUrl & = GUrl()) const;
+    FormatCheckResult checkRawTextData(const QString &dataPrefix, const GUrl & = GUrl()) const override;
 
-    virtual Document *loadTextDocument(IOAdapter *io, const U2DbiRef &dbiRef, const QVariantMap &fs, U2OpStatus &os);
+    Document *loadTextDocument(TextStreamReader &reader, const U2DbiRef &dbiRef, const QVariantMap &fs, U2OpStatus &os) override;
```
evelinabis, line 48: Where is `DNASequence *loadTextSequence(TextStreamReader &reader, U2OpStatus &os) override`?

src/corelibs/U2Formats/src/TextDocumentFormat.h
```diff
@@ -41,6 +49,61 @@ class U2FORMATS_EXPORT TextDocumentFormat : public DocumentFormat {
...
+     * If loading is failed the nullptr is returned and the error is stored in 'os', otherwise a non-nullptr value is returned.
+     */
+    virtual Document *loadTextDocument(TextStreamReader &reader, const U2DbiRef &dbiRef, const QVariantMap &hints, U2OpStatus &os) = 0;
+
+    /**
+     * Loads a DNASequence object from the current state of the reader.
+     * Returns a valid sequence object or sets an error to the 'os'.
+     * By default returns nullptr & error that sequence loading is not supported by the current format.
+     */
+    virtual DNASequence *loadTextSequence(TextStreamReader &reader, U2OpStatus &os);
```
evelinabis: Why isn't this method pure virtual like `loadTextDocument`?

src/corelibs/U2Formats/src/TextDocumentFormat.h
```diff
@@ -41,6 +49,61 @@ class U2FORMATS_EXPORT TextDocumentFormat : public DocumentFormat {
...
+     * By default text documents do not support writing, so this method must be re-implemented in every format to support serialization.
+     */
+    virtual void storeTextDocument(TextStreamWriter &writer, Document *document, U2OpStatus &os);
+
+    /**
+     * Stores all supported document types from the given map of objects.
+     * The document format should preserve the objects order in the map when possible.
+     * The default implementation does nothing and sets 'unsupported' error to 'os'.
+     */
+    virtual void storeTextEntry(TextStreamWriter &writer, const QMap<GObjectType, QList<GObject *>> &objectsMap, U2OpStatus &os);
```
evelinabis: Why aren't these methods pure virtual?

evelinabis (changes requested):

src/corelibs/U2Core/src/io/TextStream.cpp
yalgaer: IO adapter internally uses buffering, we do not need to repeat it twice.

src/corelibs/U2Formats/src/ClustalWAlnFormat.h
yalgaer: ClustalW does not support DNA sequence. The default impl that sets error flag is enough

src/corelibs/U2Formats/src/TextDocumentFormat.h
yalgaer: They have default impl that that says that writing is not supported. This is a valid case for all document formats

src/corelibs/U2Formats/src/TextDocumentFormat.h
yalgaer: It has a default impl that is reusable for all text formats that do not have DNASequence objects inside.
The loadTextDocument is the 'must' method to override for all text document formats, so it is pure virtual.

src/corelibs/U2Core/src/io/TextStream.h
yalgaer: Fixed

src/corelibs/U2Core/src/io/TextStream.h
yalgaer: The overall suggestion to reduce usage of protected state is reasonable.

Could you please elaborate which improvements do you suggest in this particular case? 
For example what should I do with a "QTextStream stream" field?

evelinabis: Everything suits me https://isocpp.org/wiki/faq/basics-of-inheritance#protected-data-not-evil

src/corelibs/U2Core/src/io/TextStream.cpp
evelinabis: Perhaps this buffer is more low-level

evelinabis (dismissed):

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+///////////////////////////////////////////////
+////      IOAdapterReader
+///////////////////////////////////////////////
+IOAdapterReader::IOAdapterReader(IOAdapter *ioAdapter)
+    : IOAdapterReaderAndWriterBase(ioAdapter), unreadCharsBufferPos(0) {
+}
+
+IOAdapterReader::~IOAdapterReader() {
+    if (unreadCharsBufferPos < unreadCharsBuffer.length() && ioAdapter->isOpen()) {
```
DmitriiSukhomlinov, line 111: And again - destructior is supposed to clean memory, it's not about doing some post-actions. It's better to connect to [destroyed](https://doc.qt.io/qt-5/qobject.html#destroyed)

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
```diff
@@ -0,0 +1,204 @@
+/**
```
DmitriiSukhomlinov, line 1: I think, we don't have a rule us to be punched for each new file each of us has add to the project. Cuz I have no Idea why several classes are concentrated in a single ".cpp/.h" pair of files so often. I offer to split these classes - one class one ".cpp/.h" pair of files.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+    return result.length();
+}
+
+QChar IOAdapterReader::get() {
+    QChar ch;
+    if (unreadCharsBuffer.isEmpty()) {
+        stream >> ch;
+    } else {
+        ch = unreadCharsBuffer[unreadCharsBufferPos];
```
DmitriiSukhomlinov, line 161: If `unreadCharsBuffer `doesn't have a char with index `unreadCharsBufferPos`? I offer to add a SAFE_POINT

src/corelibs/U2Core/src/io/IOAdapterTextStream.h
```diff
@@ -0,0 +1,155 @@
...
+     * Proxies the call to ioAdapter.getProgress method.
+     */
+    int getProgress() const;
+
+    /** Undo last read() operation. */
+    void undo();
+
+private:
+    /** Reads a single character from the stream. Can be called only in the context of 'read' operation. */
+    QChar get();
```
DmitriiSukhomlinov: If function has name, starts with `get`, than this function should be `const`.

src/corelibs/U2Formats/src/TextDocumentFormat.cpp
```diff
@@ -22,23 +22,24 @@
...
 
 #include <U2Core/IOAdapter.h>
+#include <U2Core/IOAdapterTextStream.h>
 #include <U2Core/TextUtils.h>
 #include <U2Core/U2OpStatus.h>
 
 namespace U2 {
 
-TextDocumentFormat::TextDocumentFormat(QObject *p, const DocumentFormatId &id, DocumentFormatFlags _flags, const QStringList &fileExts)
+TextDocumentFormatDeprecated::TextDocumentFormatDeprecated(QObject *p, const DocumentFormatId &id, DocumentFormatFlags _flags, const QStringList &fileExts)
```
DmitriiSukhomlinov, line 31: I don't get why this class is called deprecated now. Are we about to replace it? As far as I understand, we just switched the adapter

src/corelibs/U2Formats/src/TextDocumentFormat.cpp
```diff
@@ -48,17 +49,70 @@ FormatCheckResult TextDocumentFormat::checkRawData(const QByteArray &rawData, co
...
+/// TextDocumentFormat
+///////////////////////////////////////////
+
+TextDocumentFormat::TextDocumentFormat(QObject *p, const DocumentFormatId &id, DocumentFormatFlags _flags, const QStringList &fileExts)
+    : DocumentFormat(p, id, _flags, fileExts) {
+}
+
+FormatCheckResult TextDocumentFormat::checkRawData(const QByteArray &rawBinaryData, const GUrl &url) const {
+    QTextStream stream(rawBinaryData, QIODevice::ReadOnly);    // Use QTextStream to auto-detect multi-byte encoding.
+    QString text = stream.readAll();
```
DmitriiSukhomlinov, line 75: Byte Order Marks are still a problem, I suppose

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: "And again" - it won't work here. There are no qobjects here and even if they were this very explicit way of lifecycle handling is what needed and it is documented.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: Added

src/corelibs/U2Formats/src/TextDocumentFormat.cpp
yalgaer: I suppose not.

src/corelibs/U2Core/src/io/IOAdapterTextStream.h
yalgaer: It is not. Same with read()

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: I do not support this level of separation: 1 file per 1 class.

Multiple very related classes can live within 1 file if the overall file size is small.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: I have a second thought here: there is no "safe" way to return any character value from the SAFE_POINT here. If something is wrong here "crash" or "exception" is the best solution.
Returning of a non-meaningful character is wrong because it may lead to unpredictable behavior. And crash is better than any unpredictable behavior.

DmitriiSukhomlinov: Crash or unpredictable behavior is the worst thing we can have in this very often used class. This problem HAS to be resolved, and as far as this class is a part of the solution of this issue, this problem HAS to be resolved right here and right now. 
As far as any symbol could be met in file, we can't use one of them as an error symbol (I mean, symbol which is returned in case of the error). The best thing, I think, is to throw the exception and catch it one level below on the call stack.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
DmitriiSukhomlinov: `IOAdapterReader` is `QObject`, actually, and it's enough to realize my offering. I don't get why you disagree, actually. I mean, I don't get the explanation about lifestyle.

src/corelibs/U2Core/src/io/IOAdapterTextStream.h
DmitriiSukhomlinov: So rename it to `readChar()`

src/corelibs/U2Formats/src/TextDocumentFormat.cpp
DmitriiSukhomlinov: You suppose or you have checked it?

DmitriiSukhomlinov (changes requested):
Still comments. And please, don't resolve conversations, I do it by myself when I understand that the problem I have described is solved.

src/corelibs/U2Core/src/io/IOAdapterTextStream.h
yalgaer: My initial idea was that everyone who is familiar with streams will find this name appropriate;
http://www.cplusplus.com/reference/istream/istream/get/

src/corelibs/U2Formats/src/TextDocumentFormat.cpp
yalgaer: Hope this will help: https://doc.qt.io/qt-5/qtextstream.html.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: You are wrong: IOAdapterReader is not a QObject and your suggestion to always use slot/signals for object destruction is not supported. I use it only when I feel that the pattern reads/works better than the basic plain language.

In this case I want to go low-level as much as possible, stay on the low-level for Reader & Writer. Keeping the destructor to manage internal resources of the stream looks like a reasonable option.

In general I agree - yes: destructors should be avoided. But not in 100% of cases and use of QT slots/signals is not always the best solutions for that.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: Why do you think there is a problem to solve here? Both variables are managed internally only and they are never out of sync.

We have tests that prove that the variables state is always correct and will have even more after all formats are migrated. 
Do you have any scenario where there is a chance of error here? If yes, lets add the test?

Because otherwise you ask me to add a boilerplate code for the impossible situation. This code will spoil the readability and will never work: the situation is impossible. 

The extra checks will be valid if the state could be affected by user, but here it is purely internal state.
Here is an example of the similar QT code: https://code.woboq.org/qt5/qtbase/src/corelib/io/qbuffer.cpp.html

yalgaer:
> Still comments. And please, don't resolve conversations, I do it by myself when I understand that the problem I have described is solved.

Ok. 
Please start finding the real issues, but not only focus on code-style only we have never set any rules in the project before.

If what you need is a learning course about QTextStream or why something works like this or that in UGENE I think it does not matter if conversation resolved or not: it is just a side discussion not related to the review directly.

I confirm that I will not close these conversations if you want.

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: Finally I think I found an option that will suite both of us:
 I added SAFE_POINT with os.setError() inside. Reporting errors via OpStatus is the way this class works for all other cases.

DmitriiSukhomlinov: Yes, I like this solution.

src/corelibs/U2Core/src/io/IOAdapterTextStream.h
DmitriiSukhomlinov: OK, but I didn't get it)

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
DmitriiSukhomlinov: Actually, `IOAdapterReader` is the inheritor of `QObject`. But I don't want to argue anymore, so OK

DmitriiSukhomlinov (approved):

src/corelibs/U2Core/src/io/IOAdapterTextStream.cpp
yalgaer: I think you mean IOAdapter, but here it is IOAdapterReader that has no QObject in the hierarchy

PR #300, opened 2021-03-23
UGENE-7138. Enable warning about unused variables in GCC build

yalgaer:
The logic behind the warnings enabling/suppression
1) Is it UGENE code? Enable the warning, fix issues (if any). Reformat the file if needed.
2) Is it a 3rd party code? Disable all warnings. Do not reformat to keep diffs with the original tool simple.

### DIFF
--- a/src/plugins_3rdparty/primer3/src/primer3_core/oligotm.c
+++ b/src/plugins_3rdparty/primer3/src/primer3_core/oligotm.c
@@ -86,7 +86,7 @@ All rights reserved.
  */
 
 /* Table 1 (old parameters):
- * See table 2 in the paper [Breslauer KJ, Frank R, Bl�cker H and
+ * See table 2 in the paper [Breslauer KJ, Frank R, Blöcker H and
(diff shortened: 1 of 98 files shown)


src/plugins_3rdparty/primer3/src/primer3_core/oligotm.c
```diff
@@ -83,7 +86,7 @@ All rights reserved.
  */
 
 /* Table 1 (old parameters):
- * See table 2 in the paper [Breslauer KJ, Frank R, Bl�cker H and
+ * See table 2 in the paper [Breslauer KJ, Frank R, Bl�cker H and
```
evelinabis: It was better

evelinabis (dismissed):

src/plugins_3rdparty/primer3/src/primer3_core/oligotm.c
yalgaer: Looks like the original sources were written with some non-standard encoding and now after the change my editor fails to convert them correctly to UTF-8. 
I fixed the bad char manually.

PR #303, opened 2021-03-23
UGENE-7116. Unignore GUITest_common_scenarios_shared_db_wd_run_workflow_gui_test_0003 test

evelinabis:


src/ugene_version.pri
```diff
@@ -12,7 +12,7 @@ U2_DISTRIBUTION_INFO=sources
 }
 
 # int version levels for executables
-UGENE_VER_MAJOR=38
+UGENE_VER_MAJOR=39
```
yalgaer: This file should not be changed - looks like it is some merge issue.

The patch is is good, I will approve it once you remove this file from the patch.

yalgaer (changes requested):

DmitriiSukhomlinov (approved):

yalgaer (approved):

PR #306, opened 2021-03-24
UGENE-7141. Remove trivial GCC warnings suppression

yalgaer:
The final goal is to remove suppression for all warnings, so UGENE will be warning free and our GCC-warning-check build  in the Teamcity can alert when new warnings are found.

The warnings that left after this PR require UGENE code modification and will be done in dedicated PRs.

### DIFF
--- a/src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignment.cpp
+++ b/src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignment.cpp
@@ -604,7 +604,7 @@ MultipleChromatogramAlignment MultipleChromatogramAlignmentData::getExplicitCopy
 void MultipleChromatogramAlignmentData::copy(const MultipleAlignmentData &other) {
     try {
         copy(dynamic_cast<const MultipleChromatogramAlignmentData &>(other));
-    } catch (std::bad_cast) {
+    } catch (std::bad_cast &) {
         FAIL("Can't cast MultipleAlignmentData to MultipleChromatogramAlignmentData", );
     }
 }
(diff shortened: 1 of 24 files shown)


evelinabis (approved):

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignment.cpp
```diff
@@ -604,7 +604,7 @@ MultipleChromatogramAlignment MultipleChromatogramAlignmentData::getExplicitCopy
 void MultipleChromatogramAlignmentData::copy(const MultipleAlignmentData &other) {
     try {
         copy(dynamic_cast<const MultipleChromatogramAlignmentData &>(other));
-    } catch (std::bad_cast) {
+    } catch (std::bad_cast &) {
```
DmitriiSukhomlinov, line 607: I suppose, it's better to catch `const std::bad_cast &`

yalgaer: I also do not like the code here, but behavior changes are out of scope in this PR

DmitriiSukhomlinov (approved):
Minor comments

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignment.cpp
DmitriiSukhomlinov: In this case, why did you change `std::bad_cast` to `std::bad_cast &`?

yalgaer: The warning (-Wcatch-value) was not related to the empty catch clause but to the way the value is caught. See other places with non-empty catch clause that were also changed .

PR #307, opened 2021-03-24
UGENE-7140. Translations initialization cleanup

yalgaer:
Check the JIRA issue for more details.


evelinabis (dismissed):

src/ugenecl/src/Main.cpp
```diff
@@ -247,41 +247,37 @@ int main(int argc, char **argv) {
...
+    // Keep only valid entries.
+    translationFileList.removeAll("");
+    translationFileList.removeAll("transl_");
+    translationFileList.removeDuplicates();
+    // Use the first translation from the list that works.
+    for (const QString &translationFile : qAsConst(translationFileList)) {
+        if (translationFile == "transl_en" || translator.load(translationFile, AppContext::getWorkingDirectoryPath())) {
             break;
         }
+        traceLogFromTranslator << "Translation not found: " + translationFile;
```
DmitriiSukhomlinov: I think, `"Translation not found: "` better be translated

src/ugeneui/src/Main.cpp
```diff
@@ -496,42 +496,38 @@ int main(int argc, char **argv) {
...
+        // Use the first translation from the list that works.
+        for (const QString &translationFile : qAsConst(translationFileList)) {
+            if (translationFile == "transl_en" || translator.load(translationFile, AppContext::getWorkingDirectoryPath())) {
                 break;
             }
-        }
-        if (!trOK) {
-            fprintf(stderr, "No translations found, exiting\n");
-            return 1;
+            traceLogFromTranslator << "Translation not found: " + translationFile;
```
DmitriiSukhomlinov: The same, better to translate it if the target language will bee different from English or Russian (who knows, maybe in the future we will support other languages)

DmitriiSukhomlinov (dismissed):
Minor comments, approved

src/ugenecl/src/Main.cpp
yalgaer: This may be a little bit tricky because I can't translate before the translator is loaded.
Tried to fix this issue by collecting file names only during translator initialization

PR #316, opened 2021-03-26
UGENE-7147. Splash screen doesn't show a minor postfix.

yalgaer:
The original fix is technically correct and works, but should be made a little bit different: no need to parse the version string in the 'update' function.

### DIFF
--- a/src/ugeneui/src/main_window/SplashScreen.cpp
+++ b/src/ugeneui/src/main_window/SplashScreen.cpp
@@ -86,7 +86,7 @@ SplashScreenWidget::SplashScreenWidget() {
     setObjectName("splash_screen_widget");
 
     Version v = Version::appVersion();
-    version = v.text;
+    version = QString::number(v.major) + (v.minor == 0 ? "" : "." + QString::number(v.minor)) + (v.suffix.isEmpty() ? "" : "-" + v.suffix);
 
     QImage image(":ugene/images/ugene_splash.png");
     QSize widgetSize = image.size();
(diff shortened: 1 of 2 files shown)


src/ugeneui/src/main_window/SplashScreen.cpp
```diff
@@ -86,7 +86,7 @@ SplashScreenWidget::SplashScreenWidget() {
     setObjectName("splash_screen_widget");
 
     Version v = Version::appVersion();
-    version = v.text;
+    version = QString::number(v.major) + (v.minor == 0 ? "" : "." + QString::number(v.minor)) + (v.suffix.isEmpty() ? "" : "-" + v.suffix);
```
ichebyki: if v.major == 38 and v.minor == 0 then
   version will be equal 380-`<suffix>`

Doesn't it?

yalgaer: When minor is 0 the suffix like ".0" should not be shown at all. 
This way we have "38" for the version 38.0 and "38.1" for the version 38.1

yalgaer: if v.major == 38 and v.minor == 0 and the suffix is "dev" then the version text shown on splash screen will be "38-dev"

ichebyki: O, yes!
I see now the `))`
You need to format the line
We have a rules!

ichebyki: It's > 80 chars length

ichebyki: $ echo `'version = QString::number(v.major) + (v.minor == 0 ? "" : "." + QString::number(v.minor)) + (v.suffix.isEmpty() ? "" : "-" + v.suffix);' | wc -c`
136

ichebyki (approved):
lgtm

src/ugeneui/src/main_window/SplashScreen.cpp
yalgaer: I reformatted this piece of code to be shorter.

We have clang-format that has all formatting rules. I try to apply it always to the files before submit.
It has no 80-chars rule and I'm not sure we really need it to be the 80-chars rule: it will make the code less readable because we will have to use either shortened unreadable variable names or a lot of forced wrapping in unexpected places. 
But you are right and the problem is valid: unlimited lines are bad.

If you care about code review experience with 1920px screen GitHub fits 120 characters per line in diffs with no wraps. This limit looks reasonable to me.

When, for some reason you have a smaller screen you can switch into "unified" diff mode - that will show even more characters per line.

src/ugeneui/src/main_window/SplashScreen.cpp
```diff
@@ -86,7 +86,9 @@ SplashScreenWidget::SplashScreenWidget() {
     setObjectName("splash_screen_widget");
 
     Version v = Version::appVersion();
-    version = v.text;
+    version = QString::number(v.major) +
+              (v.minor == 0 ? "" : "." + QString::number(v.minor)) +
+              (v.suffix.isEmpty() ? "" : "-" + v.suffix);
```
evelinabis, line 91: You can make this code a method of the Version class

evelinabis (approved):

src/ugeneui/src/main_window/SplashScreen.cpp
yalgaer: It is very specific of the SplashScreen UX that we do not want to show 0. 
Otherwise the Version class already has v.text and other places show the version like v.text.

DmitriiSukhomlinov (approved):

PR #317, opened 2021-03-28
UGENE-7153. Remove -Wimplicit-fallthrough supression from the build

yalgaer:
Some files affected files includes white-spaces reformatting by clangformat.

### DIFF
--- a/src/plugins/workflow_designer/src/ItemViewStyle.cpp
+++ b/src/plugins/workflow_designer/src/ItemViewStyle.cpp
@@ -28,7 +28,6 @@
 #include <QFontDialog>
 #include <QGraphicsSceneMouseEvent>
 #include <QGraphicsSimpleTextItem>
-#include <QGraphicsTextItem>
 #include <QGraphicsView>
 #include <QPainter>
 #include <QRadialGradient>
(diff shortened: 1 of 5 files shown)


evelinabis (approved):

src/plugins/workflow_designer/src/ItemViewStyle.cpp
```diff
@@ -316,8 +314,10 @@ bool ExtendedProcStyle::sceneEventFilter(QGraphicsItem *watched, QEvent *event)
         ret = updateCursor(he->pos());
     } break;
     case QEvent::GraphicsSceneMouseRelease:
-        desc->mouseReleaseEvent(dynamic_cast<QGraphicsSceneMouseEvent *>(event));
     case QEvent::GraphicsSceneHoverLeave:
+        if (event->type() == QEvent::GraphicsSceneMouseRelease) {
```
DmitriiSukhomlinov, line 318: This is a bit strange, I think, the previous version was better. Restore, please

DmitriiSukhomlinov (approved):
A minor comment

src/plugins/workflow_designer/src/ItemViewStyle.cpp
yalgaer: The original code has a reasonable warning: non-empty fall-thru switch cases are bad.
I reworked the code to avoid merging of these 2 cases completely (see the updated diff).

PR #324, opened 2021-03-31
UGENE-7139. Crash in SearchQualifierDialog.cpp:61

evelinabis:
https://ugene.dev/tracker/browse/UGENE-7139


src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
```diff
@@ -1922,7 +1922,9 @@ void AnnotationsTreeView::sl_onRemoveColumnByHeaderClick() {
 
 void AnnotationsTreeView::sl_searchQualifier() {
     QObjectScopedPointer<SearchQualifierDialog> d = new SearchQualifierDialog(this, this);
-    d->exec();
+    if (d->isValid()) {
```
yalgaer: We should avoid creation of new states as much as possible.
If you check for the tree/currentItem before you create the dialog you can avoid this "isValid" state in the dialog.
In this case the dialog constructor should accept a valid currentItem.

yalgaer (changes requested):
Is the problem reproducible? If yes, we need a UI test for this.
If not we need to understand/describe the root cause problem. 

I think one of the  problems is that the dialog has wrong parent: it has treeView set as a parent but internally access to the treeView->tree which can be destroyed before the treeView. This race condition in destruction may lead to other crashes when the dialog runs a task inside, receives updates from the task, treeView is not destroyed yet, but the treeView->tree and all its children does.

There is even bigger problem here: the FindQualifierTask run by the dialog access UI state from another thread.
There are 2 possible solutions here:
1) Make the task to run in main thread (a simple flag)
2) We do not need a task here at all: the search can be run in-place without any task. The FindQualifierTask searches inside of UI tree and UGENE does not support big enough UI trees to make the task block anything (it is always fast, otherwise).

These are 2 improvements I see:
1) Use correct parent.
2) Remove SearchTask or at least move it to the main thread and create another bug about this task cleanup.
(+ do not introduce a new state to the dialog - see a comment above)

evelinabis:
@yalgaer
> 1. Use correct parent.

How do I know if an parent is "correct"? Check `tree`/`currentItem` before creating the dialog?

Also help me please understand what I need to do.

1. Remove the state that I made in this commit
2. Move execution of `FindQualifierTask` to the main thread
3. Create an improvement-type issue to remove `FindQualifierTask`

Is this plan correct?

yalgaer:
I would do it like this

1) Remove the state. Move the check that you added to the constructor of the dialog to the place before you create the dialog.

2) Remove the task. Compute the result in the main thread. (the task is written the way it can be executed only in the main thread, so no task here is needed at all).

3) The last: re-check what is the correct parent for the dialog: AnnotationsTreeView or QTreeView inside of the AnnotationTreeView? I think this may be not important at all once you do 2) because there will be no parallel thread after that and no race-conditions between the main QT thread and that Task thread during AnnotationTreeView destruction. So this part becomes not important.

evelinabis:
> Is the problem reproducible?

No.

Tests: GUITest_regression_scenarios_0057

yalgaer (approved):
Good job

PR #327, opened 2021-04-05
UGENE-7024 Increase file format detection buffer size

rasputinkirill:
Checked by "format detection" test suite.

### DIFF
--- a/src/corelibs/U2Core/src/util/IOAdapterUtils.h
+++ b/src/corelibs/U2Core/src/util/IOAdapterUtils.h
@@ -33,10 +33,10 @@ class U2CORE_EXPORT IOAdapterUtils {
 public:
     static IOAdapterId url2io(const GUrl &url);
 
-    static QByteArray readFileHeader(const GUrl &url, int size = 65536);
+    static QByteArray readFileHeader(const GUrl &url, int size = 524228);
 
     // io - opened IO adapter. before and after the call pos in file the same
-    static QByteArray readFileHeader(IOAdapter *io, int size = 65536);
+    static QByteArray readFileHeader(IOAdapter *io, int size = 524228);
 
     static IOAdapter *open(const GUrl &url, U2OpStatus &os, IOAdapterMode mode = IOAdapterMode_Read, IOAdapterFactory *iof = NULL);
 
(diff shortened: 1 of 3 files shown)


src/corelibs/U2Core/src/util/IOAdapterUtils.h
```diff
@@ -33,10 +33,10 @@ class U2CORE_EXPORT IOAdapterUtils {
 public:
     static IOAdapterId url2io(const GUrl &url);
 
-    static QByteArray readFileHeader(const GUrl &url, int size = 65536);
+    static QByteArray readFileHeader(const GUrl &url, int size = 524228);
```
DmitriiSukhomlinov, line 36: Create the `static constexpr` variable with the meaningful name - now it'n not clear what is it about

DmitriiSukhomlinov (changes requested):
see comment

yalgaer:
Looks good to me.
I believe it works correctly, but it may change format detection results in real scenarios and so in our GUI tests too.

Will approve once Nightly will pass with no regressions.

src/corelibs/U2Core/src/util/IOAdapterUtils.h
rasputinkirill: IMHO No reason to do that it look like tautology. 
Because everybody know what is 'default parameter' in C++.
Also im not sure is this construction possible for compiler.
For me is look like:
X = cos(t)
t = 30°
Instead of :
X = cos(30°)

yalgaer:
I see that 3 tests are failed and 2 of them are not flaky. 
Could you please check that this fail is internal tests problem, but not the result of your change?

Or, alternatively,  re-run the build and if the second run has no same failures we will know that this is the tests problem.

src/corelibs/U2Core/src/util/IOAdapterUtils.h
DmitriiSukhomlinov: I don't like this unknown parameter, it's not obvious where it's got from. Add at least a comment

rasputinkirill:
GUI test updated. 
Because with bigger buffer size, UGENE can figure out what there is more than one sequence in file. 
Therefore it offers to choose which way to show multiple sequences.

yalgaer (changes requested):
All is good but it looks the the test 0871 fails. Could you please check?

rasputinkirill:
Test fixed and passed.

DmitriiSukhomlinov (approved):

yalgaer (approved):

PR #335, opened 2021-04-10
UGENE-6862. The "Filter SAM/BAM" element doesn't save parameters values

evelinabis:
https://ugene.dev/tracker/browse/UGENE-6862


src/corelibs/U2Designer/src/PropertyWidget.h
```diff
@@ -237,6 +237,7 @@ class U2DESIGNER_EXPORT ComboBoxWithChecksWidget : public PropertyWidget {
 
 signals:
     void valueChanged(const QString &value);
+    void si_update(QWidget *);
```
yalgaer: Please document all public API

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6044,6 +6044,110 @@ GUI_TEST_CLASS_DEFINITION(test_6860) {
...
+    // Click on an empty place on the scene
+    GTWidget::click(os, GTWidget::findWidget(os, "sceneView"));
+    // Select Filter BAM/SAM files element again
+    GTUtilsWorkflowDesigner::click(os, "Filter BAM/SAM files");
+
+    // Check that Mate strand is checked
+    GTUtilsWorkflowDesigner::clickParameter(os, "Accept flag");
+    GTUtilsWorkflowDesigner::clickParameter(os, "Accept flag");
+    checkboxValue = GTComboBox::getCurrentText(os, GTWidget::findExactWidget<QComboBox *>(os, "mainWidget"));
+    CHECK_SET_ERR(checkboxValue == "Mate strand", QString("Accept flag value (1): expected \"Mate strand\", current: \"%1\"").arg(checkboxValue))
```
yalgaer: nit: you won't need escapes if you use other kind of brackets ('', []«», etc)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6044,6 +6044,110 @@ GUI_TEST_CLASS_DEFINITION(test_6860) {
...
+    //     Expected state: Motif database value is "hpdi.xml"
+    // Click on empty place on the scene
+    // Call schema wizard
+    // Click Next 5 times
+    //     Expected state: Motif database value is "hpdi.xml"
+    // Set "cistrome.xml" as the Motif database parameter
+    // Click Next and Apply
+    //     Expected state: Motif database value is "cistrome.xml"
+
+    StartupDialogFiller *filler = new StartupDialogFiller(os);
```
yalgaer: nit: no need in local var

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6044,6 +6044,110 @@ GUI_TEST_CLASS_DEFINITION(test_6860) {
...
+    GTFileDialog::openFile(os, dataDir + "workflow_samples/NGS/cistrome/chip_seq.uwl");
+
+    GTUtilsWorkflowDesigner::click(os, "Collect Motifs with SeqPos");
+    GTUtilsWorkflowDesigner::setParameter(os, "Motif database", "hpdi.xml", GTUtilsWorkflowDesigner::ComboChecks);
+
+    // Click on the name of the Motif database parameter
+    QTableView *table = GTUtilsWorkflowDesigner::getParametersTable(os);
+    QAbstractItemModel *parametersModel = table->model();
+    CHECK_SET_ERR(parametersModel != nullptr, "ComboBox model is nullptr")
+    table->scrollTo(parametersModel->index(4, 0));
```
yalgaer: Tests are run in a separate thread. Do not use QT widgets API from a non-main thread.

src/corelibs/U2Designer/src/PropertyWidget.cpp
```diff
@@ -433,42 +432,19 @@ QVariant ComboBoxWithChecksWidget::value() {
...
-    disconnect(cm, SIGNAL(itemChanged(const QModelIndex &)), this, SLOT(sl_itemChanged(const QModelIndex &)));
-    QStringList curList = value.toString().split(",");
-    if (cm == NULL) {
-        cm = new QStandardItemModel(items.size(), 1, comboBox);
-    } else {
-        cm->clear();
-        delete cm;
-        cm = new QStandardItemModel(items.size(), 1, comboBox);
+    if (cm == nullptr) {
+        coreLog.error("Model of ComboBoxWithChecksWidget is nullptr");
```
yalgaer: Any log send to user should be wrapped with tr()

src/corelibs/U2Designer/src/PropertyWidget.cpp
```diff
@@ -433,42 +432,19 @@ QVariant ComboBoxWithChecksWidget::value() {
...
 
-    const QList<QString> &keys = items.keys();
-    int i = 0;
-
-    QStandardItem *ghostItem = new QStandardItem();
-    cm->setItem(i++, ghostItem);
-
-    foreach (const QString &key, keys) {
+    QStringList curList = value.toString().split(',', QString::SkipEmptyParts);
+    for (int i = 1; i < cm->rowCount(); i++) {
```
yalgaer, line 437: It is unclear why you start with 1 but no 0. Please add a comment.

src/corelibs/U2Designer/src/PropertyWidget.cpp
```diff
@@ -433,42 +432,19 @@ QVariant ComboBoxWithChecksWidget::value() {
...
 void ComboBoxWithChecksWidget::setValue(const QVariant &value) {
-    disconnect(cm, SIGNAL(itemChanged(const QModelIndex &)), this, SLOT(sl_itemChanged(const QModelIndex &)));
-    QStringList curList = value.toString().split(",");
-    if (cm == NULL) {
-        cm = new QStandardItemModel(items.size(), 1, comboBox);
-    } else {
-        cm->clear();
-        delete cm;
-        cm = new QStandardItemModel(items.size(), 1, comboBox);
+    if (cm == nullptr) {
```
yalgaer: This field is assigned in the constructor (initModelView method) and is never changed. Why do you want to check it for nullptr here?

src/corelibs/U2Designer/src/DelegateEditors.cpp
```diff
@@ -414,6 +414,7 @@ QWidget *ComboBoxWithChecksDelegate::createEditor(QWidget *parent,
                                                   const QModelIndex & /* index */) const {
     ComboBoxWithChecksWidget *editor = new ComboBoxWithChecksWidget(items, parent);
     connect(editor, SIGNAL(valueChanged(const QString &)), this, SIGNAL(si_valueChanged(const QString &)));
+    connect(editor, SIGNAL(si_update(QWidget *)), this, SIGNAL(commitData(QWidget *)));
```
yalgaer: ComboBoxWithChecksWidget already has si_valueChanged signal that should be emitted every time when something changes. Why do you need a new signal here?

yalgaer (changes requested):

src/corelibs/U2Designer/src/DelegateEditors.cpp
EvelinaBiserova: > ComboBoxWithChecksWidget already has si_valueChanged

Where?

evelinabis: If my changes are stupid, why does this bug exist?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
evelinabis: `GTUtilsWorkflowDesigner::setParameter` is implemented just like this one. How to do it right?

src/corelibs/U2Designer/src/DelegateEditors.cpp
evelinabis: `si_valueChanged` has the signature `void si_valueChanged(const QVariant &)`
`si_update` has the signature `void si_update(QWidget *)`

I could

1. Save the `editor` as a member of the class
2. Create a new slot for the `ComboBoxWithChecksDelegate` that `emit commitData (editor)`
3. Connect this slot to `ComboBoxWithChecksWidget::si_valueChanged`

But I can't do the first one, because `ComboBoxWithChecksDelegate::createEditor` is a const method. I am open to other ideas and discussion

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
evelinabis: Okay, I have an idea

yalgaer: You are right we have some API that works incorrectly and I'm trying to migrate it to work from the main thread when it comes to QT GUI state.

The migration priority is:
1) Migrate all places that update GUI state.
2) Migrate all places that read GUI state.

Until this long term task is finished the best we can do is not to introduce new usages, at least for the 1)

src/corelibs/U2Designer/src/DelegateEditors.cpp
yalgaer: AFAIU the widget is the sender?

evelinabis: > AFAIU the widget is the sender?

Yes

yalgaer: ComboBoxWithChecksWidget has a correct interface: it can notify clients when something is changed using si_valueChanged.

To translate this signal into another you should listen for it in the ComboBoxWithChecksDelegate and re-emit the correct one. Check the SpinBoxDelegate::createEditor

Everything else looks good. Will approve once this part is fixed.

yalgaer (approved):

PR #342, opened 2021-04-16
UGENE-7189. Fix UGENE Linux build in Teamcity from dependency on the …

yalgaer:
See comments inside + bug description in JIRA.
The patch affects only gcc build (Linux).


src/ugene_globals.pri
```diff
@@ -55,6 +57,19 @@ macx {
...
+    # version on the build machine. See https://github.com/wheybags/glibc_version_header
+    #
+    # For a wider range of supported platforms this value should be aligned with the 'glibc'
+    # used to build QT binaries.
+    #
+    # This solution is not bulletproof, because it won't replace methods from a newer 'glibc'
+    # that have no older counterparts. To address 'new API' problem we should either use a post-build 'glibc'
+    # version check for all binaries we build or/and run our pre-release tests on the old Ubuntu 16.04 host with
+    # the binary we want to release.
+    QMAKE_CFLAGS += -include $$ROOT_SRC_DIR/include/3rdparty/force_link_glibc_2.17.h
```
DmitriiSukhomlinov: I think, it's better to locate this file as "ugene/src/libs_3rdparty/glibc/force_link_glibc_2.17.h"

DmitriiSukhomlinov (changes requested):
Does it work at all?

src/ugene_globals.pri
yalgaer: Good idea. Done

yalgaer:
> Does it work at all?

What kind of answer do you expect? "No, this code does not work" ?

It works with all limits described in the comment. Locally it does allow us to build UGENE on any Linux host but it still requires testing.

evelinabis (approved):

PR #344, opened 2021-04-18
UGENE-7192. Move GUI testing related code from U2Test to GUITestBase plugin

yalgaer:
Both Linux and Windows (mini) runs were able to launch all tests successfully.
After this change UGENE core has no compile time dependencies on GUI tests so we can re-use the same build both for testing and for release (see moe details in the JIRA).

The files were moved with no changes. The changes are in the build files, the ugeneui's Main.cpp,  AppContext and in the include directives.


src/plugins/GUITestBase/GUITestBase.pro
```diff
@@ -2,6 +2,13 @@ include (GUITestBase.pri)
 
 # Input
 HEADERS +=  src/GUITestBasePlugin.h \
+            src/harness/UGUITestBase.h \
```
DmitriiSukhomlinov, line 5: I'm really sorry, but "harness" doesn't look like the name with the obvious meaning)

src/plugins/GUITestBase/src/harness/UGUITestBase.h
```diff
@@ -31,13 +31,14 @@
...
-#include <U2Test/UGUITest.h>
+#include <harness/UGUITest.h>
 
 #include <U2View/ADVSingleSequenceWidget.h>
 
 namespace U2 {
 
-class U2TEST_EXPORT UGUITestBase {
+class UGUITestBase {
+    UGUITestBase();
```
DmitriiSukhomlinov: Remove

src/plugins/GUITestBase/src/harness/UGUITestBase.cpp
```diff
@@ -25,11 +25,12 @@
...
-UGUITestBase::~UGUITestBase() {
-    qDeleteAll(tests);
-    qDeleteAll(preAdditional);
-    qDeleteAll(postAdditionalActions);
-    qDeleteAll(postAdditionalChecks);
+UGUITestBase::UGUITestBase() {
+}
+
+UGUITestBase *UGUITestBase::getInstance() {
+    static UGUITestBase *instance = new UGUITestBase();
```
DmitriiSukhomlinov, line 29: There are no problems to create 2 instances of `UGUITestBase` now - just call this function twice, but the description of this function says "There is only 1 instance of the database per application.". I offer to make it like this:

.h:
```
private:
    static UGUITestBase *instance;
```

.cpp:
```
UGUITestBase*UGUITestBase::instance = nullptr;

UGUITestBase *UGUITestBase::getInstance() {
    if (instance == nullptr) {
        instance = new UGUITestBase();
    }
    
    return instance;
}
```

DmitriiSukhomlinov (changes requested):
See comments

src/plugins/GUITestBase/GUITestBase.pro
yalgaer: https://en.wikipedia.org/wiki/Test_harness
Test harness is a common name for the code that "supports testing" (not the test itself, but the infrastructure, helpers, etc...)

src/plugins/GUITestBase/src/harness/UGUITestBase.h
yalgaer: Done

src/plugins/GUITestBase/src/harness/UGUITestBase.cpp
yalgaer: This function will never create the harness twice: the var instance is a local _static_ var.

evelinabis (approved):

src/plugins/GUITestBase/src/harness/UGUITestBase.cpp
DmitriiSukhomlinov: I don't get it, actually...

yalgaer: Sorry, I do not understand the question.

Local static vars is compact way to implement singleton pattern in thread safe environment. 
Harness initialization is a part of plugin loading  - so there is no concurrency in initialization and this pattern is valid.

The same way we implement local per-method event and performance counters (about 100+usages across UGENE's code)

PR #347, opened 2021-04-20
UGENE-7194. Show codon counts statistics in DNA Sequence View

yalgaer:
The SequenceDbiWalkerSubtask was not correctly implemented and had no use cases with non-null aminoTT or complTT before this patch. The task always produced empty results when complTT or aminoTT was used..

As a part of this fix I re-implemented SequenceDbiWalkerSubtask and removed all unused (aka untested) methods from it. I believe these methods were copied from the non-dbi version originally.


smlnkcv (commented):
There are many info in the log if open Statistic tab and select sequence region by mouse  like
[INFO][13:29] Canceling task: Sequence walker subtask
[INFO][13:29] Canceling task: Sequence walker subtask
[INFO][13:29] Canceling task: Count codons per region
[INFO][13:29] Canceling task: Sequence walker subtask

evelinabis (commented):
Is this code thread safe? Subtasks have a `run` method that is called on a separate thread (is this true?). In this method (`SequenceDbiWalkerSubtask::run`), `onRegion` is called with a pointer to the subtask. Will there be no race condition here?

Does the `BackgroundTaskRunner` run a `BackgroundTask` in a separate thread?

yalgaer:
> There are many info in the log if open Statistic tab and select sequence region by mouse like

It  sounds reasonable: all these low level logs for the tasks that were not run manually by user should go to the 'trace', but not 'info' category. I will see if the solution can be small enough to be integrated into this patch or create another one with .

yalgaer:
> Is this code thread safe?

Yes this code must be thread safe, otherwise it is a bug.

>Subtasks have a `run` method that is called on a separate thread (is this true?). 

Yes both run() and onRegion() are called from different threads of different SequenceWalker subtasks and there is a race condition here. 
But it is the standard way how sequence walker API works: it is a responsibility of onRegion method implementation to handle data safely in this case: do not use a global state or use some kind of synchronization.

> Does the `BackgroundTaskRunner` run a `BackgroundTask` in a separate thread?

It depends on BackgroundTask flags. In this case I set TaskFlags_NoRun - the run method is not called at all and no new thread is created.

evelinabis (dismissed):

PR #348, opened 2021-04-20
UGENE-7092 UGENE doesn't open the GenBank file

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7092

### DIFF
--- a/src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
+++ b/src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
 }
 
 bool GenbankPlainTextFormat::readIdLine(ParserState *st) {
-    if (!st->hasKey("LOCUS")) {
+    if (!st->hasKey(DNAInfo::LOCUS.toLocal8Bit())) {
         QByteArray rawData(st->buff);
-        int locusStartPos = rawData.indexOf("\nLOCUS");
-        if (-1 == locusStartPos) {
-            st->si.setError(tr("LOCUS is not the first line"));
-            return false;
-        } else {
+        int locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
+        if (locusStartPos != -1) { //We are here if the "GenBank" file has some pre-description (see UGENE-4463)
             while (locusStartPos >= st->len) {
                 st->readNextLine();
                 rawData = QByteArray(st->buff);
-                locusStartPos = rawData.indexOf("\nLOCUS");
+                locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
             }
             st->buff = st->buff + locusStartPos;
+        } else {
+            //The GenBank file should have the indent (st->valOffset) before the real data
+            //Possibly, this GenBank file doesn't have the strong indent (UGENE-7092)
+            rawData = QByteArray(st->buff, st->len);
+            int locusStartPos = rawData.indexOf(DNAInfo::LOCUS.toLocal8Bit());
+            CHECK_EXT(locusStartPos != -1, st->si.setError(tr("LOCUS is not the first line")), false);
+
+            rawData = rawData.right(rawData.size() - DNAInfo::LOCUS.size());
+            int locusAndNameGapSize = 0;
+            while (rawData.front() == ' ') {
+                rawData = rawData.right(rawData.size() - 1);
+                locusAndNameGapSize++;
+            }
+            int indentBeforeName = locusStartPos + DNAInfo::LOCUS.size() + locusAndNameGapSize;
+            int shiftToLeft = st->valOffset - indentBeforeName;
+            st->buff -= shiftToLeft;
         }
     }
 
--- a/src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
+++ b/src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
@@ -85,30 +85,33 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
 }
 
 bool GenbankPlainTextFormat::readIdLine(ParserState *st) {
-    if (!st->hasKey(DNAInfo::LOCUS.toLocal8Bit())) {
+    static const QByteArray LOCUS = DNAInfo::LOCUS.toLocal8Bit();
+    if (!st->hasKey(LOCUS)) {
         QByteArray rawData(st->buff);
-        int locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
+        int locusStartPos = rawData.indexOf("\n" + LOCUS);
         if (locusStartPos != -1) { //We are here if the "GenBank" file has some pre-description (see UGENE-4463)
             while (locusStartPos >= st->len) {
                 st->readNextLine();
                 rawData = QByteArray(st->buff);
-                locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
+                locusStartPos = rawData.indexOf("\n" + LOCUS);
             }
             st->buff = st->buff + locusStartPos;
         } else {
             //The GenBank file should have the indent (st->valOffset) before the real data
             //Possibly, this GenBank file doesn't have the strong indent (UGENE-7092)
-            rawData = QByteArray(st->buff, st->len);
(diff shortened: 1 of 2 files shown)


smlnkcv (commented):
Approve

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
                 rawData = QByteArray(st->buff);
-                locusStartPos = rawData.indexOf("\nLOCUS");
+                locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
             }
             st->buff = st->buff + locusStartPos;
+        } else {
+            //The GenBank file should have the indent (st->valOffset) before the real data
+            //Possibly, this GenBank file doesn't have the strong indent (UGENE-7092)
+            rawData = QByteArray(st->buff, st->len);
+            int locusStartPos = rawData.indexOf(DNAInfo::LOCUS.toLocal8Bit());
```
yalgaer: Why do you want to re-init rawData here again and search for locusStartPoint again? You already have this value assigned above.

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
 bool GenbankPlainTextFormat::readIdLine(ParserState *st) {
-    if (!st->hasKey("LOCUS")) {
+    if (!st->hasKey(DNAInfo::LOCUS.toLocal8Bit())) {
         QByteArray rawData(st->buff);
-        int locusStartPos = rawData.indexOf("\nLOCUS");
-        if (-1 == locusStartPos) {
-            st->si.setError(tr("LOCUS is not the first line"));
-            return false;
-        } else {
+        int locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
```
yalgaer: rawData is QByteArray and parameter of the indexOf is QString. The conversion is implicit.
I would suggest to cache LOCUS.toLocal8Bit from line 88 and re-use that QByteArray here.

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
+            //The GenBank file should have the indent (st->valOffset) before the real data
+            //Possibly, this GenBank file doesn't have the strong indent (UGENE-7092)
+            rawData = QByteArray(st->buff, st->len);
+            int locusStartPos = rawData.indexOf(DNAInfo::LOCUS.toLocal8Bit());
+            CHECK_EXT(locusStartPos != -1, st->si.setError(tr("LOCUS is not the first line")), false);
+
+            rawData = rawData.right(rawData.size() - DNAInfo::LOCUS.size());
+            int locusAndNameGapSize = 0;
+            while (rawData.front() == ' ') {
+                rawData = rawData.right(rawData.size() - 1);
```
yalgaer: This code may be good if QByteArray does not create a full copy. Do you know if the full copy is created every time you call .right()? Otherwise you can use fromRawData with offset here instead of the 'right'

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
                 st->readNextLine();
                 rawData = QByteArray(st->buff);
-                locusStartPos = rawData.indexOf("\nLOCUS");
+                locusStartPos = rawData.indexOf(QString("\n%1").arg(DNAInfo::LOCUS));
             }
             st->buff = st->buff + locusStartPos;
+        } else {
+            //The GenBank file should have the indent (st->valOffset) before the real data
+            //Possibly, this GenBank file doesn't have the strong indent (UGENE-7092)
+            rawData = QByteArray(st->buff, st->len);
```
yalgaer: Use QByteArray::fromRawData to avoid deep copy.
The same is true for line 88.

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
+            int locusStartPos = rawData.indexOf(DNAInfo::LOCUS.toLocal8Bit());
+            CHECK_EXT(locusStartPos != -1, st->si.setError(tr("LOCUS is not the first line")), false);
+
+            rawData = rawData.right(rawData.size() - DNAInfo::LOCUS.size());
```
yalgaer: Why do you shift right to LOCUS.size() here but do not use locusStartPos offset?

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,32 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
+            int locusAndNameGapSize = 0;
```
yalgaer: nit: rename to 'gapBetweenLocusAndName' ? 'And' does not say a lot where the gap is : may read like a gap after locusAndName.

yalgaer (changes requested):

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: Done

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: Done

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: It's a bit different. We are looking for "\nLOCUS" above, and here I suggest, that "LOCUS" could not be at the beginning of the line.  I renamed the variable to avoid misunderstandings.

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: I checked this scenario - if "LOCUS" isn't located in the beginning of the line the file wouldn't be recognized as "GenBank" at all. In this case, we can suggest, that "LOCUS" is in the beginning and `locusStartPos` equals 0.

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: Done

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: Done

smlnkcv (commented):
Approved again

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
```diff
@@ -85,19 +85,35 @@ bool GenbankPlainTextFormat::isStreamingSupport() {
...
+            do {
+                if (indentBeforeName == 0) {
+                    indentBeforeName = LOCUS.size();
+                } else {
+                    indentBeforeName++;
+                }
+                rawData = QByteArray::fromRawData(st->buff + indentBeforeName, st->len - indentBeforeName);
+            } while (rawData.front() == ' ');
+            int shiftToLeft = st->valOffset - indentBeforeName;
+            st->buff -= shiftToLeft;
```
evelinabis, line 116: Do I understand correctly that you want to go beyond the beginning of the array?

evelinabis (approved):

DmitriiSukhomlinov:
@yalgaer

src/corelibs/U2Formats/src/GenbankPlainTextFormat.cpp
DmitriiSukhomlinov: Yes, you are right

yalgaer (commented):
All is good will approve after tests are passed

yalgaer (approved):

PR #350, opened 2021-04-22
UGENE-6170 Correct names of document formats in two workflow elements

rasputinkirill:


src/plugins/workflow_designer/src/library/ConvertFilesFormatWorker.cpp
```diff
@@ -187,7 +186,14 @@ void ConvertFilesFormatWorker::init() {
     inputUrlPort = ports.value(INPUT_PORT);
     outputUrlPort = ports.value(OUTPUT_PORT);
     targetFormat = getValue<QString>(BaseAttributes::DOCUMENT_FORMAT_ATTRIBUTE().getId());
-    excludedFormats = getValue<QString>(EXCLUDED_FORMATS_ID).split(",", QString::SkipEmptyParts);
+    QStringList excludedFormatNames = getValue<QString>(EXCLUDED_FORMATS_ID).split(",", QString::SkipEmptyParts);
```
yalgaer: BaseAttributes::DOCUMENT_FORMAT_ATTRIBUTE  has type and the type is "DOCUMENT_FORMAT_ATTR_ID"
It can't be name it must always be id.

The correct fix is to change the rendering of the format id in combo-box-> use "name" as a rendered value, but do not change what is passed inside WD bus during workflow run.

The WD bus content may be passed not only from the combo-box you tweaked, but also from the command line params, WD-schema file, etc.

yalgaer (changes requested):

src/corelibs/U2Designer/src/PropertyWidget.cpp
```diff
@@ -490,6 +494,26 @@ void ComboBoxWithChecksWidget::initModelView() {
     comboBox->setView(vw);
 }
 
+/************************************************************************/
```
yalgaer: This code is not needed. Remove?

src/corelibs/U2Designer/src/DelegateEditors.h
```diff
@@ -325,9 +325,14 @@ private slots:
 class U2DESIGNER_EXPORT ComboBoxWithChecksDelegate : public PropertyDelegate {
     Q_OBJECT
 public:
-    ComboBoxWithChecksDelegate(const QVariantMap &items, QObject *parent = 0)
+    ComboBoxWithChecksDelegate(const QVariantMap &items, QObject *parent = nullptr)
         : PropertyDelegate(parent), items(items) {
     }
+
+    ComboBoxWithChecksDelegate(const QVariantMap &items, const QVariantMap &visibleKeyToBusValueTranslationMap, QObject *parent = nullptr)
```
yalgaer: nit: word 'translation' is not needed here

src/corelibs/U2Designer/src/DelegateEditors.h
```diff
@@ -352,6 +357,7 @@ private slots:
 
 protected:
     QVariantMap items;
+    QVariantMap visibleKeyToBusValueTranslationMap;
```
yalgaer: Why do you have QVariant as a value here? Looks like it can be a QString->QString map?

src/plugins/workflow_designer/src/library/ConvertFilesFormatWorker.cpp
```diff
@@ -99,12 +99,11 @@ QVariantMap getFormatsMap(MapType mapType) {
...
         }
-        if (format->checkFlags(DocumentFormatFlag_SupportWriting) || (BOOLEANS == mapType)) {
-            if (BOOLEANS == mapType) {
-                result[fid] = false;
-            } else {
-                result[fid] = fid;
-            }
+        const QString formatName = format->getFormatName();
+        if (BOOLEANS == mapType) {
+            result[formatName] = false;
```
yalgaer, line 104: Worker should not deal with display values  (?)
The map returned here should use ids

src/plugins/workflow_designer/src/library/FilterBamWorker.cpp
```diff
@@ -181,8 +182,8 @@ void FilterBamWorkerFactory::init() {
         delegates[CUSTOM_DIR_ID] = new URLDelegate("", "", false, true);
 
         QVariantMap formatMap;
-        formatMap[BaseDocumentFormats::BAM] = BaseDocumentFormats::BAM;
-        formatMap[BaseDocumentFormats::SAM] = BaseDocumentFormats::SAM;
+        formatMap[AppContext::getDocumentFormatRegistry()->getFormatById(BaseDocumentFormats::BAM)->getFormatName()] = BaseDocumentFormats::BAM;
```
yalgaer, line 185: Same here: the worker should use ids, not names

yalgaer (changes requested):

src/plugins/workflow_designer/src/library/ConvertFilesFormatWorker.cpp
rasputinkirill: Why not? Only the worker knows what data should be in the combo box.

src/corelibs/U2Designer/src/DelegateEditors.h
```diff
@@ -352,6 +357,7 @@ private slots:
 
 protected:
     QVariantMap items;
+    QMap<QString, QString> visibleKeyToBusValueMap;
```
yalgaer, line 360: This name is misleading. The correct name is "busValueToDisplayNameMap": by the value in the bus (format-id here) you can get the display value for the combo-box: format name.

src/corelibs/U2Designer/src/PropertyWidget.cpp
```diff
@@ -423,16 +423,22 @@ ComboBoxWithChecksWidget::ComboBoxWithChecksWidget(const QVariantMap &_items, QW
 QVariant ComboBoxWithChecksWidget::value() {
     QStringList sList;
     const QList<QString> &keys = items.keys();
-    foreach (const QString &key, keys) {
+    for (const QString &key: qAsConst(keys)) {
```
yalgaer, line 426: 'keys' are already bus values here.
The 'keys' should be converted to display names here:
QStandardItem *item = new QStandardItem(keyToDisplayNameMap[it.key()]);

src/plugins/workflow_designer/src/library/ConvertFilesFormatWorker.cpp
yalgaer: Everywhere inside WD (runtime, uwl file) you should use 'format-id' and only in the visual components (combo box) show a 'rendered' form of the format-id. It is "no-change" from the current behavior.

The only change in this patch should be inside of the combo box component: it should take format id and show a display value for it from the map you added. Once user select a value in the combo box the result returned by the combo-box should have only format-ids as it is now.

yalgaer (changes requested):

src/corelibs/U2Designer/src/PropertyWidget.h
```diff
@@ -249,6 +249,14 @@ protected slots:
...
 private:
     void initModelView();
+
+private:
+    QMap<QString, QString> visibleKeyToBusValueMap;
+    /*
+    * variable to get 'visible value' when it needed instead of writing very similar value() method
+    * also we cant change value() signature because it virtual
+    */
+    bool getVisibleValue;
```
DmitriiSukhomlinov, line 259: Don't use `get` , `set`, `is` or `has` in the beginning of variable names - leave it for getters and setters. Name this variable differently, e.g. `visibleValueHas2BeGot2`, but it's a bit complected too, so, please, turn on your imagination)

PR #353, opened 2021-04-24
UGENE-7197. Multiline molecule name in pdb-file is parsed incorrectly

evelinabis:
[Issue](https://ugene.dev/tracker/browse/UGENE-7197)


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -321,6 +322,24 @@ GUI_TEST_CLASS_DEFINITION(test_7152) {
     CHECK_SET_ERR(bottomRight == "11/40/35", "Bottom right position is wrong: " + bottomRight);
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7197) {
```
yalgaer: What is the reason to have GUI (heavy & slow) test for parsers? This should be an XML test.

src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -280,12 +281,22 @@ void PDBFormat::PDBParser::parseMacromolecularContent(bool firstCompndLine, U2Op
     list
     Details
     */
+
+    // Returns the end-of-name index for `specification`.
+    auto endOfNameInd = [&](const QString &specification) {
+        int index = QRegExp(";\\s*$").indexIn(specification);
```
yalgaer: Here you create a new reg-ex object on every invocation (may be in loop) which is anti-pattern
Normally you should re-use reg-ex objects. You can make is function static, or a class field.

yalgaer (changes requested):

smlnkcv (commented):
Approved

src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -606,6 +609,15 @@ void PDBFormat::PDBParser::createMolecule(char chainIdentifier, BioStruct3D &bio
 void PDBFormat::PDBParser::updateResidueIndexes(BioStruct3D & /*biostruc*/) {
 }
 
+int PDBFormat::PDBParser::endOfNameInd(const QString &specification) {
+    int index = QRegExp(";\\s*$").indexIn(specification);
```
yalgaer: Could you please re-use regex instance and avoid new instance creation on every method call?

src/corelibs/U2Formats/src/PDBFormat.h
```diff
@@ -99,6 +103,9 @@ class U2FORMATS_EXPORT PDBFormat : public TextDocumentFormatDeprecated {
         bool seqResContains(char chainIdentier, int residueIndex, char acronym);
         QByteArray getNextSpecLine();
 
+        // Returns the end-of-name index for `specification`.
+        int endOfNameInd(const QString &specification);
```
yalgaer: The method is not a simple getter but changes the internal parser state.
Please update the doc and change the name accordingly.

src/plugins/CoreTests/src/BioStruct3DObjectTests.h
```diff
@@ -87,6 +87,17 @@ class GTest_BioStruct3DAtomResidueName : public XmlTest {
     int modelId;
 };
 
+class GTest_BioStruct3DMoleculeName : public XmlTest {
+    Q_OBJECT
+    SIMPLE_XML_TEST_BODY_WITH_FACTORY(GTest_BioStruct3DMoleculeName, "check-biostruct3d-molecule-name");
+
+    ReportResult report();
```
yalgaer: add override

src/plugins/CoreTests/src/BioStruct3DObjectTests.h
```diff
@@ -87,6 +87,17 @@ class GTest_BioStruct3DAtomResidueName : public XmlTest {
     int modelId;
 };
 
+class GTest_BioStruct3DMoleculeName : public XmlTest {
```
yalgaer, line 98: Add class-level doc + usage example

src/plugins/CoreTests/src/BioStruct3DObjectTests.h
```diff
@@ -87,6 +87,17 @@ class GTest_BioStruct3DAtomResidueName : public XmlTest {
...
+    Q_OBJECT
+    SIMPLE_XML_TEST_BODY_WITH_FACTORY(GTest_BioStruct3DMoleculeName, "check-biostruct3d-molecule-name");
+
+    ReportResult report();
+
+    QString objContextName;
+    QString molName;
+    int chainInd;
```
yalgaer: nit: use fully qualified names: 'Index', not 'Ind' when the length difference is not significant.

src/plugins/CoreTests/src/BioStruct3DObjectTests.cpp
```diff
@@ -391,6 +393,63 @@ Task::ReportResult GTest_BioStruct3DAtomResidueName::report() {
...
+    if (!v.isEmpty()) {
+        bool ok = false;
+        chainInd = v.toInt(&ok);
+        if (!ok) {
+            stateInfo.setError(QString("invalid value type %1, int required").arg(CHAIN_IND_ATTR));
+        }
+    }
+
+    // molecule name
+    v = el.attribute(MOL_NAME_ATTR);
```
yalgaer: nit: use molName=el.attr.. directly (same as for the objContextName above)

src/plugins/CoreTests/src/BioStruct3DObjectTests.cpp
```diff
@@ -391,6 +393,63 @@ Task::ReportResult GTest_BioStruct3DAtomResidueName::report() {
 
     return ReportResult_Finished;
 }
+///////////////////////////////////////////////////////////////////////////////////////////
+
+void GTest_BioStruct3DMoleculeName::init(XMLTestFormat *tf, const QDomElement &el) {
+    Q_UNUSED(tf);
```
yalgaer: Do not declare 'tf' name in the params list and avoid Q_UNUSED

src/plugins/CoreTests/src/BioStruct3DObjectTests.cpp
```diff
@@ -38,6 +38,8 @@ namespace U2 {
 #define X_COORD_ATTR "x"
 #define Y_COORD_ATTR "y"
 #define Z_COORD_ATTR "z"
+#define CHAIN_IND_ATTR "chain-ind"
```
yalgaer: Use full name: "chain-index"

yalgaer (changes requested):

src/corelibs/U2Formats/src/PDBFormat.cpp
DmitriiSukhomlinov: The best way is to create a `static const` variable

src/corelibs/U2Formats/src/PDBFormat.h
DmitriiSukhomlinov: There is no need to short "index" as "ind", rename, please

src/plugins/CoreTests/src/BioStruct3DObjectTests.cpp
DmitriiSukhomlinov: I can argue with that. We don't have any agreements about unused variables, actually, so, for example, I prefer wrap them with `Q_UNUSED` too

src/corelibs/U2Formats/src/PDBFormat.h
evelinabis: The new title looks awful, but the conversation is resolved.

src/plugins/CoreTests/src/BioStruct3DObjectTests.cpp
evelinabis: `tf` declaration removed.

rasputinkirill (approved):

yalgaer (approved):

smlnkcv (approved):

PR #359, opened 2021-04-29
UGENE-7183 Crash on sequence exporting

rasputinkirill:
For some reason method QMap::contains(const Key &key) works wrong with U2EntityRef, so i've replaced it with my own method.


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -321,6 +328,40 @@ GUI_TEST_CLASS_DEFINITION(test_7152) {
     CHECK_SET_ERR(bottomRight == "11/40/35", "Bottom right position is wrong: " + bottomRight);
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7183) {
+    class ExportSequencesScenario : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus &os) {
```
yalgaer: nit: add override to virtual functions

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -321,6 +328,40 @@ GUI_TEST_CLASS_DEFINITION(test_7152) {
...
+            QWidget *dialog = QApplication::activeModalWidget();
```
yalgaer: Use GTWidget::getActiveModalWidget so you do not need to check for nulls

src/plugins/dna_export/src/ExportSequenceTask.h
```diff
@@ -70,6 +70,7 @@ class ExportSequenceItem {
 
     void releaseOwnedSeq();
 
+    static bool isRefAlreadyCounted(const U2EntityRef &seqRef);
```
yalgaer: If you do not need this method in the public API (aka header file) make it a cpp-local.

src/plugins/dna_export/src/ExportSequenceTask.cpp
```diff
@@ -109,29 +109,38 @@ void ExportSequenceItem::releaseOwnedSeq() {
     }
 }
 
+bool ExportSequenceItem::isRefAlreadyCounted(const U2EntityRef& seqRef) {
```
yalgaer: How this method is different from 'sequencesRefCount.contains(seqRef)' ?

yalgaer (changes requested):

src/plugins/dna_export/src/ExportSequenceTask.cpp
rasputinkirill: My method use operator == which work correct.
QMap::contains() use hash method or something like what  leads to wrong result - it returns 'true' for object which is not present in container.

src/plugins/dna_export/src/ExportSequenceTask.h
rasputinkirill: I'm not agree with this comment.
Because, in this case we can get rid with all private static methods in that way.
But as i understand this is OOP way - to bound methods with entities(classes).
And you suggest to make 'homeless' methods.

smlnkcv (dismissed):
No crash after more than 10 exporting action

src/plugins/dna_export/src/ExportSequenceTask.h
yalgaer: What is wrong with private static (local) methods? I think they are really good and you can change your mind about them.

A header is a part of the public interface. Public API is a reason why you have the header, otherwise the whole class can be defined in the related CPP file. Adding a private code into a header makes it less readable and adds extra dependencies to users.

If you do not like private statics (aka locals) you can use PImpl pattern (used across QT sources) but in this case you will have to write more boilerplate code: PImpls are good for a statefull code.

src/plugins/dna_export/src/ExportSequenceTask.cpp
yalgaer: sequencesRefCounts is a QMap. 
Is it safe to have 'U2EntityRef' as a key in QMap? 

If yes: your code is not needed and map.contains(key) should be enough.

If it is not safe: this is the real cause of the bug: we need to make U2EntityRef work with QMap as expected or never use it (across all our code) as a key in QMap.

src/corelibs/U2Core/src/datatype/U2Type.cpp
```diff
@@ -119,7 +119,7 @@ bool U2EntityRef::operator!=(const U2EntityRef &other) const {
 }
 
 bool U2EntityRef::operator<(const U2EntityRef &other) const {
-    return dbiRef.dbiFactoryId + dbiRef.dbiId + entityId + QString::number(version) < other.dbiRef.dbiFactoryId + other.dbiRef.dbiId + other.entityId + QString::number(other.version);
+    return dbiRef.dbiFactoryId.toUtf8() + dbiRef.dbiId.toUtf8() + entityId + QString::number(version).toUtf8() < other.dbiRef.dbiFactoryId.toUtf8() + other.dbiRef.dbiId.toUtf8() + other.entityId + QString::number(other.version).toUtf8();
```
yalgaer: This looks like the correct fix. 
I suggest to make it a little bit cleaner: compare each values separately: no need to concat them at all:

```
if (dbiRef.dbiFactoryId != othetr.dbiRef.factoryId) {
    return dbiRef.dbiFactoryId < othetr.dbiRef.factoryId
}
if (...)
```

this code is longer but has 2 advantages: 
1) it has no hidden bugs with conversions like you fixed above
2) it will guarantee the order: factories compared first, dbIds next, etc... The current impl does not guarantee this order

yalgaer (changes requested):

src/corelibs/U2Core/src/datatype/U2Type.cpp
rasputinkirill: Do you agree with my idea of comparison order:
1. dbiRef.dbiFactoryId
2. dbiRef.dbiId
3. version
4. entityId
I think entityId should be the 'weakest' one.

yalgaer (approved):

PR #360, opened 2021-04-29
UGENE-6802 Assert on 'Calculate assembly coverage' task in Workflow Designer

rasputinkirill:


yalgaer (changes requested):
1. Error message should be changed to something like "Assembly length must be greater than zero". 
2.  Since this one is a user facing message a valid question is if we can test it in GUI/XML tests?

smlnkcv (commented):
Approved. No assert now.

yalgaer (approved):

PR #364, opened 2021-05-01
UGENE-7105 Show symbol of the second strongest chromatogram trace

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7105


src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <U2Core/U2SafePoints.h>
+#include <U2View/MSAEditor.h>
+#include <U2Gui/ShowHideSubgroupWidget.h>
+
+#include <QVBoxLayout>
+
+namespace U2 {
+
+const QString McaReadsTabFactory::GROUP_ID = "OP_MCA_READS";
+const QString McaReadsTabFactory::GROUP_ICON_STR = ":core/images/todo.png";
```
DmitriiSukhomlinov: https://ugene.dev/tracker/browse/UGENE-7218

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,69 @@
...
+const QString McaReadsTabFactory::GROUP_DOC_PAGE = "TODO";
```
DmitriiSukhomlinov, line 36: https://ugene.dev/tracker/browse/UGENE-7219

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
```diff
@@ -59,6 +69,8 @@ class U2CORE_EXPORT DNAChromatogram {
 
     static const ushort INVALID_VALUE;
     static const char DEFAULT_PROBABILITY;
+    /* The @U2::DNAChromatogram::Trace enum and the corresponding @char symbol */
```
yalgaer: What is the '@char symbol' here? Is it a visual representation of the 'Trace' or a character from the DNA alphabet?
If it is the visual one - it should not be in the data struct, but in the visual struct.
If it is the alphabet one - the comment should say about alphabet, not about '@char' (aka C++ raw type)

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
```diff
@@ -30,13 +30,23 @@ namespace U2 {
 
 class U2CORE_EXPORT DNAChromatogram {
 public:
-    enum Trace {
+    enum class Trace {
```
yalgaer, line 33: Is there any benefit of using enum to repeat 'A', 'C', 'G', 'T' constants from the DNA alphabet here?
We do not use this renaming in other places and use 'A','C','G','T' directly.

I know it is not you who introduced it, but you are building more code on top of it, so may be building new code a simpler way will be a better investment.

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
```diff
@@ -30,13 +30,23 @@ namespace U2 {
...
     };
 
+    /* Chromatogram trace and the corresponding peak height */
+    struct ChromatogramTraceAndValue {
+        ChromatogramTraceAndValue(Trace t, ushort v) : trace(t), value(v) {}
+
+        // Chromatogram trace
+        Trace trace = Trace::Trace_A;
+        //Height of the @trace peak
+        ushort value = 0;
```
yalgaer: Keep non-standard types to the serializers. We can use simple common int here to avoid warnings about type conversions inside our code.

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
```diff
@@ -346,6 +352,23 @@ qint64 MultipleChromatogramAlignmentRowData::getBaseCount(qint64 before) const {
...
+                MultipleChromatogramAlignmentRowData::getTwoHighestPeaks(const qint64 position) const {
+    const int baseCall = chromatogram.baseCalls[position];
+    QList<DNAChromatogram::ChromatogramTraceAndValue> peaks =
+                                    { { DNAChromatogram::Trace::Trace_A, chromatogram.A[baseCall] },
+                                      { DNAChromatogram::Trace::Trace_C, chromatogram.C[baseCall] },
+                                      { DNAChromatogram::Trace::Trace_G, chromatogram.G[baseCall] },
+                                      { DNAChromatogram::Trace::Trace_T, chromatogram.T[baseCall] } };
+    std::sort(peaks.begin(),
+              peaks.end(),
+        [](const DNAChromatogram::ChromatogramTraceAndValue& first,
```
yalgaer: nit: good place to use 'auto' instead of the long names. The 'peaks' is defined 1 lines above.

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
```diff
@@ -346,6 +352,23 @@ qint64 MultipleChromatogramAlignmentRowData::getBaseCount(qint64 before) const {
...
+            const DNAChromatogram::ChromatogramTraceAndValue& second) {
+        return first.value > second.value;
+    });
+    return { peaks.first(), peaks.at(1) };
```
yalgaer: nit: peaks[0], peaks[1] would be more readable than using names from difference name spaces.

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.h
```diff
@@ -100,6 +100,9 @@ class U2CORE_EXPORT MultipleChromatogramAlignmentRowData : public MultipleAlignm
     const DNAChromatogram &getChromatogram() const;
     DNAChromatogram getGappedChromatogram() const;
 
+    /** Returns the position of @pos, including gaps */
+    int getGappedPosition(int pos) const;
```
yalgaer: int -> qint64

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.h
```diff
@@ -191,6 +194,9 @@ class U2CORE_EXPORT MultipleChromatogramAlignmentRowData : public MultipleAlignm
      */
     qint64 getBaseCount(qint64 before) const;
 
+    /** Returns pair of the first and the second (by peak height) chromatogram trace characted in the @pos position */
+    QPair<DNAChromatogram::ChromatogramTraceAndValue, DNAChromatogram::ChromatogramTraceAndValue> getTwoHighestPeaks(const qint64 position) const;
```
yalgaer: nit: example when "const everywhere" makes public API looks bad:
As a client of this function I do not care if you accept this 'int' as const or not, but I have to read it.

Overall I would suggest use const only when it is important and do not worry a lot about it for the locals in trivial methods.

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
```diff
@@ -260,6 +260,47 @@ void MultipleChromatogramAlignmentObject::trimRow(const int rowIndex, int curren
     updateCachedMultipleAlignment(modificationInfo);
 }
 
+void MultipleChromatogramAlignmentObject::updateAlternativeMutations(const McaReadsTabSettings& settings, U2OpStatus& os) {
+    for (int i = 0; i < getNumRows(); i++) {
+        const MultipleChromatogramAlignmentRow& mcaRow = static_cast<const MultipleChromatogramAlignmentRow&>(getRow(i));
+        const int ungappedLength = mcaRow->getUngappedLength();
```
yalgaer: int -> qint64

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
```diff
@@ -260,6 +260,47 @@ void MultipleChromatogramAlignmentObject::trimRow(const int rowIndex, int curren
...
+
+        QMap<qint64, char> newCharList;
+        for (int j = 0; j < ungappedLength; j++) {
+            auto res = mcaRow->getTwoHighestPeaks(j);
+            const double minimumThresholdValue = (double)res.second.value / res.first.value * 100;
```
yalgaer: nit: + same 'const' madeness here. You do not use a consistent style with const even in this method (see gappedPos or modifiedRowId below) so it is better not to use const for locals with no important reason at all. 

More than that other UGENE code does not follow the convention of 'const everywhere' so we have 2 code styles across our code which is bad: we should use 1 unified code style and 'const everywhere' breaks it.

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
```diff
@@ -260,6 +260,47 @@ void MultipleChromatogramAlignmentObject::trimRow(const int rowIndex, int curren
     updateCachedMultipleAlignment(modificationInfo);
 }
 
+void MultipleChromatogramAlignmentObject::updateAlternativeMutations(const McaReadsTabSettings& settings, U2OpStatus& os) {
+    for (int i = 0; i < getNumRows(); i++) {
+        const MultipleChromatogramAlignmentRow& mcaRow = static_cast<const MultipleChromatogramAlignmentRow&>(getRow(i));
+        const int ungappedLength = mcaRow->getUngappedLength();
+
+        QMap<qint64, char> newCharList;
```
yalgaer: Why map but not hash?

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
```diff
@@ -260,6 +260,47 @@ void MultipleChromatogramAlignmentObject::trimRow(const int rowIndex, int curren
...
+
+        const MultipleAlignment& ma = getMultipleAlignment();
+        qint64 modifiedRowId = ma->getRow(i)->getRowId();
+        McaDbiUtils::replaceCharactersInRow(getEntityRef(), modifiedRowId, newCharList, os);
+        SAFE_POINT_OP(os, );
+    }
+    updateCachedMultipleAlignment();
+}
+
+void MultipleChromatogramAlignmentObject::sl_viewIsAbout2BeDestroyed() {
```
yalgaer: What if I have 2 views opened with different settings?

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.h
```diff
@@ -70,10 +71,21 @@ class U2CORE_EXPORT MultipleChromatogramAlignmentObject : public MultipleAlignme
     void deleteColumnsWithGaps(U2OpStatus &os);
 
     void trimRow(const int rowIndex, int currentPos, U2OpStatus &os, TrimEdge edge);
+
+    /**
+      * Updates the corresponding alternative mutations.
+      * Shows the second strongest character, if it's peak height persantage is more then @settings.threshold.
```
yalgaer: Data object should not mess with visual mode here: 'shows' is related  to the view state, not to the object state.

yalgaer (changes requested):

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
DmitriiSukhomlinov: Well, I suppose, it's much more convenient to use `enum` in `switch-case` statements. Also it visualizes the fact that we have only four optional characters here. And this `enum` is widely used in the code, so I don't think that there's a sense to get rid of it.

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
DmitriiSukhomlinov: There are a lot `usort` here and I replaced them all, if there are connected

DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
DmitriiSukhomlinov: I don't know what means "C++ raw type"

DmitriiSukhomlinov: And the idea of the comment is't clear for me. Actually, I didn't want this comment at all, cuz. as far as I'm concerned, this is a very obvious member. What is "visual" and "alphabet" means in you comment?

DmitriiSukhomlinov: It'll be easier if you just text me the comment you want here

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
DmitriiSukhomlinov: Done

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
DmitriiSukhomlinov: Done

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.h
DmitriiSukhomlinov: Ok

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.h
DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: OK, but we don't have any agreements about it.

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: `MultipleChromatogramAlignmentObject` instance "exists" in one view. If you open two Sanger Reads Editor, you will have two instances of `MultipleChromatogramAlignmentObject`, each of them have their own settings saved.

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.h
DmitriiSukhomlinov: Ok

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: Done

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: Well.... OK, replaced, now this code executes 0.01 seconds faster)

src/corelibs/U2Core/src/datatype/DNAChromatogram.h
DmitriiSukhomlinov: `char` is a type, I don't now how to distinguish it better... Ok, fixed

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
```diff
@@ -346,6 +352,22 @@ qint64 MultipleChromatogramAlignmentRowData::getBaseCount(qint64 before) const {
...
+                MultipleChromatogramAlignmentRowData::getTwoHighestPeaks(qint64 position) const {
+    const int baseCall = chromatogram.baseCalls[position];
+    QList<DNAChromatogram::ChromatogramTraceAndValue> peaks =
+                                    { { DNAChromatogram::Trace::Trace_A, chromatogram.A[baseCall] },
+                                      { DNAChromatogram::Trace::Trace_C, chromatogram.C[baseCall] },
+                                      { DNAChromatogram::Trace::Trace_G, chromatogram.G[baseCall] },
+                                      { DNAChromatogram::Trace::Trace_T, chromatogram.T[baseCall] } };
+    std::sort(peaks.begin(),
+              peaks.end(),
+        [](const auto& first, const auto& second) {
```
DmitriiSukhomlinov, line 396: At least C++14 has to be turned on on Linux and macOS, I don't want do this change in this very issue, so I rollback `auto`

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
DmitriiSukhomlinov: At least C++14 must be enabled on Linux and macOS

yalgaer: Lets update the compatibility level in our config if needed.

We should care about the minimal supported compiler version here, not about the C++ standard version, because different version of compilers supports different sub-sets of each standard.

For example for Linux we use gcc 5.4 (Ubuntu 16.04) as a minimal version. If this version supports the part of C++14 needed for the fix lets just change the constant in our build files (can be a separate PR).

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
yalgaer: This looks like a big issue: we mix persistent & view data

All GObjects in the code are data objects. A view is allowed to modify these objects, but in this case this modification is persistent (should not be rolled back on view close) and should be visible to all other views.

Building a new functionality on incorrectly written base will produce even more tech debt than before.

Here is some questions and ideas: 
1) do we store MCA object in the database when it is added to a project and no view is opened?
2) If 1) is yes - does the MCA view clones this object before use?
3) if 2) is yes - the view should create its own in-memory DBI with a cloned object and does not need to 'rollback' anything on close.

yalgaer (commented):
I see only 1 big blocker here: how MCA view works with GObjects: there should be no roll-back on view close and if the object is shared by multiple views the changes made in 1 view should be visible in other views.

Will review minor (aka code-style, etc) stuff after we find a solution for the 'objects' part.

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
DmitriiSukhomlinov: OK, I'll try

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: 1. Chromatogram files are databases itself (they are stored in the "ugenedb" format, which is SQL database), so MultipleChromatogramAlignmentObject data are stored in there permanently. 
2. Sorry, don' get this question. Could you detail it?

yalgaer: OK, the main concern/question is why do you want to do anything when the view is closed?

DmitriiSukhomlinov: **Scenario**:
1. Open ".ugenedb" in Sanger Reads Editor.
2. Open the "Reads" tab, turn "Alternative mutations" on and decrease the "Threshold" value.
    Expected: some characters has been replaced with alternative.
3. Close **the project** (or just remove the ".ugenedb" file from the project).
4. Open the same ".ugenedb" file in Sanger Reads Editor again.
    Expected: **no** alternative mutations are shown, the "Alternative mutations" feature on the "Reads" tab **is disabled**.

`MultipleChromatogramAlignmentObject::sl_viewIsAbout2BeDestroyed` is caught when we remove file from the project. It makes alternative mutations not visible. If we don't have this slot, we will have the following state.
    State: alternative mutations from point "2." are **still visible**, but the "Alternative mutations" feature on the "Reads" tab **is disabled**.

yalgaer: When we remove a document from the project a signal is emitted that the document is removed from the project. 
Why don't you use this signal instead?

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+McaReadsTabFactory::McaReadsTabFactory() {
+    objectViewOfWidget = ObjViewType_ChromAlignmentEditor;
+}
+
+QWidget* McaReadsTabFactory::createWidget(GObjectView* objView, const QVariantMap& options) {
+    Q_UNUSED(options);
```
yalgaer: nit: if you do not use this param just do not give it a name (Q_UNUSED is not needed in this case)

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+    SAFE_POINT(objView != nullptr,
+        QString("Internal error: unable to create widget for group '%1', object view is NULL.").arg(GROUP_ID),
+        nullptr);
+
+    MaEditor* ma = qobject_cast<MaEditor*>(objView);
+    SAFE_POINT(ma != nullptr,
+        QString("Internal error: unable to cast object view to MaEditor for group '%1'.").arg(GROUP_ID),
+        nullptr);
+
+    QWidget* widget = new QWidget(objView->getWidget());
```
yalgaer, line 52: nit: a good place for 'auto': type is already specified in this line

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.h
```diff
@@ -0,0 +1,48 @@
...
+    Q_OBJECT
+public:
+    McaReadsTabFactory();
+
+    QWidget* createWidget(GObjectView* objView, const QVariantMap& options) override;
+
+    OPGroupParameters getOPGroupParameters() override;
+
+private:
+    static const QString GROUP_ID;
```
yalgaer, line 40: nit: keep the private stuff private (in .cpp file) if you can. This header is a public API (U2VIEW_EXPORT)

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: Well, I already told about it - we don't have any agreements about it and I prefer use `Q_UNUSED`. Also there's the attribute `[[maybe_unused]]` from C++17  for these cases, so either we make the agreement or anyone can use the way they like the most

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: What signal exactly? I couldn't find something fit

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
yalgaer: I do not understand why do you prefer extra line with a boilerplate code here. Is there any reason for it?

yalgaer:
I tried the following scenario:

0) open alignment_mca_0014.ugenedb (can be found in tests)

1) activate the first read: SZYD_Cas9_5B70
See this picture:
mca-1

2) Check 'show alternative mutations'
See this picture:
mca-2

3) Uncheck 'show alternative mutations'
The picture stays still. I would expect the picture from the step 1 is shown (the original state)
mca-3




Q: Is it is bug?

yalgaer (changes requested):
I posted a test case with a question

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
yalgaer: > Close the project (or just remove the ".ugenedb" file from the project).

in ProjectModel.h
```
signals:
    void si_documentAdded(Document *d);
    void si_documentRemoved(Document *d);
```

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: Actually, I think that the best salvation for this point is `maybe_unused`, because it's the solution which was specially developed for these cases on the low-level of C++ itself (not some hand-made macro like `Q_UNUSED` or a trick like the way you use). But I'm not sure that we can use it (cuz it's from C++17), so I prefer `Q_UNUSED` because it seems more similar to `maybe_unused`.

DmitriiSukhomlinov:
> I tried the following scenario:
> 
> 1. open alignment_mca_0014.ugenedb (can be found in tests)
> 2. activate the first read: SZYD_Cas9_5B70
>    See this picture:
>    mca-1
> 3. Check 'show alternative mutations'
>    See this picture:
>    mca-2
> 4. Uncheck 'show alternative mutations'
>    The picture stays still. I would expect the picture from the step 1 is shown (the original state)
>    mca-3
> 
> Q: Is it is bug?

Yes, I see. I found the problem - it's because the highest choratogram value is not necessary a peak. Look at the picture - "red" (T) has a peak, but "black" (G) is just a transition between neighborhood characters. Need to improve algorithm.
image

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: It works when you remove a doc from the project, but if you simply close the project it doesn't work

DmitriiSukhomlinov:
OK, I improved the algorithm of peaks detection

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
```diff
@@ -352,14 +352,44 @@ qint64 MultipleChromatogramAlignmentRowData::getBaseCount(qint64 before) const {
...
-                                      { DNAChromatogram::Trace::Trace_G, chromatogram.G[baseCall] },
-                                      { DNAChromatogram::Trace::Trace_T, chromatogram.T[baseCall] } };
+                MultipleChromatogramAlignmentRowData::getTwoHighestPeaks(qint64 position, bool& hasTwoPeaks) const {
+    hasTwoPeaks = true;
+    int previousBaseCall = chromatogram.baseCalls[position != 0 ? position - 1 : position];
+    int baseCall = chromatogram.baseCalls[position];
+    int nextBaseCall = chromatogram.baseCalls[position != (chromatogram.baseCalls.size() - 1) ? position + 1 : position];
+    QList<DNAChromatogram::ChromatogramTraceAndValue> peaks;
+
+    for (auto peak : PEAKS.keys()) {
```
yalgaer: qAsConst

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
```diff
@@ -352,14 +352,44 @@ qint64 MultipleChromatogramAlignmentRowData::getBaseCount(qint64 before) const {
...
+        const auto& chromatogramVector = chromatogram.*PEAKS.value(peak);
```
yalgaer: Use type name instead of "auto" if the type name is not clear (not in the same line or in a several lines above of the current)
As a reviewer (aka code reader) I can't understand what is the type name here.

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
yalgaer: If the var is not used we can omit it: this is a feature of C++ intentionally designed for it and available even in the old standard. Why not using the standard feature and wait for some fancy future standard?

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
yalgaer: I understand about the view, but I quoted your text and answered your question.
The big problem with your patch that you have not separated what is View and what is Data.

If the view changes the state (turns on alternative bases) it has 2 options:

Option 1) Do not change the data: keep the change "view-local"
In this case you do not need the "on-close-callback"

Option 2) The view can change the data (like rename row) but in this case the change to the data is persistent.
In this case you do not need the "on-close-callback" again.

Looks like you implemented some other non-standard option that will cause a lot of issues.

Here is another issue:
1) Open view A
2) Open view B of the same MCA
3) Change the view mode to show alternative mutations in A
4) Switch to B: I see the alternative mutations too, but the options panel (view settings) does not show the correct state.

Check the screenshots (options panel, base at 298 and active view tab):
mca-1

mca-2



I would suggest you to implement either Option-1 or Option-2: make the change view-local or persistent.

yalgaer (changes requested):

yalgaer:
Question: why when I change C to T the signal image is not updated

mca-1
mca-2

I use 80% in the threshold. Does it mean that the second signal is at least 80% from the top one?

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: You wont change my opinion about this. If you insist on using your way, I will do it

src/corelibs/U2Core/src/util/McaDbiUtils.cpp
```diff
@@ -329,6 +329,51 @@ void McaDbiUtils::replaceCharacterInRow(const U2EntityRef &mcaRef, qint64 rowId,
...
+
+        char newChar = newCharList.value(pos);
+        qint64 posInSeq = -1;
+        qint64 endPosInSeq = -1;
+        MaDbiUtils::getStartAndEndSequencePositions(seq, row.gaps, pos, 1, posInSeq, endPosInSeq);
+        if (posInSeq >= 0 && endPosInSeq > posInSeq) {    // not gap
+            U2OpStatus2Log os;
+            DNASequenceUtils::replaceChars(seq, posInSeq, QByteArray(1, newChar), os);
+            SAFE_POINT_OP(os, );
+        } else {
```
evelinabis, line 363: What if the `posInSeq` is less than zero?

src/corelibs/U2Core/src/util/McaDbiUtils.cpp
```diff
@@ -329,6 +329,51 @@ void McaDbiUtils::replaceCharacterInRow(const U2EntityRef &mcaRef, qint64 rowId,
...
+            U2OpStatus2Log os;
+            DNAChromatogram chrom = ChromatogramUtils::exportChromatogram(os, U2EntityRef(mcaRef.dbiRef, row.chromatogramId));
+            ChromatogramUtils::insertBase(chrom, posInSeq, row.gaps, pos);
+            ChromatogramUtils::updateChromatogramData(os, mcaRef.entityId, U2EntityRef(mcaRef.dbiRef, row.chromatogramId), chrom);
```
evelinabis, line 367: Connections are created here, if I'm not mistaken.

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.h
```diff
@@ -22,6 +22,7 @@
 #ifndef _U2_MULTIPLE_CHROMATOGRAM_ALIGNMENT_OBJECT_H_
 #define _U2_MULTIPLE_CHROMATOGRAM_ALIGNMENT_OBJECT_H_
 
+#include <U2View/McaReadsTabSettings.h>
```
evelinabis: Can be forward declared.

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
```diff
@@ -0,0 +1,66 @@
...
+private slots:
+    void sl_mutationsGroupBoxToggeled(bool on);
+    void sl_mutationsThresholdValueChanged(int newValue);
+    void sl_updateAlternativeMutations();
+
+private:
+    void updateAlternativeMutations();
+
+    McaEditorSequenceArea* seqArea = nullptr;
+    MultipleChromatogramAlignmentObject* mcaObject = nullptr;
```
evelinabis, line 81: Is it possible to replace it with smart pointers?

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
```diff
@@ -0,0 +1,66 @@
...
+    McaAlternativeMutationsWidget(QWidget* parent = nullptr);
+
+    /*
+     * Initialize object.
+     * Call this function right after instance creation.
+     */
+    void init(MultipleAlignmentObject* maObject, MaEditorSequenceArea* seqArea);
+
+private slots:
+    void sl_mutationsGroupBoxToggeled(bool on);
```
evelinabis: Typo.

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,71 @@
...
+    QVBoxLayout* layout = new QVBoxLayout();
+    layout->setContentsMargins(0, 0, 0, 0);
+    widget->setLayout(layout);
+
+    McaAlternativeMutationsWidget* alternativeMutationsWgt = new McaAlternativeMutationsWidget(widget);
+    alternativeMutationsWgt->init(ma->getMaObject(), ma->getUI()->getSequenceArea());
```
evelinabis: Check that `ma->getUI()` isn't nullptr, if you want.

src/include/U2View/MaEditorSequenceArea.h
```diff
@@ -0,0 +1 @@
+#include "../../corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.h"
```
evelinabis: Where is this file used?

src/include/U2View/McaReadsTabSetting.h
```diff
@@ -0,0 +1 @@
+#include "../../corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabSetting.h"
```
evelinabis: Where is this file used?

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
```diff
@@ -188,6 +180,23 @@ void GTUtilsOptionPanelMca::pushExportButton(HI::GUITestOpStatus &os) {
     QToolButton *result = GTWidget::findExactWidget<QToolButton *>(os, "exportBtn");
     result->click();
 }
+void GTUtilsOptionPanelMca::showAlternativeMutations(HI::GUITestOpStatus& os, bool show, int value, bool withSpinbox) {
```
evelinabis: Does it need to prescribe `#define GT_METHOD_NAME`?

evelinabis: Please, add empty line.

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
```diff
@@ -188,6 +180,23 @@ void GTUtilsOptionPanelMca::pushExportButton(HI::GUITestOpStatus &os) {
...
+    if (withSpinbox) {
+        GTSpinBox::setValue(os, "mutationsThresholdSpinBox", value, nullptr);
+    } else {
+        GTSlider::setValue(os, GTWidget::findExactWidget<QSlider*>(os, "mutationsThresholdSlider"), value);
+    }
+
+    GTWidget::click(os, GTWidget::findExactWidget<QPushButton*>(os, "updateMutationsPushButton"));
+    GTGlobals::sleep(1000);
+}
 #undef GT_METHOD_NAME
```
evelinabis, line 204: Please move this line to line 183.

evelinabis (commented):
Good job!
You ignore the reference alignment style (Clang-Format).

evelinabis:
@yalgaer, I can't repeat this scenario. Please describe it in more detail.

> I tried the following scenario:
> 
> 1. open alignment_mca_0014.ugenedb (can be found in tests)
> 2. activate the first read: SZYD_Cas9_5B70
>    See this picture:
>    mca-1
> 3. Check 'show alternative mutations'
>    See this picture:
>    mca-2
> 4. Uncheck 'show alternative mutations'
>    The picture stays still. I would expect the picture from the step 1 is shown (the original state)
>    mca-3
> 
> Q: Is it is bug?

evelinabis:
**Scenario**

1. Open any sanger file.
2. Open Reads tab, check "Show alternative mutations:".
3. Remove sanger doc from project.
4. Open it again. Check that button "Undo" enable.

Is that how it should be?

yalgaer:
> @yalgaer, I can't repeat this scenario. Please describe it in more detail.

You need to check out this branch with the date of the comment to reproduce it. I think it was already fixed today.

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/datatype/msa/MultipleChromatogramAlignmentRow.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
> Question: why when I change C to T the signal image is not updated
> 
> mca-1
> mca-2
> 
> I use 80% in the threshold. Does it mean that the second signal is at least 80% from the top one?

Sorry, don't get what you mean. For this character main peak (blue) corresponds to C, the second peak (red) corresponds to C. The alternative mutation is enabled starting with 95% and lower. Could you, pleas, describe your idea more clearly?

src/corelibs/U2Core/src/util/McaDbiUtils.cpp
DmitriiSukhomlinov: SAFE_POINT has been added

src/corelibs/U2Core/src/util/McaDbiUtils.cpp
DmitriiSukhomlinov: Don't get it

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.h
DmitriiSukhomlinov: ?

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
DmitriiSukhomlinov: Possible, but I don't see a need of this

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: OK

src/include/U2View/MaEditorSequenceArea.h
DmitriiSukhomlinov: Nowhere, removed

src/include/U2View/McaReadsTabSetting.h
DmitriiSukhomlinov: Nowhere, removed

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
DmitriiSukhomlinov: Don't get the idea

DmitriiSukhomlinov:
> **Scenario**
> 
> 1. Open any sanger file.
> 2. Open Reads tab, check "Show alternative mutations:".
> 3. Remove sanger doc from project.
> 4. Open it again. Check that button "Undo" enable.
> 
> Is that how it should be?

Difficult to explain, in the current realization - yes.

yalgaer (changes requested):
https://github.com/ugeneunipro/ugene/pull/364#discussion_r630706275
what about this problem?

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
```diff
@@ -0,0 +1,169 @@
...
+//
+//    bool setUpFromDb = checkedObjectAttributes.size() == 1;
+//    if (setUpFromDb) {
+//        checkedStateAttribute.id = checkedObjectAttributes.first();
+//    }
+//    mcaDbiObj.dbiId = mcaObject->getEntityRef().dbiRef.dbiId;
+//    mcaDbiObj.id = mcaObject->getEntityRef().entityId;
+//    mcaDbiObj.version = mcaObject->getModificationVersion();
+//    U2AttributeUtils::init(checkedStateAttribute, mcaDbiObj, attributeName);
+//}
```
evelinabis: Not necessary.

src/corelibs/U2View/src/ov_msa/McaEditor.h
```diff
@@ -76,6 +80,10 @@ class U2VIEW_EXPORT McaEditor : public MaEditor {
         return gotoSelectedReadAction;
     }
 
+signals:
+    /* Emit this signal to notify that McaEditor is about to be destroyed. */
+    void si_isAbout2BeDestroyed();
```
evelinabis: Need to?

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
```diff
@@ -0,0 +1,87 @@
...
+class MultipleChromatogramAlignmentObject;
+
+
+/*
+ * Widget to set up alternative mutations. Relates to UGENE-7105.
+ * When the "Alternative mutations" feature is enabled,
+ * all symbols with the second chromatogram trace percentage more than the defined "threshold" value
+ * has to be replaced with the symbol, which corresponds to the second chromatogram trace.
+ */
+class McaAlternativeMutationsWidget : public QWidget, public Ui_McaAlternativeMutationsWidget {
```
evelinabis: `public` or `private` `Ui_McaAlternativeMutationsWidget`?

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
```diff
@@ -0,0 +1,169 @@
...
+    mcaObject->updateAlternativeMutations(mutationsGroupBox->isChecked(), mutationsThresholdSlider->value(), os);
+    CHECK_OP(os, );
+
+    updateDb(os);
+    CHECK_OP(os, );
+}
+
+void McaAlternativeMutationsWidget::sl_updateDb() {
+    U2OpStatus2Log os;
+    updateDb(os);
```
evelinabis: Why is there a `U2UseCommonUserModStep` variable in the `sl_updateAlternativeMutations`, but not here?

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
```diff
@@ -0,0 +1,169 @@
...
+    SAFE_POINT(mcaObject != nullptr, "MaConsensusModeWidget can not be initialized: MultipleChromatogramAlignmentObject is nullptr", );
+
+    mutationsGroupBox->setChecked(false);
+    mutationsThresholdSlider->setValue(99);
+
+    updateValuesFromDb();
+
+    connect(mutationsGroupBox, SIGNAL(toggled(bool)), this, SLOT(sl_updateAlternativeMutations()));
+    connect(mutationsThresholdSlider, SIGNAL(valueChanged(int)), this, SLOT(sl_updateDb()));
+    connect(mutationsThresholdSpinBox, SIGNAL(valueChanged(int)), this, SLOT(sl_updateDb()));
```
evelinabis: Why are changes written to the database, but not displayed to the user?

evelinabis (commented):
If you press Undo, the alignment (view) changes, but the settings in the widget remain the same -- a mismatch occurs. Reopening the Reads tab doesn't help.

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
DmitriiSukhomlinov: Ok

src/corelibs/U2View/src/ov_msa/McaEditor.h
DmitriiSukhomlinov: No, removed

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
DmitriiSukhomlinov: It's OK to have it `public`, but I made it `private` as far as you noticed it

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
DmitriiSukhomlinov: Don't get the idea

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
DmitriiSukhomlinov: Because it won't work for this change. The "undo/redo" framework works with `U2ObjectDbi`, but changes about "Alternative mutations" settings are written to `U2AttributeDbi`. I don't think that this behavior should be changed

DmitriiSukhomlinov:
> If you press Undo, the alignment (view) changes, but the settings in the widget remain the same -- a mismatch occurs. Reopening the Reads tab doesn't help.

See https://github.com/ugeneunipro/ugene/pull/364#discussion_r640402355

DmitriiSukhomlinov:
> [#364 (comment)](https://github.com/ugeneunipro/ugene/pull/364#discussion_r630706275)
> what about this problem?

"Alternative mutations" settings now are saved to database. If you want something different, please, describe it in details

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
DmitriiSukhomlinov: I removed this lines, cuz I think, that the correct behaviour is to save date to database on "Update" press only

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
evelinabis: When I click Update, the Undo button enables. Does this mean that changes have been made to Object (not Attribute)?

yalgaer (changes requested):
It is good that we do not 'rollback' state during the view closing. Now logic will be more predictable.

Here is 2 problems I found:

**1)** 
I tried the following scenario in debug mode and got this crash:
`
/McaAlternativeMutationsWidget.cpp:87: U2::McaAlternativeMutationsWidget::updateValuesFromDb()::<lambda(U2::U2IntegerAttribute&, const QString&)>: Assertion `objectAttributes.size() == 0 || objectAttributes.size() == 1' failed.
`

Scenario:
1) Open 2 views with MCA file from alignment_alternative_mutations.ugenedb
2) Set 90% in the first view and 80% in the second view
3) Switch back to the first view and uncheck
4) Close the first view and the second view
5) Open view again
6) Open options panel tab: UGENE crashes.

I would like to ask to add a GUI test like this, because this feature affects cross-view interaction.

**2)** 
We need some kind of state indicator for MCA view and options panel is not the best choice here.
I would suggest adding a text into the status bar like "Mode: normal/alternative" as a minimal solution.

smlnkcv (dismissed):

yalgaer:
I tested the latest changes, it works better, but still have some issues. Example:
1) Open the file from GUI tests.
2) Set threshold to 8%.
3) See that the second read at position 308 has 'A' now (the original base was 'G').
4) Close the view (no need to close UGENE).
5) Open the same view again: alternative mutations are ON, threshold at 8% -> this is correct.
6) Check the second read at position 308: you will see G -> looks like mutations are not shown anymore.

open-1
open-2

Could you please add a test for this bug too?

yalgaer (changes requested):
I described one more scenario that should be fixed.

DmitriiSukhomlinov:
Fixed

DmitriiSukhomlinov:
Fixed!

yalgaer (changes requested):
Here is another scenario that works in version 39, but crashes UGENE with this patch
1) Make the file from the tests read-only
2) Open the file:
UGENE-39: the file is opened. In the log we see a warning that the file is read-only
This patch: UGENE crashes.

Please add a UI test for this problem too.

DmitriiSukhomlinov:
> Here is another scenario that works in version 39, but crashes UGENE with this patch
> 
> 1. Make the file from the tests read-only
> 2. Open the file:
>    UGENE-39: the file is opened. In the log we see a warning that the file is read-only
>    This patch: UGENE crashes.
> 
> Please add a UI test for this problem too.

I checked this scenario - it's been happening before this changes. I created the issue [UGENE-7312](https://ugene.dev/tracker/browse/UGENE-7312) about it, but it won't be fixed in this very issue.

DmitriiSukhomlinov:
@yalgaer please, approve this feature, I think, it's enough for it.

Tests: [Windows](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsSummaryWindows1032bit/218482), [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/bt153/221803), [macOS](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsMacOsX10125Qt5macX64summary/221809)

yalgaer:
> > Here is another scenario that works in version 39, but crashes UGENE with this patch
> > 
> > 1. Make the file from the tests read-only
> > 2. Open the file:
> >    UGENE-39: the file is opened. In the log we see a warning that the file is read-only
> >    This patch: UGENE crashes.
> > 
> > Please add a UI test for this problem too.
> 
> I checked this scenario - it's been happening before this changes. I created the issue [UGENE-7312](https://ugene.dev/tracker/browse/UGENE-7312) about it, but it won't be fixed in this very issue.

I added 2 videos: 1 with the patch crashes and 1 with no patch opens normally.

Please check below:
https://user-images.githubusercontent.com/17791039/122934772-d41bf580-d33d-11eb-9444-853718df48d0.mp4
https://user-images.githubusercontent.com/17791039/122934775-d41bf580-d33d-11eb-995a-fa8924d9a5e4.mp4

yalgaer:
> @yalgaer please, approve this feature, I think, it's enough for it.
> 
> Tests: [Windows](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsSummaryWindows1032bit/218482), [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/bt153/221803), [macOS](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsMacOsX10125Qt5macX64summary/221809)

I agree you did a lot of work and I did a lot of work too reviewing and testing this patch.
The problem is that the current state of the feature is not good enough to be included in the release (master).

Question: will you say yes to releasing this feature in the current state? 

Here is the final list of improvements that need to be done to have this PR approved:
1) Ensure that newly added controls process read-only state correctly: controls are disabled.
2) Replace the 'TODO' icon with the correct one. The 'correct' one may be borrowed from the MSA or other view or you can create your own using the same palette with existing ones. 
I would recommend the icon "Show chromatograms" (":/core/images/graphs.png").

yalgaer:
> @yalgaer please, approve this feature, I think, it's enough for it.
> 
> Tests: [Windows](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsSummaryWindows1032bit/218482), [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/bt153/221803), [macOS](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsMacOsX10125Qt5macX64summary/221809)



> > Here is another scenario that works in version 39, but crashes UGENE with this patch
> > 
> > 1. Make the file from the tests read-only
> > 2. Open the file:
> >    UGENE-39: the file is opened. In the log we see a warning that the file is read-only
> >    This patch: UGENE crashes.
> > 
> > Please add a UI test for this problem too.
> 
> I checked this scenario - it's been happening before this changes. I created the issue [UGENE-7312](https://ugene.dev/tracker/browse/UGENE-7312) about it, but it won't be fixed in this very issue.

You are right, I checked new debug vs old release version.

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
```diff
@@ -190,6 +182,26 @@ void GTUtilsOptionPanelMca::pushExportButton(HI::GUITestOpStatus &os) {
 }
 #undef GT_METHOD_NAME
 
+#define GT_METHOD_NAME "showAlternativeMutations"
+void GTUtilsOptionPanelMca::showAlternativeMutations(HI::GUITestOpStatus &os, bool show, int value, bool withSpinbox, QWidget *parent) {
+    GTUtilsOptionPanelMca::openTab(os, Tabs::Reads, parent);
+    GTGroupBox::setChecked(os, "mutationsGroupBox", show, parent);
+    if (!show) {
+        GTGlobals::sleep(1000);
```
yalgaer: Do not use sleeps in tests (unless the case is out of QT logic scope)

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
```diff
@@ -190,6 +182,26 @@ void GTUtilsOptionPanelMca::pushExportButton(HI::GUITestOpStatus &os) {
...
+    }
+
+    if (withSpinbox) {
+        GTSpinBox::setValue(os, "mutationsThresholdSpinBox", value, parent);
+    } else {
+        GTSlider::setValue(os, GTWidget::findExactWidget<QSlider *>(os, "mutationsThresholdSlider", parent), value);
+    }
+
+    GTWidget::click(os, GTWidget::findExactWidget<QPushButton *>(os, "updateMutationsPushButton", parent));
+    GTGlobals::sleep(1000);
```
yalgaer: Same here: wait for the real action to finish, do not use sleeps

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
```diff
@@ -0,0 +1,89 @@
...
+private slots:
+    /*
+     * Update the sequence area and write new alternative mutations settings to the database
+     **/
+    void sl_updateAlternativeMutations();
+
+private:
+    void showEvent(QShowEvent* e) override;
+    /*
+     * Update GUI with values from the database
```
yalgaer: nit: use "Updates", not "Update"

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+void McaAlternativeMutationsWidget::sl_updateAlternativeMutations() {
+    U2OpStatus2Log os;
+    U2UseCommonUserModStep userModStep(mcaObject->getEntityRef(), os);
+    Q_UNUSED(userModStep);
+
+    mcaObject->updateAlternativeMutations(mutationsGroupBox->isChecked(), mutationsThresholdSlider->value(), os);
+    CHECK_OP(os, );
+
+    updateDb(os);
+    CHECK_OP(os, );
```
yalgaer, line 80: not: this check is not needed

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
```diff
@@ -260,6 +260,44 @@ void MultipleChromatogramAlignmentObject::trimRow(const int rowIndex, int curren
...
+        for (int j = 0; j < ungappedLength; j++) {
+            bool ok = false;
+            auto res = mcaRow->getTwoHighestPeaks(j, ok);
+            if (!ok) {
+                continue;
+            }
+
+            double minimumThresholdValue = (double)res.second.value / res.first.value * 100;
+            char newChar = 'A';
+            if (minimumThresholdValue < threshold || !showAlternativeMutations) {
```
yalgaer, line 278: Unused value. You can keep newChar uninitialized or use ? operator instead of the if/else

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,72 @@
...
+    SAFE_POINT(ma != nullptr,
+        QString("Internal error: unable to cast object view to MaEditor for group '%1'.").arg(GROUP_ID),
+        nullptr);
+
+    QWidget* widget = new QWidget(objView->getWidget());
+    QVBoxLayout* layout = new QVBoxLayout();
+    layout->setContentsMargins(0, 0, 0, 0);
+    widget->setLayout(layout);
+
+    McaAlternativeMutationsWidget* alternativeMutationsWgt = new McaAlternativeMutationsWidget(widget);
```
yalgaer: nit: good place to use auto and do not use type twice in the line

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto ui = ma->getUI();
+    SAFE_POINT(ui != nullptr, "UI isn't found", nullptr);
+
+    alternativeMutationsWgt->init(ma->getMaObject(), ui->getSequenceArea(), ui->getStatusBar());
+    ShowHideSubgroupWidget* alternativeMutations = new ShowHideSubgroupWidget("ALTERNATIVE_MUTATIONS_MODE", tr("Alternative mutations"), alternativeMutationsWgt, true);
```
yalgaer: nit: same here: auto

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorWgt.h
```diff
@@ -87,6 +87,10 @@ class U2VIEW_EXPORT MaEditorWgt : public QWidget {
         return offsetsViewController;
     }
 
+    MaEditorStatusBar* getStatusBar() const {
```
yalgaer: Move body of the method into cpp.

yalgaer (changes requested):

DmitriiSukhomlinov:
> > @yalgaer please, approve this feature, I think, it's enough for it.
> > Tests: [Windows](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsSummaryWindows1032bit/218482), [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/bt153/221803), [macOS](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/UgeneNightlyBuilds_GuiTests_GuiTestsMacOsX10125Qt5macX64summary/221809)
> 
> I agree you did a lot of work and I did a lot of work too reviewing and testing this patch.
> The problem is that the current state of the feature is not good enough to be included in the release (master).
> 
> Question: will you say yes to releasing this feature in the current state?
> 
> Here is the final list of improvements that need to be done to have this PR approved:
> 
> 1. Ensure that newly added controls process read-only state correctly: controls are disabled.
> 2. Replace the 'TODO' icon with the correct one. The 'correct' one may be borrowed from the MSA or other view or you can create your own using the same palette with existing ones.
>    I would recommend the icon "Show chromatograms" (":/core/images/graphs.png").

1. There was a problem and it's fixed now.
2. OK, your suggestion'h been used

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_msa/mca_reads/McaAlternativeMutationsWidget.cpp
DmitriiSukhomlinov: Why? `updateDb` has `os` as argument and it's possible that it (`os`) may be changed.

src/corelibs/U2Core/src/gobjects/MultipleChromatogramAlignmentObject.cpp
DmitriiSukhomlinov: OK, Uninitialized.

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_msa/mca_reads/McaReadsTabFactory.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorWgt.h
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/GTUtilsOptionPanelMca.cpp
DmitriiSukhomlinov: OK

yalgaer (approved):

PR #365, opened 2021-05-01
UGENE-7212. Crash if Pairwise re-align.

evelinabis:
[Issue](https://ugene.dev/tracker/browse/UGENE-7212)


src/corelibs/U2View/src/ov_msa/MSAEditor.h
```diff
@@ -48,7 +48,7 @@ class PairwiseAlignmentWidgetsSettings {
     QString algorithmName;
     bool inNewWindow;
     QString resultFileName;
-    PairwiseAlignmentTask *pairwiseAlignmentTask;
```
yalgaer, line 51: I see that 'pairwiseAlignmentTask' sometimes is deleted by 3rd party code and you want to be protected using QPointer and this is OK.
Question: which code deleted the task?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -321,6 +324,31 @@ GUI_TEST_CLASS_DEFINITION(test_7152) {
     CHECK_SET_ERR(bottomRight == "11/40/35", "Bottom right position is wrong: " + bottomRight);
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7212) {
+    // Open _common_data/clustal/shortened_big.aln.
+    // Click the Pairwise Alignment tab of the Options Panel.
+    // Select two sequence from the original alignment and click on the Align button.
+    // Until the task completes, click the Pairwise Alignment tab again.
+    // Wait for task finish.
+    // Return to big.aln and click the Pairwise Alignment tab of the Options Panel.
```
yalgaer: Is the 'big.aln' the same as 'shortened_big.aln' from the description before?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -321,6 +324,31 @@ GUI_TEST_CLASS_DEFINITION(test_7152) {
...
+    GTFileDialog::openFile(os, testDir + "_common_data/clustal/shortened_big.aln");
+    GTUtilsMsaEditor::checkMsaEditorWindowIsActive(os);
+    GTUtilsOptionPanelMsa::toggleTab(os, GTUtilsOptionPanelMsa::PairwiseAlignment);
+    GTUtilsOptionPanelMsa::addFirstSeqToPA(os, "seq1");
+    GTUtilsOptionPanelMsa::addSecondSeqToPA(os, "seq2");
+    GTWidget::click(os, GTUtilsOptionPanelMsa::getAlignButton(os));
+    GTUtilsOptionPanelMsa::toggleTab(os, GTUtilsOptionPanelMsa::PairwiseAlignment);
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    GTUtilsDocument::removeDocument(os, "PairwiseAlignmentResult.aln");
```
yalgaer, line 347: Does this test fail if run with no fix?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: I will adjust the text of the test scenario to match the test

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: This test fails if run with no fix, because UGENE will crash on Windows.

src/corelibs/U2View/src/ov_msa/MSAEditor.h
evelinabis: https://github.com/ugeneunipro/ugene/blob/83018b0ad6624087bf8b1251e19b6ba3b52a8a79/src/corelibs/U2Private/src/TaskSchedulerImpl.cpp#L927

DmitriiSukhomlinov (approved):

src/corelibs/U2View/src/ov_msa/MSAEditor.h
```diff
@@ -39,7 +39,7 @@ class PairwiseAlignmentWidgetsSettings {
     PairwiseAlignmentWidgetsSettings()
         : firstSequenceId(U2MsaRow::INVALID_ROW_ID),
           secondSequenceId(U2MsaRow::INVALID_ROW_ID), inNewWindow(true),
-          pairwiseAlignmentTask(NULL), showSequenceWidget(true), showAlgorithmWidget(false),
+          pairwiseAlignmentTask(), showSequenceWidget(true), showAlgorithmWidget(false),
```
DmitriiSukhomlinov: You don't need `pairwiseAlignmentTask` here at all, cuz QPointer has the default constructor

DmitriiSukhomlinov (changes requested):
See comment

src/corelibs/U2View/src/ov_msa/MSAEditor.h
yalgaer: This is true for all tasks in UGENE: all tasks are deleted by the task scheduler. 
But any code that holds a reference to a task usually is required to listen for the task state changes like run/report/finished, etc. and update the pointer to the task: null it when task is finished.

I see that this kind of the code is written here: it listens to state changes:
```connect(task, SIGNAL(si_stateChanged()), SLOT(sl_alignComplete()));) but does it incorrectly.```
It nulls the pointer when task is finished (before it is deleted).

This way your fix may hide the original problem and leave some invalid code unfixed just because it does not crash anymore. I'm OK with QPointer (as I said) just want to understand the original problem.

Question: what would be the fix with no QPointer. Why we are keeping a pointer to a deleted task and do not null it in in sl_alignComplete?

evelinabis: Because if we close the "Pairwise Alignment" tab, the object of the `PairAlign` class is deleted, which should have null the current `pairwiseAlignmentTask` in `pairwiseAlignmentWidgetsSettings`. These settings are a shared object (via a simple pointer) between `PairAlign` and `MSAEditor`: https://github.com/ugeneunipro/ugene/blob/0addaa0f72c9f484244b3ddf1c258bc38d1e278a/src/corelibs/U2View/src/ov_msa/PairAlign/PairAlign.cpp#L71 When we open the "Pairwise Alignment" tab again, we create a new `PairAlign` object, the `pairwiseAlignmentWidgetsSettings` of which are taken from `MSAEditor` -- the outdated settings with a nonnull pointer: https://github.com/ugeneunipro/ugene/blob/0addaa0f72c9f484244b3ddf1c258bc38d1e278a/src/corelibs/U2Gui/src/options_panel/OptionsPanel.cpp#L136

Could there be any other problems here, do you think?

evelinabis:
@yalgaer, please review me. My [GUI](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/viewLog.html?buildId=217664&buildTypeId=bt153) tests passed.

src/corelibs/U2View/src/ov_msa/MSAEditor.h
yalgaer: This sounds correct. Thank you for the explanation.

yalgaer (approved):

DmitriiSukhomlinov (approved):

DmitriiSukhomlinov (changes requested):
Need to check code

DmitriiSukhomlinov (approved):

smlnkcv (approved):

PR #367, opened 2021-05-03
UGENE-7222. Migrate FASTA format to the non-deprecated DocumentTextFo…

yalgaer:
…rmat base class


src/corelibs/U2Core/src/util/TextUtils.cpp
```diff
@@ -239,4 +239,14 @@ bool TextUtils::isWhiteSpace(const QString &text, int charIndex) {
     return WHITES.testBit(bitIndex);
 }
 
+QString TextUtils::skip(const QBitArray &map, const QString &text) {
```
DmitriiSukhomlinov, line 242: `map` isn't `QMap`, so, I think, it's better to rename this variable

src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -70,180 +54,184 @@ FastaFormat::FastaFormat(QObject *p)
...
 
-FormatCheckResult FastaFormat::checkRawTextData(const QByteArray &rawData, const GUrl &) const {
-    const char *data = rawData.constData();
-    int size = rawData.size();
-
-    int n = TextUtils::skip(TextUtils::WHITES, data, size);
-    int newSize = size - n;
-    const char *newData = data + n;
-    if (newSize <= 0 || (newData[0] != FASTA_HEADER_START_SYMBOL && newData[0] != FASTA_COMMENT_START_SYMBOL)) {
+FormatCheckResult FastaFormat::checkRawTextData(const QString &dataPrefix, const GUrl &) const {
```
DmitriiSukhomlinov, line 109: You rewrote this function and several checks are gone now (e.g `newSize <= 0` or `TextUtils::contains(TextUtils::BINARY, data, size)`). Maybe it's better to get them back?

src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -70,180 +54,184 @@ FastaFormat::FastaFormat(QObject *p)
...
-    if (hasBinaryBlocks) {
-        return FormatDetection_NotMatched;
+    // Ok, format is matched -> add hints about sequence sizes.
+    FormatCheckResult result(score);
+    result.properties = buildFormatDetectionHints(data);
+    return result;
+}
+
+/** Returns true if the line is a valid comment line for FASTA format. */
+static bool isCommentLine(const QString &line) {
```
DmitriiSukhomlinov: I think, it'll be better to define all these `static` functions in the .h file

src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -253,40 +241,40 @@ static void load(IOAdapter *io, const U2DbiRef &dbiRef, const QVariantMap &fs, Q
...
                 break;
             }
             memoryLocker.tryAcquire(800);
             CHECK_OP_BREAK(os);
             U2Sequence seq = seqImporter.finalizeSequenceAndValidate(os);
-            if (os.hasError() && os.getError() == U2SequenceImporter::EMPTY_SEQUENCE_ERROR) {
+            if (os.getError() == U2SequenceImporter::EMPTY_SEQUENCE_ERROR) {
                 os.setError("");
-                emptySeqNames << headerLine;
+                emptySeqNames << headerLine.mid(1);
```
DmitriiSukhomlinov: Add a comment, that `mid` removes `>` from name, it's not pretty obvious

src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -298,192 +286,173 @@ static void load(IOAdapter *io, const U2DbiRef &dbiRef, const QVariantMap &fs, Q
...
-    if (io->writeBlock(block, blockSize) != blockSize || !io->writeBlock("\n", 1)) {
-        os.setError(L10N::errorWritingFile(io->getURL()));
-    }
+static void writeHeaderToFile(IOAdapterWriter &writer, const QString &sequenceName, U2OpStatus &os) {
+    writer.write(os, FastaFormat::FASTA_HEADER_START_SYMBOL + sequenceName + '\n');
 }
 
-#define SAVE_LINE_LEN 1024 * 1024
+// TODO: a recommended width for FASTA lines is 80 chars to be readable in text editors.
+#define SAVE_LINE_LEN qint64(1024 * 1024)
```
DmitriiSukhomlinov: Add a `static constexpr` variable instead of the preprocessor command

src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -298,192 +286,173 @@ static void load(IOAdapter *io, const U2DbiRef &dbiRef, const QVariantMap &fs, Q
...
-void FastaFormat::storeDocument(Document *doc, IOAdapter *io, U2OpStatus &os) {
-    //TODO: check saved op states!!!
-    foreach (GObject *o, doc->getObjects()) {
-        U2SequenceObject *seqObj = dynamic_cast<U2SequenceObject *>(o);
-        if (NULL != seqObj) {
-            saveSequence(io, seqObj, os);
+void FastaFormat::storeTextDocument(IOAdapterWriter &writer, Document *document, U2OpStatus &os) {
+    QList<GObject *> objects = document->getObjects();
+    for (GObject *object : qAsConst(objects)) {
+        if (auto sequenceObject = dynamic_cast<U2SequenceObject *>(object)) {
```
DmitriiSukhomlinov, line 354: Initializes in `if` statement are available since C++17. As far as I know we support C++11 only, aren't we?

src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -298,192 +286,173 @@ static void load(IOAdapter *io, const U2DbiRef &dbiRef, const QVariantMap &fs, Q
...
-        DNASequence *seq = new DNASequence(headerLine, sequence);
-        seq->alphabet = AppContext::getDNAAlphabetRegistry()->findById(BaseDNAAlphabetIds::NUCL_DNA_EXTENDED());
-        assert(seq->alphabet != NULL);
-
-        if (!seq->alphabet->isCaseSensitive()) {
-            TextUtils::translate(TextUtils::UPPER_CASE_MAP, const_cast<char *>(seq->seq.constData()), seq->seq.length());
+        QString sequenceName = headerLine.mid(1);
+        auto dnaSequence = new DNASequence(sequenceName, sequence);
+        dnaSequence->alphabet = AppContext::getDNAAlphabetRegistry()->findById(BaseDNAAlphabetIds::NUCL_DNA_EXTENDED());
+        SAFE_POINT(dnaSequence->alphabet != nullptr, "Sequence alphabet is null!", nullptr);
```
DmitriiSukhomlinov: If you fail here, `dnaSequence` memory leaks. Use `QSharedProinter`

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Core/src/util/TextUtils.cpp
yalgaer: map here is for 'bit-map'.
I agree that we can find a better name for it, but 10+ other methods in this class use 'map' for bit-maps.

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: TextDocumentFormat methods should never be called for binary data now. The check is done in the base class before any child class method is called.

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: Headers is a part of public API. Local methods should stay local and do not spoil public interface or generate extra dependencies.

To make it possible there are different patterns in C++, like PImpl (used across QT code) or local functions.
So local static functions is a good thing that keeps local stuff local - I think we should use them all the time where it is possible.

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: Done

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: We already have a bunch of places with C++17 : initializers in ifs(), shortened collections {a, b, c}, etc.

The answer is: if all our CI servers can compile the code -  the code is valid.

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: Reworked the code to avoid repeated .mid(1) calls

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: Good point. I reworked it to move the check before new()

DmitriiSukhomlinov (approved):

DmitriiSukhomlinov:
Approved, I hope, you check all tests before merge

DmitriiSukhomlinov (dismissed):

smlnkcv (dismissed):

PR #368, opened 2021-05-03
UGENE-7189. Rollback glibc version support for Linux

yalgaer:
I experimented with this technique during the last 2 weeks and found it very unstable. 

The solution works, but any random commit can introduce a new path/system function call that have a chance to be mapped to some new GLIBC function that does not exist in the old target version.

As the result I'm rolling it back and the best and the simplest way for us to target both new & old platforms on Linux is to build on the old platfrom (Debian 9/Ubuntu 16.04 today)


DmitriiSukhomlinov:
I think, the best way to rollback this changes is to make `git revert`. Click the "Revert" button on this PR https://github.com/ugeneunipro/ugene/pull/342 and the corresponding PR will be created

yalgaer:
> I think, the best way to rollback this changes is to make `git revert`. Click the "Revert" button on this PR #342 and the corresponding PR will be created

Done. Closing this PR.

PR #372, opened 2021-05-05
UGENE-7217 SAFE_POINT "Object relation is not valid!" in "In Silico PCR" sample

rasputinkirill:
Most of the XML tests from 'cmdline/pcr' failed because of this issue.
So, we do not need additional tests.

### DIFF
--- a/src/plugins/pcr/src/ExtractProductTask.cpp
+++ b/src/plugins/pcr/src/ExtractProductTask.cpp
@@ -23,6 +23,7 @@
 
 #include <QCoreApplication>
 #include <QDir>
+#include <QTemporaryFile>
 
 #include <U2Core/AnnotationTableObject.h>
 #include <U2Core/AppContext.h>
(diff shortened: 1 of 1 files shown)


src/plugins/pcr/src/ExtractProductTask.cpp
```diff
@@ -240,7 +241,11 @@ void ExtractProductTask::run() {
     AnnotationTableObject *annotations = new AnnotationTableObject(productSequence.getName() + " features", dbiRef);
     annotations->addAnnotations(QList<SharedAnnotationData>() << getPrimerAnnotation(product.forwardPrimerMatchLength, U2Strand::Direct, productSequence.length()));
     annotations->addAnnotations(QList<SharedAnnotationData>() << getPrimerAnnotation(product.reversePrimerMatchLength, U2Strand::Complementary, productSequence.length()));
-    annotations->addObjectRelation(GObjectRelation(GObjectReference(sequenceObject), ObjectRole_Sequence));
+    GObjectReference ref(sequenceObject);
```
yalgaer, line 244: Why adding a random non-existing URL  here helps to solve the problem?

rasputinkirill: Impossible to add object relation with empty url in GObjectReference. 
So we add temporary url for object to pass this.
As i understand this situation is only possible when this task started through WD.

DmitriiSukhomlinov: Actually, it looks like a crutch, but I don't have a better offer

DmitriiSukhomlinov (approved):

smlnkcv (approved):

src/plugins/pcr/src/ExtractProductTask.cpp
yalgaer: A long story short: hacks like this won't go.

If GObjectReference requires an URL the URL must be real. Why can't we re-use the URL of the document for the sequence object?

yalgaer (changes requested):

rasputinkirill:
This document has an empty URL.
Because this is workflow and its URL will be defined in "Write Sequence" worker.

PR #374, opened 2021-05-05
UGENE-6806 Opening of the big sequence causes crash

rasputinkirill:
I didn't find any other index to replace "rtree_i32" index, so i just remove it.
Yes, we lost possible performance with massive operations with annotations. 
But now we do not need to maintain this index at runtime.

File in test - the small one with annotation which location is out of range for 32bit integer.


yalgaer (changes requested):
When I use your test file I see this error in U2OpStatus:
Unexpected query result code: 19 (rtree constraint failed: FeatureLocationRTreeIndex.(start<=end))

This error is not handled correctly and UGENE crashes later.

Please file another bug with this error and fix it correctly. The test file you added has nothing to do with UGENE-6806.

rasputinkirill:
I checked performance with removed index on file with 100k annotations and found no problems.

src/corelibs/U2Formats/src/sqlite_dbi/SQLiteFeatureDbi.cpp
```diff
@@ -371,9 +364,6 @@ QSharedPointer<SQLiteQuery> SQLiteFeatureDbi::createFeatureQuery(const QString &
     if (useKeyTable) {
         tablesPart += ", FeatureKey AS fk";
     }
-    if (useRegion && (1 != fq.intersectRegion.length || oneClosestFeature)) {
-        tablesPart += ", FeatureLocationRTreeIndex AS fr";
```
yalgaer, line 375: This line defines "fr" as a reference to the FeatureLocatoinRTreeIndex.
You removed this line but "fr" is still referenced inside of the query: see "order by" and other parts with "fr".

Unfortunately our tests do not catch this situation, so we need new tests that will 
- With the old code:  call the 'createFeatureQuery' the way that "fr" is present in the result query -> check that it works
- With the new code: call the 'createFeatureQuery' with the same params, check that the result code still works.

The best candidates here are API tests, where you can call the method directly.

yalgaer (changes requested):

yalgaer:
Any updates?

rasputinkirill:
In this commit 73d7534 I altered table Feature with new field, and remove all 'fr' from code. There is no need in new tests because old unit tests check this changed methods. Also we do not need database updater because this changes affects project database.

yalgaer:
> In this commit [73d7534](https://github.com/ugeneunipro/ugene/commit/73d75347bfd3dc0de87941c8831a91aa09955faf) I altered table Feature with new field, and remove all 'fr' from code. There is no need in new tests because old unit tests check this changed methods. Also we do not need database updater because this changes affects project database.

I see 2 problems in your current solution.

Problem N1:
Originally you forgot to remove tree-index from SQL queries and no test is failed. It means that we do not have enough tests that check the code you change: 100% non-working code didn't trigger any errors

Solution for the problem N1: we need to add these tests. 
Could you please add these kind of tests? These may be direct API tests (part of XML tests).

Problem N2:
You added new 'end' field to the UGENE-db schema. UGENE-db can be used as external file format and not only session-db: check export annotations option.

If you change the schema you need to boost a version of the schema and add a schema converter + tests.

But I believe these is not needed at all: we do not need 'end' field, because we already have 'start' and 'length'. There is no benefit having 2 fields like 'start' and 'end' because only 1 index can be used at a time and we never use 'end' in the queries without 'start'.

Solution for the problem N2: remove 'end' field.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5691,6 +5691,14 @@ GUI_TEST_CLASS_DEFINITION(test_6760) {
     //Expected result: the annotation is present in another sequence view too.
     GTUtilsAnnotationsTreeView::findItem(os, "5_prime_UTR_intron");
 }
+
+GUI_TEST_CLASS_DEFINITION(test_6806) {
```
yalgaer: Question: can the problem be reproduced with a simple XML test?

yalgaer (commented):
All looks good. Have a question about GUI test and waiting for tests to finish

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5691,6 +5691,7 @@ GUI_TEST_CLASS_DEFINITION(test_6760) {
     //Expected result: the annotation is present in another sequence view too.
     GTUtilsAnnotationsTreeView::findItem(os, "5_prime_UTR_intron");
 }
+
```
DmitriiSukhomlinov: Remove this line

DmitriiSukhomlinov (dismissed):

yalgaer (approved):

PR #375, opened 2021-05-06
UGENE-5086. Notifications are shown above all opened windows

yalgaer:
Hi! could you please check that this initial patch shows notifications stack correctly on MacOS?
I added the explicit trigger for notifications: 1) Run UGENE 2) Open "Settings->Preferences" 3) Click "Help" to show a notification

Expected behavior: notification is shown above all UGENE windows but below all other windows.

I know there may be other issues with notifications: the goal of this patch is only to fix "other windows" issue and do not break the existing behavior.

Once you confirm that it works I will remove "Help button" notification trigger.


DmitriiSukhomlinov:
I will check when I come back to the office, I think, next week

DmitriiSukhomlinov:
Yep, it works. 

Screen Shot 2021-05-11 at 6 08 11 PM

DmitriiSukhomlinov:
Also, I think, we need to decide this notifications behavior, as far as I remember it's been changed already

PR #377, opened 2021-05-10
UGENE-7194. Show both codon and amino acid counts in Sequence View

yalgaer:
About this PR: this the second iteration of the "Codons stats" feature for our client. The first one has "Amino acids", this one adds nucleic triplets (codons).

It looks like the client is ready to accept the feature in this state, but I want to go further (next PR)

What in this PR: a new sub-group with "Codons stats" along with "Amino acids" stats. Both use the same cached data.

What will be done in the separate (next) PR: the contextual information about which region the statistics is computed for should be shown globally on the "stats" tab: a single label for the 1) nucleotides 2) di-nucleotides 3) codons 4) amino acids.


src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.cpp
```diff
@@ -353,30 +362,71 @@ void SequenceInfo::updateCodonsOccurrenceData(const QList<CharOccurResult> &codo
...
+                aminoTranslation->translate(codon.constData(), 3, &acidChar, 1);
+                if (acidChar != 0) {
+                    countPerAminoAcid[acidChar] += count;
+                }
+            }
+        }
+
+        // Prepare amino acids report.
+        QList<CharOccurResult> aminoAcidList;
+        for (auto it = countPerAminoAcid.begin(); it != countPerAminoAcid.end(); it++) {
```
DmitriiSukhomlinov, line 402: I prefer we never use iterators, cuz since C++11 we have range-based loops, which is more convenient

DmitriiSukhomlinov (approved):

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.cpp
yalgaer: I also do. 

In this specific case we have QMap and the only way to iterate map entries without extra copies/lookups is to use iterator.

Otherwise I have to 1) create a copy of map keys 2) iterate keys 3) use a lookup op by key in the loop.

PR #378, opened 2021-05-10
UGENE-7230. UGENE Genome alignment reports that reads can't be mapped.

yalgaer:
The fix itself is 1-liner:  writtenReadsCount++; in GenomeAlignerIO.cpp
but I also removed dead code from the related code & checked that all required fields are initialized.
See JIRA issue for more details: https://ugene.dev/tracker/browse/UGENE-7230

### DIFF
--- a/src/plugins/genome_aligner/src/GenomeAlignerIO.cpp
+++ b/src/plugins/genome_aligner/src/GenomeAlignerIO.cpp
@@ -251,6 +251,8 @@ GenomeAlignerDbiWriter::GenomeAlignerDbiWriter(const QString &dbiFilePath,
 }
 
 void GenomeAlignerDbiWriter::write(SearchQuery *seq, SAType offset) {
+    writtenReadsCount++;
+
     U2AssemblyRead read(new U2AssemblyReadData());
 
     read->name = seq->getName().toLatin1();
--- a/src/plugins/genome_aligner/src/GenomeAlignerTask.cpp
+++ b/src/plugins/genome_aligner/src/GenomeAlignerTask.cpp
@@ -45,6 +45,7 @@
 
 namespace U2 {
 
+//TODO: calling tr() int static context! Translator is not initialized yet!
 const QString GenomeAlignerTask::taskName(QObject::tr("UGENE Genome Aligner"));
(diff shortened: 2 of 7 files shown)


src/plugins/genome_aligner/src/GenomeAlignerTask.cpp
```diff
@@ -45,6 +45,7 @@
 
 namespace U2 {
 
+//TODO: calling tr() int static context! Translator is not initialized yet!
```
DmitriiSukhomlinov, line 48: Have we ever translated "UGENE Genome Aligner"?

yalgaer: I do not remember, you can check it in the translations file.

src/plugins/genome_aligner/src/GenomeAlignerIO.cpp
```diff
@@ -251,6 +251,8 @@ GenomeAlignerDbiWriter::GenomeAlignerDbiWriter(const QString &dbiFilePath,
 }
 
 void GenomeAlignerDbiWriter::write(SearchQuery *seq, SAType offset) {
+    writtenReadsCount++;
```
DmitriiSukhomlinov, line 254: This line fixed the problem?

src/plugins/genome_aligner/src/GenomeAlignerTask.cpp
DmitriiSukhomlinov: Yep, it's called "Модуль сборки контигов UGENE"

src/plugins/genome_aligner/src/GenomeAlignerIO.cpp
yalgaer: Yes

src/plugins/genome_aligner/src/GenomeAlignerTask.cpp
yalgaer: Plugins are loaded after the translator initialized: it may be safe for them use this kind of code.
But in general (for core-libs) this style is very unsafe.

smlnkcv (commented):
Approved. Test 1154 passed after fix.

PR #379, opened 2021-05-12
Add search data/ dir in path Contents/Resources/data for MacOS

ichebyki:
Fix gui tests running on MacOS
The problem is that in case of the Apple's bundle structure the gui tests harness can't find data/ dir

### DIFF
--- a/src/plugins/GUITestBase/src/harness/UGUITest.cpp
+++ b/src/plugins/GUITestBase/src/harness/UGUITest.cpp
@@ -90,7 +90,8 @@ static QString getDataDirImpl() {
         dataDir = QString("data/");
     }
 
-    if (!QFileInfo(dataDir).exists()) {
+    if (!QFileInfo(dataDir).exists() &&
+            ! BundleInfo::getDataSearchPath().isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/plugins/GUITestBase/src/harness/UGUITest.cpp
```diff
@@ -88,6 +89,10 @@ static QString getDataDirImpl() {
     } else {
         dataDir = QString("data/");
     }
+
+    if (!QFileInfo(dataDir).exists()) {
+        dataDir = BundleInfo::getDataSearchPath() + "/";
```
yalgaer, line 95: getDataSearchPath may return empty string and in this case you will have a root "/" dir as dataDir which is not correct.

yalgaer (changes requested):

src/plugins/GUITestBase/src/harness/UGUITest.cpp
ichebyki: yes, fixed

yalgaer (approved):

PR #382, opened 2021-05-17
UGENE-7125. Crash in InputWidgetsControllers.cpp:245.

evelinabis:
See [issue comment](https://ugene.dev/tracker/browse/UGENE-7125?focusedCommentId=30892&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30892).


src/corelibs/U2Gui/src/util/InputWidgetsControllers.cpp
```diff
@@ -241,7 +241,7 @@ void ComboBoxController::addParameterToCmdLineSettings(QStringList &settings) {
     }
     settings << cmdLinePrefix;
     int curIndex = inputWidget->currentIndex();
-    if (parameters.size() > curIndex) {
+    if (curIndex >= 0 && parameters.size() > curIndex) {
         settings << parameters.at(curIndex);
     } else {
         settings << inputWidget->currentText();
```
evelinabis: If `curIndex < 0`, an empty string will be added to the `settings`. Sometimes, this is bad behavior: an external tool will not work. Should I add some log message about this?

DmitriiSukhomlinov: Maybe, it'll be better to avoid this situation at all?

DmitriiSukhomlinov (dismissed):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -295,6 +299,48 @@ GUI_TEST_CLASS_DEFINITION(test_7106) {
...
+    // In the "Build Phylogenetic Tree" dialog select the PhyML Maximum Likelihood method.
+    // Press "Build".
+    //    Expected state: no crash.
+
+    class SaveSettingsScenario : public CustomScenario {
+    public:
+        void run(GUITestOpStatus &os) override {
+            QWidget *dialog = GTWidget::getActiveModalWidget(os);
+
+            GTUtilsDialog::waitForDialogWhichMayRunOrNot(os, new LicenseAgreementDialogFiller(os));
```
yalgaer: This should never be used: tests must have a predictive behavior.

src/corelibs/U2Gui/src/util/InputWidgetsControllers.cpp
yalgaer: Yes in this case some algorithm should be pre-selected by default and an invalid index value should be processed as an error.

yalgaer (changes requested):

smlnkcv (dismissed):

smlnkcv (dismissed):

src/corelibs/U2Gui/src/util/InputWidgetsControllers.cpp
```diff
@@ -239,13 +239,29 @@ void ComboBoxController::addParameterToCmdLineSettings(QStringList &settings) {
     if (cmdLinePrefix.isEmpty()) {
         return;
     }
-    settings << cmdLinePrefix;
+
     int curIndex = inputWidget->currentIndex();
```
yalgaer: Can we make it different: always have currentIndex() in the inputWidget valid?
Otherwise we have the following: 1) user see no value selected in the inputWidget, 2) we silently apply some 'default'.

evelinabis: I can add a check [here](https://github.com/ugeneunipro/ugene/blob/5e921c2f7f8016dcbd724e9cc191c3a30d7f8464/src/corelibs/U2Gui/src/util/InputWidgetsControllers.cpp#L225) that if the current index has not been set, use the default index (0, for example). What if the combobox is empty?

> silently 

Not silently.

yalgaer: > I can add a check [here](https://github.com/ugeneunipro/ugene/blob/5e921c2f7f8016dcbd724e9cc191c3a30d7f8464/src/corelibs/U2Gui/src/util/InputWidgetsControllers.cpp#L225) that if the current index has not been set.

Yes, this looks much better: ComboBox is a UI component and user will see the selected option

> use the default index (0, for example).

0 is good as the last chance.  I would check the 'defaultValue' passed in the constructor first.

> What if the combobox is empty?

The combobox should never be empty or it should be disabled/hidden. Or have an empty value as a valid option.
There should be a SAFE_POINT in the ComboBoxController constructor that options list is not empty.
I understand that this may require to create/fix extra issue or you can add the fix right into this PR: whatever is better for you.

> Not silently.

Dumping message to the log is not an excuse for us when we can avoid this situation completely.

smlnkcv (approved):

yalgaer (approved):

PR #384, opened 2021-05-20
UGENE-7238. Step 2. Extend available options to build popup menus in v…

yalgaer:
This is a second PR in the series to fix UGENE-7238.

This change adds an option to create 'any' (not only 'static' or 'context') kind of popup menus for object views.

This functionality is needed to implement the following scenario: plugins can populate menus like 'Align' or 'Align sequence to alignment'.

Today in order to build a context menu for the 'Align' button on the MSA Editor toolbar we have to use the following workaround:
https://github.com/ugeneunipro/ugene/blob/ea3a560789fe69252cb55d9fe70d3ff0de0bff05/src/corelibs/U2View/src/ov_msa/MSAEditor.cpp#L589

This code is not correct: it builds the whole menu, 'Align context menu' contains unrelated items to 'Align button'.

After my change there will be an option to call buildMenu() method with a 'align-button' type. Next I will add 'add-sequence-to-alignment' popup menu.

There should be no change in UGENE behavior after this PR: only internal API cleanup.

### DIFF
--- a/src/corelibs/U2Gui/src/ObjectViewModel.cpp
+++ b/src/corelibs/U2Gui/src/ObjectViewModel.cpp
@@ -43,6 +43,8 @@ namespace U2 {
 
 const QString GObjectViewState::APP_CLOSING_STATE_NAME("Auto saved");
 const GObjectViewFactoryId GObjectViewFactory::SIMPLE_TEXT_FACTORY("SimpleTextView");
+const QString GObjectViewMenuType::CONTEXT("gobject-view-menu-type-context");
+const QString GObjectViewMenuType::STATIC("object-view-menu-type-static");
 
 void GObjectViewState::setViewName(const QString &newName) {
     // this method is not a real state modification: state caches view name as a reference, but not its internal data
@@ -221,6 +223,9 @@ void GObjectView::sl_onDocumentRemoved(Document *d) {
     }
 }
 
+void GObjectView::sl_onDocumentLoadedStateChanged() {
+}
+
 void GObjectView::sl_onObjectNameChanged(const QString &oldName) {
     CHECK(AppContext::getProject() != nullptr, );
     GObject *object = qobject_cast<GObject *>(sender());
@@ -248,8 +253,8 @@ void GObjectView::buildStaticToolbar(QToolBar *tb) {
     emit si_buildStaticToolbar(this, tb);
 }
 
-void GObjectView::buildStaticMenu(QMenu *m) {
-    emit si_buildStaticMenu(this, m);
+void GObjectView::buildMenu(QMenu *m, const QString &type) {
+    emit si_buildMenu(this, m, type);
 }
 
 // Returns true if view  contains this object
@@ -370,7 +375,7 @@ void GObjectViewWindow::setupMDIToolbar(QToolBar *tb) {
 }
 
 void GObjectViewWindow::setupViewMenu(QMenu *m) {
-    view->buildStaticMenu(m);
+    view->buildMenu(m, GObjectViewMenuType::STATIC);
 }
 
 //////////////////////////////////////////////////////////////////////////
@@ -577,8 +582,7 @@ void GObjectViewWindowContext::sl_windowAdded(MWMDIWindow *w) {
 
     initViewContext(objectView);
 
-    connect(objectView, SIGNAL(si_buildPopupMenu(GObjectView *, QMenu *)), SLOT(sl_buildContextMenu(GObjectView *, QMenu *)));
-    connect(objectView, SIGNAL(si_buildStaticMenu(GObjectView *, QMenu *)), SLOT(sl_buildStaticMenu(GObjectView *, QMenu *)));
+    connect(objectView, SIGNAL(si_buildMenu(GObjectView *, QMenu *, const QString &)), SLOT(sl_buildMenu(GObjectView *, QMenu *, const QString &)));
 }
 
 void GObjectViewWindowContext::sl_windowClosing(MWMDIWindow *w) {
@@ -590,15 +594,11 @@ void GObjectViewWindowContext::sl_windowClosing(MWMDIWindow *w) {
     disconnectView(objectView);
 }
 
-void GObjectViewWindowContext::sl_buildContextMenu(GObjectView *v, QMenu *m) {
-    buildMenu(v, m);
+void GObjectViewWindowContext::sl_buildMenu(GObjectView *v, QMenu *m, const QString &type) {
+    buildMenu(v, m, type);
 }
 
-void GObjectViewWindowContext::sl_buildStaticMenu(GObjectView *v, QMenu *m) {
-    buildMenu(v, m);
-}
-
-void GObjectViewWindowContext::buildMenu(GObjectView *, QMenu *) {
+void GObjectViewWindowContext::buildMenu(GObjectView *, QMenu *, const QString &) {
     // No menu by default.
 }
 
@@ -679,6 +679,19 @@ void GObjectViewAction::addToMenuWithOrder(QMenu *menu) {
         }
     }
     menu->addAction(this);
+}
+
+GObjectViewObjectHandler::~GObjectViewObjectHandler() {
+}
+
+bool GObjectViewObjectHandler::canHandle(GObjectView *, GObject *) {
+    return false;
+}
+
+void GObjectViewObjectHandler::onObjectAdded(GObjectView *, GObject *) {
+}
+
+void GObjectViewObjectHandler::onObjectRemoved(GObjectView *, GObject *) {
 }
 
 }    // namespace U2
--- a/src/corelibs/U2Gui/src/ObjectViewModel.cpp
+++ b/src/corelibs/U2Gui/src/ObjectViewModel.cpp
@@ -595,11 +595,29 @@ void GObjectViewWindowContext::sl_windowClosing(MWMDIWindow *w) {
 }
 
 void GObjectViewWindowContext::sl_buildMenu(GObjectView *v, QMenu *m, const QString &type) {
-    buildMenu(v, m, type);
+    if (type == GObjectViewMenuType::STATIC) {
+        buildStaticMenu(v, m);
+    } else if (type == GObjectViewMenuType::CONTEXT) {
+        buildContextMenu(v, m);
+    } else {
+        buildActionMenu(v, m, type);
+    }
 }
 
-void GObjectViewWindowContext::buildMenu(GObjectView *, QMenu *, const QString &) {
-    // No menu by default.
+void GObjectViewWindowContext::buildStaticMenu(GObjectView *view, QMenu *menu) {
+    buildStaticOrContextMenu(view, menu);
+}
+
+void GObjectViewWindowContext::buildContextMenu(GObjectView *view, QMenu *menu) {
+    buildStaticOrContextMenu(view, menu);
+}
+
+void GObjectViewWindowContext::buildStaticOrContextMenu(GObjectView *, QMenu *) {
+    // No extra static/context menu items by default.
+}
+
+void GObjectViewWindowContext::buildActionMenu(GObjectView *, QMenu *, const QString &) {
+    // No extra action menu items by default.
 }
 
 void GObjectViewWindowContext::disconnectView(GObjectView *v) {
--- a/src/corelibs/U2Gui/src/ObjectViewModel.cpp
+++ b/src/corelibs/U2Gui/src/ObjectViewModel.cpp
@@ -699,9 +699,6 @@ void GObjectViewAction::addToMenuWithOrder(QMenu *menu) {
     menu->addAction(this);
 }
 
-GObjectViewObjectHandler::~GObjectViewObjectHandler() {
-}
-
 bool GObjectViewObjectHandler::canHandle(GObjectView *, GObject *) {
     return false;
 }
--- a/src/corelibs/U2Gui/src/ObjectViewModel.h
+++ b/src/corelibs/U2Gui/src/ObjectViewModel.h
@@ -372,7 +372,7 @@ private:
  */
 class U2GUI_EXPORT GObjectViewObjectHandler {
 public:
-    virtual ~GObjectViewObjectHandler();
+    virtual ~GObjectViewObjectHandler() = default;
 
     /**
      * Checks if the handler can 'handle' the object.
--- a/src/corelibs/U2Gui/src/ObjectViewModel.h
+++ b/src/corelibs/U2Gui/src/ObjectViewModel.h
@@ -179,7 +179,7 @@ public:
 
     virtual void buildStaticToolbar(QToolBar *tb);
 
-    virtual void buildStaticMenu(QMenu *n);
+    virtual void buildMenu(QMenu *menu, const QString &type);
 
     virtual bool canAddObject(GObject *obj);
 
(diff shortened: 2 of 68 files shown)


DmitriiSukhomlinov:
`This code is not correct: it builds the whole menu, 'Align context menu' contains unrelated items to 'Align button'.`

Sorry, I don't get it? could you explain, please&

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -43,6 +43,8 @@ namespace U2 {
 
 const QString GObjectViewState::APP_CLOSING_STATE_NAME("Auto saved");
 const GObjectViewFactoryId GObjectViewFactory::SIMPLE_TEXT_FACTORY("SimpleTextView");
+const QString GObjectViewMenuType::CONTEXT("gobject-view-menu-type-context");
```
DmitriiSukhomlinov, line 46: I think, `enum` will be much better here

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -681,4 +681,17 @@ void GObjectViewAction::addToMenuWithOrder(QMenu *menu) {
     menu->addAction(this);
 }
 
+GObjectViewObjectHandler::~GObjectViewObjectHandler() {
```
DmitriiSukhomlinov: Remove as far as it's empty

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -248,8 +253,8 @@ void GObjectView::buildStaticToolbar(QToolBar *tb) {
     emit si_buildStaticToolbar(this, tb);
 }
 
-void GObjectView::buildStaticMenu(QMenu *m) {
-    emit si_buildStaticMenu(this, m);
+void GObjectView::buildMenu(QMenu *m, const QString &type) {
```
DmitriiSukhomlinov, line 256: Could you rename this function as `emitBuildMenu`? Cuz now it looks like this function builds menu itself, but it just emits a signal

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -221,6 +223,9 @@ void GObjectView::sl_onDocumentRemoved(Document *d) {
     }
 }
 
+void GObjectView::sl_onDocumentLoadedStateChanged() {
```
DmitriiSukhomlinov, line 226: Make this function pure virtual

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -590,15 +594,11 @@ void GObjectViewWindowContext::sl_windowClosing(MWMDIWindow *w) {
...
-}
-
-void GObjectViewWindowContext::sl_buildStaticMenu(GObjectView *v, QMenu *m) {
-    buildMenu(v, m);
+void GObjectViewWindowContext::sl_buildMenu(GObjectView *v, QMenu *m, const QString &type) {
+    buildMenu(v, m, type);
 }
 
-void GObjectViewWindowContext::buildMenu(GObjectView *, QMenu *) {
+void GObjectViewWindowContext::buildMenu(GObjectView *, QMenu *, const QString &) {
```
DmitriiSukhomlinov: Should be pure virtual

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -681,4 +681,17 @@ void GObjectViewAction::addToMenuWithOrder(QMenu *menu) {
...
 }
 
+GObjectViewObjectHandler::~GObjectViewObjectHandler() {
+}
+
+bool GObjectViewObjectHandler::canHandle(GObjectView *, GObject *) {
+    return false;
+}
+
+void GObjectViewObjectHandler::onObjectAdded(GObjectView *, GObject *) {
```
DmitriiSukhomlinov, line 706: Pure virtual

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -681,4 +681,17 @@ void GObjectViewAction::addToMenuWithOrder(QMenu *menu) {
...
+}
+
+void GObjectViewObjectHandler::onObjectRemoved(GObjectView *, GObject *) {
```
DmitriiSukhomlinov, line 709: Pure virtual

src/corelibs/U2Gui/src/ObjectViewModel.cpp
```diff
@@ -681,4 +681,17 @@ void GObjectViewAction::addToMenuWithOrder(QMenu *menu) {
     menu->addAction(this);
 }
 
+GObjectViewObjectHandler::~GObjectViewObjectHandler() {
+}
+
+bool GObjectViewObjectHandler::canHandle(GObjectView *, GObject *) {
```
DmitriiSukhomlinov, line 702: Check if this function is used somewhere. It yes - leave it as now, if not - make pure virtual

src/corelibs/U2Gui/src/ObjectViewModel.h
```diff
@@ -245,6 +248,16 @@ protected slots:
     OptionsPanel *optionsPanel;
 };
 
+/** Constants for known GObject view menu types. */
+class U2GUI_EXPORT GObjectViewMenuType {
+public:
+    /** A constant for building a contextual (popup) menu. See buildMenu for details. */
+    const static QString CONTEXT;
```
DmitriiSukhomlinov, line 255: I think, it's better to use `enum`

src/corelibs/U2Gui/src/ObjectViewModel.h
```diff
@@ -353,24 +366,25 @@ class U2GUI_EXPORT GObjectViewAction : public QAction {
     int actionOrder;
 };
 
+/**
+ * Abstraction of the object handler for a view.
```
DmitriiSukhomlinov, line 370: It's not "abstraction" now, but you could make it abstract by marking function as pure virtual

src/corelibs/U2View/src/ov_assembly/AssemblyBrowser.h
```diff
@@ -54,11 +54,12 @@ class AssemblyBrowser : public GObjectView {
...
-    virtual void buildStaticMenu(QMenu *m);
-    virtual QVariantMap saveState();
-    virtual Task *updateViewTask(const QString &stateName, const QVariantMap &stateData);
-    virtual OptionsPanel *getOptionsPanel();
+    void buildStaticToolbar(QToolBar *tb) override;
+    void buildMenu(QMenu *menu, const QString &type) override;
+
+    QVariantMap saveState() override;
+    Task *updateViewTask(const QString &stateName, const QVariantMap &stateData) override;
+    OptionsPanel *getOptionsPanel() override;
```
DmitriiSukhomlinov, line 62: `OptionsPanel *getOptionsPanel() const override;`

src/corelibs/U2View/src/ov_assembly/AssemblyBrowser.h
```diff
@@ -247,9 +248,9 @@ private slots:
     QAction *posSelectorAction;
     PositionSelector *posSelector;
     QList<QAction *> overviewScaleTypeActions;
-    QAction *showCoordsOnRulerAction;
+    QAction *showCoordsOnRulerAction = nullptr;
```
DmitriiSukhomlinov, line 251: Why just these two initialized with `nullptr`?

src/corelibs/U2View/src/ov_msa/McaEditor.cpp
```diff
@@ -128,8 +132,8 @@ SequenceObjectContext *McaEditor::getReferenceContext() const {
 
 void McaEditor::sl_onContextMenuRequested(const QPoint & /*pos*/) {
     QMenu menu;
-    buildStaticMenu(&menu);
-    emit si_buildPopupMenu(this, &menu);
+    buildMenu(&menu, GObjectViewMenuType::STATIC);    // TODO: this call triggers extra signal for static menu.
```
DmitriiSukhomlinov, line 135: What extra signal and what are you about to do with it?

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
```diff
@@ -674,51 +674,51 @@ void AnnotationsTreeView::sl_onAnnotationsRemoved(const QList<Annotation *> &as)
...
-            }
-        } break;
-        case AnnotationModification_AddedToGroup: {
-            const AnnotationGroupModification &gmd = static_cast<const AnnotationGroupModification &>(annotationModification);
-            AVGroupItem *gi = findGroupItem(gmd.getGroup());
-            SAFE_POINT(NULL != gi, L10N::nullPointerError("annotation view group item"), );
-            buildAnnotationTree(gi, gmd.annotation);
-            gi->updateVisual();
-        } break;
+            } break;
```
DmitriiSukhomlinov, line 693: Newline missed

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
```diff
@@ -674,51 +674,51 @@ void AnnotationsTreeView::sl_onAnnotationsRemoved(const QList<Annotation *> &as)
...
+                    }
+                }
+            } break;
+            case AnnotationModification_AddedToGroup: {
+                const AnnotationGroupModification &gmd = static_cast<const AnnotationGroupModification &>(annotationModification);
+                AVGroupItem *gi = findGroupItem(gmd.getGroup());
+                SAFE_POINT(NULL != gi, L10N::nullPointerError("annotation view group item"), );
+                buildAnnotationTree(gi, gmd.annotation);
+                gi->updateVisual();
+            } break;
```
DmitriiSukhomlinov, line 711: And here. And everywhere it the same place

DmitriiSukhomlinov (changes requested):
See comments. Also - why is SO MUCH refactoring in your commit?

yalgaer:
> `This code is not correct: it builds the whole menu, 'Align context menu' contains unrelated items to 'Align button'.`
> 
> Sorry, I don't get it? could you explain, please&

The code I pointed builds the whole context menu just to pick "Align" part out of it. This looks like a workaround. More than that it is not correct: "Align" button may not want to include everything under "Align" menu - they are semantically different.

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: Default descructors are not virtual and here is virtual one. I have not changed this code, just moved it from .h to .cpp (our code style agreement)

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: The current name is correct: the view should build a menu as the result of the method call.

The signal "si_buildMenu" is the way to notify view clients (plugins) to add their action into the menu.

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: This is not correct: this method is not required to be implemented if nothing should be done on document state change (implementation is optional)

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: Same here: menu is optional.

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: Same here: implementation is optional.

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: Same here: impl is optional.

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: It is a core function for the Handler class. Of course it is used.

src/corelibs/U2Gui/src/ObjectViewModel.h
yalgaer: Enums are sealed (closed) types. Here we have an open set of values that can be extended by views. I described the way I plan extending it in MSA editor.

src/corelibs/U2Gui/src/ObjectViewModel.h
yalgaer: It is abstraction. Abstract classes/interfaces can have default impls.

src/corelibs/U2View/src/ov_assembly/AssemblyBrowser.h
yalgaer: I can't make it const because it is not designed to be const. I think name 'get' is misleading here but this kind of refactoring is out of scope of this huge PR: I just added 'override'.

src/corelibs/U2View/src/ov_assembly/AssemblyBrowser.h
yalgaer: These values were not initialized at all. 
I think we should initialize all primitive type values in-place, but this again out of scope of this PR: it is already very huge.

src/corelibs/U2View/src/ov_msa/McaEditor.cpp
yalgaer: buildMenu emits a signal. This is old behavior and it was not changed by me.
I do not plan to do anything inside this PR, just documented this problem in the code.

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
yalgaer: This is how our clangformat style is set up. I also do not like this style - we need to update clangformat and reformat all sources. Will do it in the separate PR next.

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
yalgaer: Same here

yalgaer:
> See comments. Also - why is SO MUCH refactoring in your commit?

I tried to make it minimal to reach the goal pointed in the description.
If something can be excluded, please point me to that place and I will remove it from the PR

src/corelibs/U2Gui/src/ObjectViewModel.cpp
DmitriiSukhomlinov: In this case remove it and use the `default` keyword like `virtual ~GObjectViewObjectHandler() = default;`

src/corelibs/U2Gui/src/ObjectViewModel.h
DmitriiSukhomlinov: Strongly say, abstract class in C++ is a class that has at least one pure virtual function, interfaces are not presented in C++ at all, but we can define them as "abstract classes where the all functions are pure virtual". This class doesn't have pure virtual functions at all, so it's just the base class with the overridable content. Bit it's about word choice, I won't insist.

src/corelibs/U2Gui/src/ObjectViewModel.cpp
DmitriiSukhomlinov: Not really agree, but OK

DmitriiSukhomlinov (changes requested):
Just one small comment about `default`

src/corelibs/U2Gui/src/ObjectViewModel.cpp
yalgaer: Good idea. Fixed

PR #385, opened 2021-05-20
UGENE-7135. Crash on database error.

evelinabis:
Can't reproduced. The user just opened the file GFX0437388_variants.filtered.EEF.snp.gb and then there was a database error and a crash. I don't know if my fixes will work, how to check them. Any ideas?
Added `return` before the crash.

### DIFF
--- a/src/corelibs/U2Formats/src/EMBLGenbankAbstractDocument.cpp
+++ b/src/corelibs/U2Formats/src/EMBLGenbankAbstractDocument.cpp
@@ -203,6 +203,7 @@ void EMBLGenbankAbstractDocument::load(const U2DbiRef &dbiRef, IOAdapter *io, QL
                 }
                 CHECK_OP(os, );
             }
+            CHECK_OP(os, )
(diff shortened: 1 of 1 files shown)


src/corelibs/U2Formats/src/EMBLGenbankAbstractDocument.cpp
```diff
@@ -203,6 +203,7 @@ void EMBLGenbankAbstractDocument::load(const U2DbiRef &dbiRef, IOAdapter *io, QL
                 }
                 CHECK_OP(os, );
             }
+            CHECK_OP(os, )
```
DmitriiSukhomlinov, line 206: This fix do nothing. OperationStatus has already been checked above.

DmitriiSukhomlinov:
I checked the report and I'm not sure, that you found the reason of this crash property. I think, there is no need of trying to guess, you either reproduce this crash and fix it or can't reproduce and don't touch at all

evelinabis:
Ok then I close the pull request and the issue as non-reproducible.

PR #387, opened 2021-05-22
UGENE-7238. Step 4. Prepare a drop down for "Align sequence to Alignm…

yalgaer:
In this PR:
1) Align sequence to alignment logic was moved out from the Msa Editor into a dedicated place.
2) "Align sequence(s) to alignment" button now shows a drop down menu like "Align" button.
3) Muscle plugin actions to align sequences to alignment were moved from "Align" to "Align sequence(s) to alignment" menu.
4) The built-in MAFFT action was added as a menu item to the "Align sequence(s) to alignment".

In the next steps:
1) Add support for ClustalO
2) Build menu items for all "AddToAlignment" AlignmentAlgorithmType types, not only for MAFFT

### DIFF
--- a/src/corelibs/U2View/U2View.pro
+++ b/src/corelibs/U2View/U2View.pro
@@ -96,6 +96,7 @@ HEADERS += src/LicenseDialog.h \
            src/ov_msa/MsaOpSavableTab.h \
            src/ov_msa/MSASelectSubalignmentDialog.h \
            src/ov_msa/MsaEditorUserModStepController.h \
+           src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.h \
            src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentTask.h \
            src/ov_msa/helpers/BaseWidthController.h \
            src/ov_msa/helpers/DrawHelper.h \
@@ -371,6 +372,7 @@ SOURCES += src/LicenseDialog.cpp \
            src/ov_msa/MsaOpSavableTab.cpp \
            src/ov_msa/MSASelectSubalignmentDialog.cpp \
            src/ov_msa/MsaEditorUserModStepController.cpp \
+           src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp \
            src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentTask.cpp \
            src/ov_msa/helpers/BaseWidthController.cpp \
            src/ov_msa/helpers/DrawHelper.cpp \
--- a/src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
+++ b/src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
@@ -51,7 +51,7 @@ void AlignSequencesToAlignmentSupport::initViewContext(GObjectView *view) {
     //  The final code should build action per each algorithm registered in the AlignmentAlgorithmsRegistry.
     auto alignAction = new AlignSequencesToAlignmentAction(this, msaEditor, tr("Add sequences with MAFFT..."), 100);
     alignAction->setIcon(QIcon(":/core/images/add_to_alignment.png"));    //TODO: add a dedicated icon for MAFFT.
-    alignAction->setObjectName("Align with MAFFT");
+    alignAction->setObjectName("Add sequences to alignment with MAFFT");
     alignAction->setMenuTypes({MsaEditorMenuType::ALIGN_SEQUENCES_TO_ALIGNMENT});
     alignAction->sl_updateState();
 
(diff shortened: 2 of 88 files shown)


src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
```diff
@@ -0,0 +1,126 @@
...
+    connect(alignAction, SIGNAL(triggered()), SLOT(sl_alignSequencesToAlignment()));
+    addViewAction(alignAction);
+}
+
+void AlignSequencesToAlignmentSupport::sl_alignSequencesToAlignment() {
+    auto action = qobject_cast<AlignSequencesToAlignmentAction *>(sender());
+    SAFE_POINT(action != nullptr, "Not a AlignSequencesToAlignmentAction!", );
+    auto msaEditor = action->getEditor();
+    auto msaObject = msaEditor->getMaObject();
+    CHECK(!msaObject->isStateLocked(), );
```
DmitriiSukhomlinov: Add a note to log in this case, cuz a user, possibly, won't understand why it doesn't work

src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    const GObjectSelection *selection = projectView->getGObjectSelection();
+    SAFE_POINT(selection != nullptr, "GObjectSelection is null", );
+
+    QList<GObject *> objects = selection->getSelectedObjects();
+    bool selectFromProject = !objects.isEmpty();
+
+    for (const GObject *object : qAsConst(objects)) {
+        const GObjectType &objectType = object->getGObjectType();
+        if (object == msaObject || (objectType != GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT && objectType != GObjectTypes::SEQUENCE)) {
```
DmitriiSukhomlinov, line 80: I can't get it. What kind of objects are you looking for here?

src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
```diff
@@ -0,0 +1,126 @@
...
+            selectFromProject = false;
+            break;
+        }
+    }
+
+    // TODO: ask user if to select from the project or from files.
+    if (selectFromProject) {
+        SequenceObjectsExtractor extractor;
```
DmitriiSukhomlinov, line 88: I think, it's better to log this branch - it could be not obvious where the aligned sequences were got from

src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
```diff
@@ -0,0 +1,126 @@
...
+        extractor.setAlphabet(msaObject->getAlphabet());
+        extractor.extractSequencesFromObjects(objects);
+
+        if (!extractor.getSequenceRefs().isEmpty()) {
+            auto task = new AlignSequencesToAlignmentTask(msaObject, extractor);
+            TaskWatchdog::trackResourceExistence(msaObject, task, tr("A problem occurred during adding sequences. The multiple alignment is no more available."));
+            AppContext::getTaskScheduler()->registerTopLevelTask(task);
+        }
+    } else {
+        QString filter = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
```
DmitriiSukhomlinov, line 98: It's possible to choose alignments from the project, but here sequences only are available? I think it's better to unify it.

src/corelibs/U2View/U2View.pro
```diff
@@ -96,6 +96,7 @@ HEADERS += src/LicenseDialog.h \
            src/ov_msa/MsaOpSavableTab.h \
            src/ov_msa/MSASelectSubalignmentDialog.h \
            src/ov_msa/MsaEditorUserModStepController.h \
+           src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.h \
```
DmitriiSukhomlinov, line 99: We usually name the inner directories is a "snake_case" style. Like `align_sequences_to_alignment`

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2View/U2View.pro
yalgaer: I agree, but I have not added any folders here + there are 10 other subfolders on the same level to be fixed.
Will fix them all in the separate PR.

src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
yalgaer: Changed to SAFE_POINT. This should never happen. The original check is a part of the old code.

src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
yalgaer: Same here: this is the unchanged original code moved out of MSAEditor.cpp into this class.
I think we should not rely on the log at all in UGENE and use very explicit UX. I added a "TODO" comment above.

src/corelibs/U2View/src/ov_msa/AlignSequencesToAlignment/AlignSequencesToAlignmentSupport.cpp
yalgaer: Same here: out of scope of this PR: this is an old behavior copied into the new cpp file out of MSAEditor.cpp.

DmitriiSukhomlinov (approved):

PR #389, opened 2021-05-25
Revert "Update ugene_version.pri"

DmitriiSukhomlinov:
Reverts ugeneunipro/ugene#315

I just found out, that it's incorrect. The "release" brunch should contain the last released version, not the current developing one


yalgaer (approved):
We won't use anonymous 'release' branch anymore. 
There will be only 'release-39', 'release-40' etc branches + the current (master) branch.

PR #390, opened 2021-05-26
UGENE-7151. Crash when opening file if "Sanger Reads" is running.

evelinabis:
https://ugene.dev/tracker/browse/UGENE-7151


yalgaer (approved):
The patch is good but the overall API of HoverQLabel should be improved. This can be done in the separate PR/bug
What should be improved:
1) clicked -> si_activated
2) mousePressEvent -> mouseReleaseEvent.

PR #392, opened 2021-05-28
UGENE-7238. Step 6. Use snake_case for 'align-to-alignment' and other…

yalgaer:
This was a request from the previous PR (UGENE-7238 step 5) that I promised to fix in the separate PR.

No any changes is expected, only internal dirs rename.


src/include/U2View/AddSequencesToAlignmentTask.h
```diff
@@ -1 +1,2 @@
 #include "../../corelibs/U2View/src/ov_msa/AddSequencesToAlignmentTask.h"
+
```
DmitriiSukhomlinov: Could be removed

DmitriiSukhomlinov (dismissed):

src/include/U2View/AddSequencesToAlignmentTask.h
yalgaer: Removed the empty line

PR #393, opened 2021-05-30
UGENE-7238. Step 7. Support 'Align-to-Alignment' with ClustalO algori…

yalgaer:
Next step of "Align-sequences-to-alignment" feature support: ClustalO support is added. "Align-sequences-to-alignment" menu now have "ClustalO" variant of the algorithm.

I refactored AlignMsaAction used by different external tools to include a common functionality and remove some copy-paste from these tools.


src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
```diff
@@ -67,7 +67,7 @@ qint64 U2SequenceUtils::length(const U2EntityRef &ref, U2OpStatus &os) {
 
 U2Sequence U2SequenceUtils::copySequence(const DNASequence &srcSeq, const U2DbiRef &dstDbi, const QString &dstFolder, U2OpStatus &os) {
     U2Sequence res;
-    res.alphabet = srcSeq.alphabet->getId();
+    res.alphabet = srcSeq.alphabet == nullptr ? nullptr : srcSeq.alphabet->getId();
```
DmitriiSukhomlinov, line 71: Does it possible? Maybe `SAFE_POINT` is better?

DmitriiSukhomlinov (dismissed):

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
yalgaer: Yes it is possible and looks like a big MSA related bug/feature.
Any new MSA object read from a ClustalW file has no per-sequence alphabet, but has only global per-MSA alphabet.

I decided not to change this behavior in my commit and address is separately.

yalgaer: Added TODO, will fix this issue in the separate PR. Need to re-check all formats

PR #395, opened 2021-06-01
UGENE-6807 Crash on running Generate alignment profile

rasputinkirill:
As I discovered, this crash is not caused by the C ++ exception system, but by QT, which wraps and converts them into its own QtMessages. Yes, they can be handled, but this mechanism prevents you from recovering from an error and continuing to run the program.
https://ru.stackoverflow.com/questions/664753/qvector-out-of-memory
Also I discovered the following thing, even if you generate a grid report for a large alignment view, it can be very slow. Experimentally I've found that the alignment size when displaying the report to the grid is acceptable. Then I added a warning before creating the grid report, if the alignment length is too long.


DmitriiSukhomlinov (dismissed):

yalgaer (changes requested):
I see 2 problems with this solution:
1. No GUI test
2. You check only 1 dimension of the MSA, but "Grid Profile" is 2D

This could be a "minimal" solution. The best solution is to move grid profile computation into a separate process and generate an HTML file which can be opened by a browser.

rasputinkirill:
As I think there is no need to check 'height' of alignment because of meaning of multiple aligment. Multiple alignment is designed to check differences and similarities between similar organisms and with increasing of their number alignment will start to lose it meaning. In our samples biggest height of alignment is 200 sequence and Im sure there is no alignment with 1k sequences or more. So I think it is no need to check alignment height here.

yalgaer:
> As I think there is no need to check 'height' of alignment because of meaning of multiple aligment. Multiple alignment is designed to check differences and similarities between similar organisms and with increasing of their number alignment will start to lose it meaning. In our samples biggest height of alignment is 200 sequence and Im sure there is no alignment with 1k sequences or more. So I think it is no need to check alignment height here.

Question: 
You set GRID_PROFILE_LENGTH_LIMIT to 20000 (length, horizontal dimension).
If we have 2 sequences in an alignment, should the horizontal dimension be divided by 2? 
As far as I understand the problem is with the total report size (width * height), not with a 1 row only?

About alignment size:
Check this popular database: http://pfam.xfam.org/family/browse

rasputinkirill:
I've checked performance on 22k height x 1.1k length alignment from http://pfam.xfam.org/family/PF19278/alignment/uniprot/gzipped and found no performance problems with such alignment.
The dimension of resulting grid profile table depends only on sequences maximum length and  alphabet of the current alignment.

yalgaer:
> The dimension of resulting grid profile table depends only on sequences maximum length and alphabet of the current alignment.

I see now and I'm surprised, because in this case I would expect completely different solution.

As far as I understand the only 'memory' problems we may have with this report are related to 2D structure storage and visualization: the generated report string looks like "<td style>somenumber</td>" which is about 10 chars per cell (20 bytes per cell)

The linear (length dependent) algorithm should never have memory issues at all: much earlier than any problems arise with 1D algorithm we should see problems with 2D report.

Question:
If the original code works for some length (10k) why can't we process the whole alignment  by blocks of 10k?
For example, for the length 25k we can compute these blocks: 10k + 10k + 5k?

src/plugins/dna_stat/src/DNAStatPlugin.cpp
```diff
@@ -92,6 +93,13 @@ void DNAStatMSAEditorContext::buildStaticOrContextMenu(GObjectView *v, QMenu *m)
 void DNAStatMSAEditorContext::sl_showMSAProfileDialog() {
     GObjectViewAction *viewAction = qobject_cast<GObjectViewAction *>(sender());
     MSAEditor *msaEd = qobject_cast<MSAEditor *>(viewAction->getObjectView());
+    if (msaEd->getAlignmentLen() >= GRID_PROFILE_LENGTH_LIMIT) {
```
yalgaer, line 97: Here is the best solution I see here:
1) If the profile is too long: force "Save to file" option to be ON.
2) Show warning that "Profile is too large and won't be opened in UGENE" and do not even try opening it.

This is much better than saying that the next user step will potentially "crash" UGENE: there are no crash at all here.

More details:
- The warning should be shown not as a separate annoying popup, but as a context label at the bottom of the dialog.
- You are right that opening/browsing very large profiles in UGENE viewer has no value. So as an optional extension you can even add a check box "Open profile in UGENE" which can be disabled for large alignments.

yalgaer (changes requested):
See comments inside.

src/plugins/GUITestBase/src/runnables/ugene/corelibs/U2View/ov_msa/GenerateAlignmentProfileDialogFiller.cpp
```diff
@@ -35,8 +35,22 @@
 
 namespace U2 {
 
+GenerateAlignmentProfileDialogFiller::GenerateAlignmentProfileDialogFiller(HI::GUITestOpStatus &os, CustomScenario *customScenario)
+    : Filler(os, "DNAStatMSAProfileDialog", customScenario),
+      counts(true),
```
yalgaer: nit: initialize in the header.

src/plugins/dna_stat/src/DNAStatMSAProfileDialog.cpp
```diff
@@ -123,6 +124,14 @@ void DNAStatMSAProfileDialog::accept() {
     QDialog::accept();
 }
 
+void DNAStatMSAProfileDialog::alignmentIsTooBig() {
+    warningLabel->setText(tr("<b><font color=%1>%2</font><br></br></b>")
+        .arg(Theme::errorColorLabelHtmlStr())
+        .arg(tr("Warning: Alignment is too big. Generated grid profile couldn't be shown properly or even will crash application.<br>You can open generated profile by your own.")));
```
yalgaer: Change the long text to a simple "Warning: report is too big to be shown in UGENE"

src/plugins/dna_stat/src/DNAStatMSAProfileDialog.h
```diff
@@ -43,7 +43,7 @@ class DNAStatMSAProfileDialog : public QDialog, public Ui_DNAStatMSAProfileDialo
     DNAStatMSAProfileDialog(QWidget *p, MSAEditor *ctx);
 
     void accept();
-
+    void alignmentIsTooBig();
```
yalgaer: Start method names with a verb: rename to showAlignmentIsTooBigWarning() or switchIntoBigAlignmentMode()

src/plugins/dna_stat/src/DNAStatPlugin.h
```diff
@@ -49,6 +49,9 @@ protected slots:
 
 protected:
     void initViewContext(GObjectView *view) override;
+
+private:
+    static constexpr int GRID_PROFILE_LENGTH_LIMIT = 20000;
```
yalgaer, line 55: Add a minimal doc why do we have this constant.

yalgaer (changes requested):
The solution is good, I added only style-level comments.
Please merge with master, run UI tests and I will approve.

yalgaer (commented):
Looks like you need to merge with master before running the tests.
Waiting for the tests result and will approve.

DmitriiSukhomlinov (approved):

yalgaer (approved):

PR #397, opened 2021-06-01
Add exclude list option for gui tests running

ichebyki:
The options will help gui-test running on macos

### DIFF
--- a/src/corelibs/U2Core/src/cmdline/CMDLineCoreOptions.cpp
+++ b/src/corelibs/U2Core/src/cmdline/CMDLineCoreOptions.cpp
@@ -43,6 +43,7 @@ const QString CMDLineCoreOptions::LAUNCH_GUI_TEST = "gui-test";
 const QString CMDLineCoreOptions::LAUNCH_GUI_TEST_NO_IGNORED = "gui-test-no-ignored";
 const QString CMDLineCoreOptions::LAUNCH_GUI_TEST_BATCH = "gui-test-batch";
 const QString CMDLineCoreOptions::LAUNCH_GUI_TEST_SUITE = "gui-test-suite";
+const QString CMDLineCoreOptions::LAUNCH_GUI_TEST_EXCLUDE_LIST = "gui-test-exclude-list";
 const QString CMDLineCoreOptions::LAUNCH_GUI_TEST_CRAZY_USER = "gui-test-crazy-user";
 const QString CMDLineCoreOptions::USAGE = "usage";
 const QString CMDLineCoreOptions::DOWNLOAD_DIR = "download-dir";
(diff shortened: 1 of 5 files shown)


ichebyki:
The issue UGENE-7253 was created

yalgaer (commented):
This option is not needed and here are the reasons why

1) You can mark test as "Ignored on Mac" in the place of the registration. The test won't run on Mac in this case.

2) You can use already supported "gui-test-suite" option to create "include-list"

These 2 basic options cover all needs.

Why your variant is not good:
1) It adds an option that requires support in the future for the functionality that is already covered by other option.

2) Having a very large exclude list is the same as having a very large include-list.

3) The information which test is excluded and why should be located/versioned in GIT: the place where the test is registered is that place.

ichebyki:
> This option is not needed and here are the reasons why
> 
> 1. You can mark test as "Ignored on Mac" in the place of the registration. The test won't run on Mac in this case.

You need to rebuild the UGENE

> 2. You can use already supported "gui-test-suite" option to create "include-list"

One exclude list vs. 4 testlists. 
What is better?

> 
> These 2 basic options cover all needs.
> 
> Why your variant is not good:
> 
> 1. It adds an option that requires support in the future for the functionality that is already covered by other option.

Don't agree, the old one is inconvenient
The new option adds testing usability

> 2. Having a very large exclude list is the same as having a very large include-list.

Don't agree
Having 4 large testlists to run is 4 times worst

> 3. The information which test is excluded and why should be located/versioned in GIT: the place where the test is registered is that place.

This is not an argument
The same info may be embedded into the exclude list file

----
I strongly disagree at all.
It is very inconvenient, costly and inefficient.
1) To ignore the test, you need to rebuild UGENE, this is unacceptable for me
QA/QE must create pull request and etc. While (s)he can just change exclude list.
2) You can mute the test in teamcity, but the test will still be executed and, as experience shows, it can affect the behavior of the OS and / or subsequent tests (at least on Mac) Also, the time to run is spent for this test
3) Using the testlists for runs is also inconvenient - when adding / excluding / changing tests, you need to keep track of these lists (there are now 4 of them for macos)

At the same time, we, by default, use the option `suite number`
This means that ugene splits all registered tests into a hardcoded number of suites, then runs the one specified in the option. 
We don't have to manually split tests into suites.
For infrequent periodic checks, it is easy to run only tests in an exclude list.
The is a single file, so it's 4 times easier to maintain vs. 4 testlists to run.
I felt all this for myself when I started calculating `bad tests` for macos.

yalgaer:
Closing as duplicating functionality.
In general I agree that using lists may be more flexible than hard-coding ignored tests into the code, but we do not need 2 competing way to do it to avoid the mess in the tests structure.

Please re-open/create another issue if your goal is to replace the current in-code lists with the new external lists.

PR #406, opened 2021-06-12
UGENE-7273. Unable to undo deleting a character in alignment

evelinabis:
[7273](https://ugene.dev/tracker/browse/UGENE-7273)


src/corelibs/U2Formats/src/ClustalWAlnFormat.cpp
```diff
@@ -130,7 +130,7 @@ void ClustalWAlnFormat::load(IOAdapterReader &reader, const U2DbiRef &dbiRef, QL
             valEndPos++;
         }
         if (valEndPos != len) {    //there were numbers trimmed -> trim spaces now
-            while (valEndPos > valStartPos && buf[valEndPos] == ' ') {
+            while (valEndPos > valStartPos && (buf[valEndPos] == ' ' || buf[valEndPos] == '\t')) {
```
DmitriiSukhomlinov, line 133: `buf[valEndPos] == ' '` could be removed, I suppose

DmitriiSukhomlinov (dismissed):

src/corelibs/U2Formats/src/ClustalWAlnFormat.cpp
yalgaer, line 133: Can we use TextUtils::isWhiteSpace() here, same as above?

yalgaer: How valEndPos can grow and include spaces? We use !TextUtils::isWhiteSpace() on the line 129.

yalgaer (changes requested):
Changes in document format readers/writers should be testable with XML tests: GUI tests are overkill for that and only test indirect side-effects, but not the main issue.

Please replace GUI with XML test if the problem with format only.

src/corelibs/U2Formats/src/ClustalWAlnFormat.cpp
evelinabis: > Can we use TextUtils::isWhiteSpace() here, same as above?

No. For newline this method returns `true`, it's wrong.

> How valEndPos can grow and include spaces?

Loop on line 133 is executed 0 or 1 times.

yalgaer:
All is good now. Please re-run the tests and I will approve. Do not forget to merge with master first.

yalgaer (approved):

PR #408, opened 2021-06-17
UGENE-5657 KAlign crashes if the input alignment contains '*'

rasputinkirill:


src/corelibs/U2Lang/src/support/SimpleWorkflowTask.cpp
```diff
@@ -148,6 +149,14 @@ SimpleMSAWorkflow4GObjectTask::SimpleMSAWorkflow4GObjectTask(const QString &task
     MultipleSequenceAlignment al = MSAUtils::createCopyWithIndexedRowNames(msaObjectPointer->getMultipleAlignment());
 
     MultipleSequenceAlignmentObject *msaObject = MultipleSequenceAlignmentImporter::createAlignment(msaObjectPointer->getEntityRef().dbiRef, al, os);
+    for (const MultipleSequenceAlignmentRow &row : al->getMsaRows()) {
+        if (row->getData().contains("*")) {
+            ioLog.info(tr("Symbols '*' has been replaced with 'X' during alignment."));
```
DmitriiSukhomlinov, line 154: I think, this sign's better to be shown just one time, not for every read with the `*` character. Like:

```
bool contains = false;
for(..) {
    if (...contains("*")) {
        contains = true;
    }
}
if (contains) {
    ioLog...
}
```

src/corelibs/U2Lang/src/support/SimpleWorkflowTask.cpp
```diff
@@ -148,6 +149,14 @@ SimpleMSAWorkflow4GObjectTask::SimpleMSAWorkflow4GObjectTask(const QString &task
...
+            break;
+        }
+    }
+    msaObject->replaceAllCharacters('*', 'X');
```
DmitriiSukhomlinov: You should move this lines to the `if` statement above

src/plugins_3rdparty/kalign/src/KalignPlugin.cpp
```diff
@@ -176,9 +177,19 @@ void KalignMSAEditorContext::sl_align() {
     assert(action != NULL);
     MSAEditor *ed = action->getMSAEditor();
     MultipleSequenceAlignmentObject *obj = ed->getMaObject();
-
+
```
DmitriiSukhomlinov: Remove this line

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Lang/src/support/SimpleWorkflowTask.cpp
```diff
@@ -148,6 +149,14 @@ SimpleMSAWorkflow4GObjectTask::SimpleMSAWorkflow4GObjectTask(const QString &task
     MultipleSequenceAlignment al = MSAUtils::createCopyWithIndexedRowNames(msaObjectPointer->getMultipleAlignment());
 
     MultipleSequenceAlignmentObject *msaObject = MultipleSequenceAlignmentImporter::createAlignment(msaObjectPointer->getEntityRef().dbiRef, al, os);
+    for (const MultipleSequenceAlignmentRow &row : al->getMsaRows()) {
```
yalgaer: use qAsConst with for loop and QT collections

src/corelibs/U2Lang/src/support/SimpleWorkflowTask.cpp
```diff
@@ -148,6 +149,14 @@ SimpleMSAWorkflow4GObjectTask::SimpleMSAWorkflow4GObjectTask(const QString &task
...
+        if (row->getData().contains("*")) {
+            msaObject->replaceAllCharacters('*', 'X');
```
yalgaer: Do you replace '*' with 'X' for all algorithms here, not only for KAalign?

yalgaer: 'X' has semantic meaning (unknown amino acid) for protein alignment.
'N' has the same meaning (unknown nucleotide) for DNA alignment.
Why did you select 'X' but not 'N' or something else here?

src/plugins_3rdparty/kalign/src/KalignPlugin.cpp
```diff
@@ -176,9 +177,18 @@ void KalignMSAEditorContext::sl_align() {
     assert(action != NULL);
     MSAEditor *ed = action->getMSAEditor();
     MultipleSequenceAlignmentObject *obj = ed->getMaObject();
-
     KalignTaskSettings s;
     QObjectScopedPointer<KalignDialogController> dlg = new KalignDialogController(ed->getWidget(), obj->getMultipleAlignment(), s);
+    bool containAsterisk = false;
+
+    for (const MultipleSequenceAlignmentRow &row : obj->getMsa()->getMsaRows()) {
```
yalgaer: 1) Use qAsConst with for loop & QT collections
2) Do not use collections returned by value in (see getMsaRows()) in for loop. It creates temporary object with an undefined life range. (qAsConst will help here too)

src/corelibs/U2Lang/src/support/SimpleWorkflowTask.cpp
```diff
@@ -148,6 +149,14 @@ SimpleMSAWorkflow4GObjectTask::SimpleMSAWorkflow4GObjectTask(const QString &task
...
+            ioLog.info(tr("Symbols '*' has been replaced with 'X' during alignment."));
```
yalgaer: Any MSA with '*' char has RAW alphabet. It means that it may also have any other character: braces, &, > < ! etc...
Did you check that KAlign works with all of them, but only fails with '*'?

yalgaer (changes requested):
I have several questions (see comments) and would suggest to fix this issue simpler:
If the alphabet is RAW do not call KAlign at all: tell user to convert it to DNA/RNA/Amino first.

src/corelibs/U2Lang/src/support/SimpleWorkflowTask.cpp
```diff
@@ -148,6 +149,14 @@ SimpleMSAWorkflow4GObjectTask::SimpleMSAWorkflow4GObjectTask(const QString &task
...
+            break;
+        }
+    }
+    U2AlphabetUtils::assignAlphabet(al);
```
DmitriiSukhomlinov: I suppose, it's better to reassign alphabet only if you relapsed some nucleotides

DmitriiSukhomlinov (changes requested):
See a comment

src/plugins_3rdparty/kalign/src/KalignTask.cpp
```diff
@@ -112,6 +111,11 @@ Task::ReportResult KalignTask::report() {
     return ReportResult_Finished;
 }
 
+bool KalignTask::isAlphabetSupported(const DNAAlphabet *alphabet) {
+    return !(alphabet->getId() == BaseDNAAlphabetIds::RAW() ||
```
yalgaer: Since the set of alphabet ids is the open set it is safier to check for supported alphabets ids here.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -2596,6 +2596,17 @@ GUI_TEST_CLASS_DEFINITION(test_5638) {
     CHECK_SET_ERR(finishGapModel == startGapModel, "Unexpected changes of alignment");
 }
 
+GUI_TEST_CLASS_DEFINITION(test_5657) {
+    //1. Open _common_data/clustal/COI_sub_asterisks.aln
+    GTFileDialog::openFile(os, testDir + "_common_data/clustal/COI_sub_asterisks.aln");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    //2. Try to align it with Kalign
+    //Expected state: there is messagebox about incompatible alphabet
+    GTUtilsDialog::waitForDialog(os, new PopupChooser(os, QStringList() << MSAE_MENU_ALIGN << "align_with_kalign"));
```
yalgaer, line 2657: Is it possible to check message box text here too?

yalgaer (commented):
All is good, only minor comments.
Please run the tests and I will approve

src/plugins_3rdparty/kalign/src/KalignTask.cpp
```diff
@@ -112,6 +111,11 @@ Task::ReportResult KalignTask::report() {
     return ReportResult_Finished;
 }
 
+bool KalignTask::isAlphabetSupported(const QString &alphabetId) {
+    return !(alphabetId == BaseDNAAlphabetIds::RAW() ||
```
yalgaer: Could you please change the alphabet check here: check for alphabets KAlign support instead of those KAlign does not support?
If we add a new alphabet tomorrow your check may not be safe.

yalgaer (changes requested):
Please check for supported alphabets, not for the unsupported ones

yalgaer (commented):
Any updates about passed UI/XML tests?

yalgaer (approved):
Good job

DmitriiSukhomlinov (approved):

PR #415, opened 2021-06-23
UGENE-7247. Application unexpectedly closed after saving Workflow properties

evelinabis:
https://ugene.dev/tracker/browse/UGENE-7247


src/corelibs/U2Designer/src/wizard/WizardController.cpp
```diff
@@ -66,11 +66,11 @@
 
 namespace U2 {
 
-WizardController::WizardController(Schema *s, const Wizard *w)
-    : QObject(), schema(s), wizard(w), runAfterApply(false) {
+WizardController::WizardController(std::shared_ptr<Schema> s, const Wizard *w)
```
yalgaer: This code does not compile
I would suggest:
1) Use QT variants of the pointers. They have no cross platform/compiler issues.
2) When use std always use std prefix (see move())
3) For any non-primitive type method parameters (even shared structures) use (const Type&).

yalgaer (changes requested):

src/corelibs/U2Designer/src/wizard/WizardController.cpp
```diff
@@ -66,11 +66,11 @@
...
+    : QObject(), schema(move(s)), wizard(w), runAfterApply(false) {
```
DmitriiSukhomlinov: This thing looks a bit strange (I mean, `std::move`) - a bit over-designed or like you don't get the idea of the move semantic. 

It's difficult to explain why it's not really correct, I've tied to text a good comment, but it was very large and complicated in result. The main idea is: do not use `std::move` in this context, just replace your argument `std::shared_ptr<Schema> s` with `const std::shared_ptr<Schema>& s` (and it's better to use `QSharedPointer`)

src/corelibs/U2Lang/src/support/WorkflowIOTasks.cpp
```diff
@@ -38,9 +38,9 @@ using namespace Workflow;
 /************************************
  * LoadWorkflowTask
  ************************************/
-LoadWorkflowTask::LoadWorkflowTask(Schema *s, Workflow::Metadata *m, const QString &u)
+LoadWorkflowTask::LoadWorkflowTask(std::shared_ptr<Schema> s, Workflow::Metadata *m, const QString &u)
     : Task(tr("Loading workflow"), TaskFlag_None),
-      url(u), schema(s), meta(m) {
+      url(u), schema(move(s)), meta(m) {
```
DmitriiSukhomlinov: The same

src/corelibs/U2Lang/src/tasks/SchemaEstimationTask.cpp
```diff
@@ -40,8 +40,8 @@ EstimationResult::EstimationResult() {
     cpuCount = -1;
 }
 
-SchemaEstimationTask::SchemaEstimationTask(const Schema *_schema, const Metadata *_meta)
-    : Task(tr("Workflow estimation task"), TaskFlag_None), schema(_schema), meta(_meta) {
+SchemaEstimationTask::SchemaEstimationTask(std::shared_ptr<const Schema> _schema, const Metadata *_meta)
+    : Task(tr("Workflow estimation task"), TaskFlag_None), schema(move(_schema)), meta(_meta) {
```
DmitriiSukhomlinov: The same

src/corelibs/U2Lang/src/tasks/SchemaEstimationTask.cpp
```diff
@@ -75,8 +75,8 @@ EstimationResult SchemaEstimationTask::result() const {
 /************************************************************************/
 /* Utils */
 /************************************************************************/
-ExtimationsUtilsClass::ExtimationsUtilsClass(QScriptEngine &_engine, const Schema *_schema)
-    : engine(_engine), schema(_schema) {
+ExtimationsUtilsClass::ExtimationsUtilsClass(QScriptEngine &_engine, std::shared_ptr<const Schema> _schema)
+    : engine(_engine), schema(move(_schema)) {
```
DmitriiSukhomlinov: The same

DmitriiSukhomlinov (changes requested):
Use Qt smart pointers and get rid of the move semantic features - it this context it's excessively

yalgaer (approved):

DmitriiSukhomlinov (approved):

PR #416, opened 2021-06-24
Fix GT_CHECK_RESULT() result in GTUtilsMdi::isTabbedLayout() method

danfe:
`GTUtilsMdi::isTabbedLayout(HI::GUITestOpStatus &os)` returns `bool` while `GT_CHECK_RESULT()` assertion macro is currently being called with `NULL` result which is defined to `nullptr` in modern C++ which is of type `nullptr_t` and cannot initialize return object of type `bool`.
```
src/GTUtilsMdi.cpp:218:63: error: cannot initialize return object of type 'bool'
  with an rvalue of type 'nullptr_t'
    GT_CHECK_RESULT(mainWindow != NULL, "MainWindow == NULL", NULL);
                                                              ^~~~
/usr/include/sys/_null.h:37:14: note: expanded from macro 'NULL'
#define NULL    nullptr
```


yalgaer (approved):

yalgaer:
Thank you!

PR #418, opened 2021-06-25
UGENE-5794 Problem with Sanger mapping if the UGENE data folder is absent

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-5794


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -4147,6 +4147,33 @@ GUI_TEST_CLASS_DEFINITION(test_5790) {
     CHECK_SET_ERR(GTUtilsMcaEditorSequenceArea::getSelectedRect(os) == emptyselection, "Selection isn't empty but should be");
 }
 
+GUI_TEST_CLASS_DEFINITION(test_5794) {
+    // 0. Rename the "data" folder
+    QDir dir(".");
+    auto cp = dir.current();
```
yalgaer: nit: using 'auto' in this case (short trivial type names like QDir/QString) only reduce overall readability.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -4147,6 +4147,33 @@ GUI_TEST_CLASS_DEFINITION(test_5790) {
...
+    // 1. Try to map Sanger reads to reference
+    AlignToReferenceBlastDialogFiller::Settings settings;
+    settings.referenceUrl = testDir + "_common_data/sanger/reference.gb";
+    for (int i = 1; i <= 21; i++) {
+        settings.readUrls << testDir + "_common_data/sanger/" + "sanger_" + QString::number(i) + ".ab1";
+    }
+    settings.outAlignment = sandBoxDir + "test_5794.ugenedb";
+    GTUtilsDialog::waitForDialog(os, new AlignToReferenceBlastDialogFiller(settings, os));
+
+    GTMenu::clickMainMenuItem(os, QStringList() << "Tools"
```
yalgaer: nit: use short list construction.

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
```diff
@@ -96,15 +96,19 @@ AlignToReferenceBlastCmdlineTask::AlignToReferenceBlastCmdlineTask(const Setting
...
 }
 
 void AlignToReferenceBlastCmdlineTask::prepare() {
     AppContext::getAppSettings()->getUserAppsSettings()->createCurrentProcessTemporarySubDir(stateInfo);
-    const bool opened = reportFile.open();
-    SAFE_POINT_EXT(opened, setError(L10N::errorOpeningFileWrite(reportFile.fileName())), );
-    reportFile.close();
+    QList<QTemporaryFile*> tempFiles = { &reportFile, &errorStateFile };
+    for (auto tempFile : tempFiles) {
+        bool opened = tempFile->open();
```
yalgaer: nit: A good candidate for and utility method like: FileAndDirectoryUtils::checkFileIsWritable(QFile&, OpStatus&);

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
```diff
@@ -214,23 +218,33 @@ QList<Task *> AlignToReferenceBlastCmdlineTask::onSubTaskFinished(Task *subTask)
...
+        errorStateFile.close();
+        CHECK_EXT(subprocessError.isEmpty(), setError(subprocessError), result);
 
-        DocumentFormat *format = formats.first().format;
-        CHECK_EXT(format->getSupportedObjectTypes().contains(GObjectTypes::MULTIPLE_CHROMATOGRAM_ALIGNMENT), setError(tr("wrong output format")), result);
+        if (settings.addResultToProject) {
+            // add load document task
+            FormatDetectionConfig config;
+            QList<FormatDetectionResult> formats = DocumentUtils::detectFormat(settings.resultAlignmentFile, config);
+            CHECK_EXT(!formats.isEmpty() && (NULL != formats.first().format), setError(tr("wrong output format")), result);
```
yalgaer, line 240: detectFormat must never return nullptrs in the list: the check is not needed.

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
```diff
@@ -239,6 +253,7 @@ QList<Task *> AlignToReferenceBlastCmdlineTask::onSubTaskFinished(Task *subTask)
 void AlignToReferenceBlastCmdlineTask::run() {
     reportFile.open();
```
yalgaer: check file open result code

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
```diff
@@ -55,44 +55,44 @@ WorkflowRunFromCMDLineBase::WorkflowRunFromCMDLineBase()
...
 }
 
+void WorkflowRunFromCMDLineBase::sl_stateChanged() {
+    CHECK(isFinished(), );
+    CHECK(hasError(), );
+
+    CMDLineRegistry* cmdLineRegistry = AppContext::getCMDLineRegistry();
+    SAFE_POINT(cmdLineRegistry != nullptr, "CMDLineRegistry is nullptr", );
+
+    const QString errorStateFilePath = cmdLineRegistry->getParameterValue(CmdlineTaskRunner::ERROR_STATE_FILE_ARG);
```
yalgaer: nit: const for locals: some locals with no const some with const - inconsistent style within a single PR. I would suggest skipping const for locals to be consisitent with the rest UGENE code.

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
```diff
@@ -55,44 +55,44 @@ WorkflowRunFromCMDLineBase::WorkflowRunFromCMDLineBase()
...
+    CHECK(!errorStateFilePath.isEmpty(), );
+
+    QFile errorStateFile(errorStateFilePath);
+    const bool opened = errorStateFile.open(QIODevice::WriteOnly);
```
yalgaer: Same here

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
```diff
@@ -55,44 +55,44 @@ WorkflowRunFromCMDLineBase::WorkflowRunFromCMDLineBase()
...
+    if (!opened) {
```
yalgaer, line 83: nit: Use CHECK_EXT

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
```diff
@@ -55,44 +55,44 @@ WorkflowRunFromCMDLineBase::WorkflowRunFromCMDLineBase()
...
+
+    errorStateFile.write(getError().toLocal8Bit());
+}
+
 LoadWorkflowTask *WorkflowRunFromCMDLineBase::prepareLoadSchemaTask(const QString &schemaName) {
     QString pathToSchema = WorkflowUtils::findPathToSchemaFile(schemaName);
     if (pathToSchema.isEmpty()) {
-        coreLog.error(tr("Cannot find workflow: %1").arg(schemaName));
-        return NULL;
+        setError(tr("The workflow \"%1\" is absent. Check the file path and the existence of the \"Unipro UGENE/data\" folder.").arg(schemaName));
```
yalgaer: nit: Same here: CHECK_EXT will reduce 4 lines into 1

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
```diff
@@ -55,44 +55,44 @@ WorkflowRunFromCMDLineBase::WorkflowRunFromCMDLineBase()
...
-    int taskOptionIdx = CMDLineRegistryUtils::getParameterIndex(WorkflowDesignerPlugin::RUN_WORKFLOW);
-    if (taskOptionIdx != -1) {
-        processLoadSchemaTask(cmdLineRegistry->getParameterValue(WorkflowDesignerPlugin::RUN_WORKFLOW, taskOptionIdx), taskOptionIdx);
-    }
-    if (loadTask == NULL) {
-        setError(tr("no task to run"));
-        return;
-    }
-    addSubTask(loadTask);
+    connect(this, &Task::si_stateChanged, this, &WorkflowRunFromCMDLineBase::sl_stateChanged);
```
yalgaer, line 58: Is it a bug here: this connects to this? We have a dedicated state methods like prepare/run/report for each state change.

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
```diff
@@ -200,6 +223,11 @@ void WorkflowRunFromCMDLineBase::run() {
     reportFile.write(workflowRunTask->generateReport().toLocal8Bit());
 }
 
+QString WorkflowRunFromCMDLineBase::generateReport() const {
+    int i = 0;
```
yalgaer: remove i=0

src/corelibs/U2Core/src/tasks/CmdlineTaskRunner.cpp
```diff
@@ -104,6 +104,7 @@ bool containsPrefix(const QStringList &args, const QString &prefix) {
 }    // namespace
 
 const QString CmdlineTaskRunner::REPORT_FILE_ARG = "ugene-write-task-report-to-file";
+const QString CmdlineTaskRunner::ERROR_STATE_FILE_ARG = "ugene-write-task-error-state-to-file";
```
yalgaer, line 107: Why stderr is not enough in our case?
Using stderr + exit code is a standard way how command line tools report that something went wrong

yalgaer (changes requested):

src/corelibs/U2Core/src/tasks/CmdlineTaskRunner.cpp
DmitriiSukhomlinov: The process really could return you an exit code, but just the number - no details. Of course, you can have a list of errors for each return code, but I don't see any advantages of this method. I use the infrastructure which has already been used for Sanger reports - use the temporary file to save the output of the process, this solution has been introduced by @atiunov, and I like it more that your idea.

Anyway, I won't change the good solution of this issue (or any other issue) just because "you don't like it" or "you have your own idea how it should be done" - take it in account in future.

yalgaer: > Anyway, I won't change the good solution of this issue (or any other issue) just because "you don't like it" or "you have your own idea how it should be done" - take it in account in future.

It was only a question why the standard command-line solution is not used here and you selected a more complex and advanced non-standard variant. I would accept your solution if you have a reason for it.

Since I do not see the answer for the problem and only 'take it in account' statement above I'm closing this PR and re-assigning the bug to myself.

DmitriiSukhomlinov: I used this solution because it's been already used for [the similar issue](https://ugene.dev/tracker/browse/UGENE-5591). Your solution is new and the new solution is more difficult to implement that the solution which has been already used (you have said that "I selected a more complex and advanced non-standard variant", but it's not true).

It was no certain solution had to be implemented (nothing about it in the issue) and, believe it or not, I can't read your minds. I spent a day to this bug fix - it works and it works good - and I'm not about to re-make it.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
DmitriiSukhomlinov: This variable is unused, removed

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
DmitriiSukhomlinov: OK

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
DmitriiSukhomlinov: OK

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
DmitriiSukhomlinov: This is not my code, but OK, removed

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
DmitriiSukhomlinov: OK

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
DmitriiSukhomlinov: OK

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
DmitriiSukhomlinov: OK

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
DmitriiSukhomlinov: I reduced 4 lines to 3, cuz I don't like when some line is too long (it's not convenient if you use split screen)

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
DmitriiSukhomlinov: It's not a bug. the `report` function is for the report generation, not for some post-processing stuff. It's apples and oranges, no need of mixing them. Also, we use this way of connection very often, look, for example, `BaseOneOneWorker` (54), `BaseOneOneWorker` (62), `WorkflowMonitor`(70) and a lot of (actually, A LOT OF) other similar places (search for using `si_stateChanged`).

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
DmitriiSukhomlinov: `generateReport()` has been removed, it's been using for some debug stuff

DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp
```diff
@@ -194,4 +195,13 @@ bool FileAndDirectoryUtils::canWriteToPath(const QString &absoluteDirPath) {
     return true;
 }
 
+bool FileAndDirectoryUtils::checkFileIsWritable(U2OpStatus& os, QFile* file) {
```
yalgaer, line 198: nit: function name is "checkIsWritable" but you check both Read & Write. Change either name of the function or open-flags

src/corelibs/U2Core/src/util/FileAndDirectoryUtils.h
```diff
@@ -52,6 +53,8 @@ class U2CORE_EXPORT FileAndDirectoryUtils {
      */
     static bool canWriteToPath(const QString &absoluteDirPath);
 
+    static bool checkFileIsWritable(U2OpStatus& os, QFile* file);
```
yalgaer, line 56: 1. Add minimal docs for public API.
2. QFile& is more preferable than QFile* because in most cases we do not use "new File", but only create it on stack or on class instance (in some local scope).

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -4147,6 +4147,30 @@ GUI_TEST_CLASS_DEFINITION(test_5790) {
     CHECK_SET_ERR(GTUtilsMcaEditorSequenceArea::getSelectedRect(os) == emptyselection, "Selection isn't empty but should be");
 }
 
+GUI_TEST_CLASS_DEFINITION(test_5794) {
+    // 0. Rename the "data" folder
+    QDir dir(".");
+    auto tempDataDir = dataDir.left(dataDir.size() - 1) + "_temp";
```
yalgaer, line 4153: Question: why do you want to remove the last character in dataDir name?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -4147,6 +4147,30 @@ GUI_TEST_CLASS_DEFINITION(test_5790) {
...
+    settings.referenceUrl = testDir + "_common_data/sanger/reference.gb";
+    for (int i = 1; i <= 21; i++) {
+        settings.readUrls << testDir + "_common_data/sanger/" + "sanger_" + QString::number(i) + ".ab1";
+    }
+    settings.outAlignment = sandBoxDir + "test_5794.ugenedb";
+    GTUtilsDialog::waitForDialog(os, new AlignToReferenceBlastDialogFiller(settings, os));
+
+    GTMenu::clickMainMenuItem(os, { "Tools", "Sanger data analysis", "Map reads to reference..." });
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    dir.rename(tempDataDir, dataDir);
```
yalgaer, line 4167: nit: if the test fails before this line data-dir will not be renamed back and all tests run after this one will fail.
This is safe only when we use 'full-test-and-data-dir-restore' script after each test.

Question: is it possible to change location of the data for UGENE, but do not change the data dir itself?

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
```diff
@@ -96,15 +97,18 @@ AlignToReferenceBlastCmdlineTask::AlignToReferenceBlastCmdlineTask(const Setting
...
+      errorStateFile(AppContext::getAppSettings()->getUserAppsSettings()->getCurrentProcessTemporaryDirPath() + "/align_to_ref_error_state_XXXXXX.txt") {
     GCOUNTER(cvar, "AlignToReferenceBlastCmdlineTask");
 }
 
 void AlignToReferenceBlastCmdlineTask::prepare() {
     AppContext::getAppSettings()->getUserAppsSettings()->createCurrentProcessTemporarySubDir(stateInfo);
-    const bool opened = reportFile.open();
-    SAFE_POINT_EXT(opened, setError(L10N::errorOpeningFileWrite(reportFile.fileName())), );
-    reportFile.close();
+    QList<QTemporaryFile*> tempFiles = { &reportFile, &errorStateFile };
```
yalgaer, line 107: no need in loop here: calling this code twice will takes the same number of lines (4) and does require to use QFile* in API.
```
        FileAndDirectoryUtils::checkFileIsWritable(stateInfo, tempFile);
        CHECK_OP(stateInfo, );
```

src/plugins/external_tool_support/src/blast_plus/AlignToReferenceBlastDialog.cpp
```diff
@@ -214,31 +218,44 @@ QList<Task *> AlignToReferenceBlastCmdlineTask::onSubTaskFinished(Task *subTask)
...
-    } else if (subTask == cmdlineTask && settings.addResultToProject) {
-        // add load document task
-        FormatDetectionConfig config;
-        QList<FormatDetectionResult> formats = DocumentUtils::detectFormat(settings.resultAlignmentFile, config);
-        CHECK_EXT(!formats.isEmpty() && (NULL != formats.first().format), setError(tr("wrong output format")), result);
+    } else if (subTask == cmdlineTask) {
+        bool opened = errorStateFile.open();
+        SAFE_POINT_EXT(opened, setError(L10N::errorOpeningFileWrite(errorStateFile.fileName())), result);
+
+        QString subprocessError = errorStateFile.readAll();
```
yalgaer, line 232: You refer to the original code as a good example how to deal with process results: the original code reads the report file in a separate thread.  Could you do the same for the error file: do not read/block scheduler thread and move it to the 'run' method?

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
yalgaer: > actually, A LOT OF

Could you please point some places where a task uses si_stateChanged signal to get notification about its own state updates?
I checked your examples and they look like this:
```
connect(lastTickTask, SIGNAL(si_stateChanged()), SLOT(sl_taskFinished()));
```
Note: that the sender and the recipient are different objects.
In your case you have 2 'this' in the connect line.

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
yalgaer: Do not see any changes

src/corelibs/U2Core/src/tasks/CmdlineTaskRunner.cpp
yalgaer: OK, I can accept your solution here if you insist.

Just a couple of questions to mark it resolved:

1) Do you think that "I spent a day to this bug fix - it works and it works good - and I'm not about to re-make it." is a good technical reason may be used to compare 2 different options? 

2)  You say "I can't read your minds", but I can't neither: I can't give my opinion about the solution before you present it. And once you present it you consider it as 'no discussion is possible'. How this situation can be avoided in the future?

3) If you have a command line utility, which behavior do you expect from it on error:
 A) Return non-0 (error) exit code and error message into the error output stream.
 B) Return 0 (success) exit code and only report an error if there is a special 'error-output-file' parameter.

That is all I have to say about current discussion. I will mark this thread resolved and approve your variant. Please address all other minor issues.

yalgaer (changes requested):
Added comments, resolved fixed

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
yalgaer: Agree

src/plugins/workflow_designer/src/cmdline/WorkflowCMDLineTasks.cpp
DmitriiSukhomlinov: OK, you're right, your suggestions?

DmitriiSukhomlinov:
I got tired of this, PR is closed, the issue is yours

PR #426, opened 2021-06-27
UGENE-4799. Impossible to add sequence with different alphabet to ali…

yalgaer:


DmitriiSukhomlinov (changes requested):
There are a lot of changes, which are not connected to the issue at all. Please, split them to the separate commit.

yalgaer:
> There are a lot of changes, which are not connected to the issue at all. Please, split them to the separate commit.

Looks like it was incomplete merge with 'master'. Re-merged.

PR #442, opened 2021-06-29
UGENE-7328. Show group size in MSA Editor

yalgaer:


src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
```diff
@@ -2235,10 +2235,7 @@ void GUITestBasePlugin::registerTests(UGUITestBase *guiTestBase) {
     REGISTER_TEST(GUITest_common_scenarios_msa_editor::test_0017_2);
 
     REGISTER_TEST(GUITest_common_scenarios_msa_editor::test_0018);
-
     REGISTER_TEST(GUITest_common_scenarios_msa_editor::test_0019);
-    REGISTER_TEST(GUITest_common_scenarios_msa_editor::test_0019_1);
```
DmitriiSukhomlinov, line 2240: Why did you remove these tests?

DmitriiSukhomlinov (approved):

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
yalgaer: All 3 tests had 100% identical code. Looks like the original author copy/pasted them to use as a template for new tests, but have never finished it.

PR #445, opened 2021-06-30
UGENE-7327. Sort MSA sequence groups by size

yalgaer:
I moved collapse model related from MSAEditorSequenceArea to the top level (MSA Editor): all TODOs in the diff are from the old code - no new TODOs were added.


src/corelibs/U2View/src/ov_msa/MSAEditor.cpp
```diff
@@ -788,4 +810,106 @@ void MSAEditor::sl_convertRawToAminoAlphabet() {
     msaObject->morphAlphabet(resultAlphabet);
 }
 
+void MSAEditor::sl_sortGroupsBySize() {
+    groupsSortOrder = sender() == sortGroupsBySizeAscendingAction ? GroupsSortOrder::Ascending : GroupsSortOrder::Descending;
+    updateCollapseModel();
+}
+
+// TODO: move this function into MSA?
```
DmitriiSukhomlinov, line 818: I understand, that this TODO wasn't added by you, but could you either explain the idea or just remove it? "move this function into MSA", but we are already in MSA, isnt it? Or what is this supposed to mean?

src/corelibs/U2View/src/ov_msa/MSAEditor.h
```diff
@@ -102,6 +114,17 @@ class U2VIEW_EXPORT MSAEditor : public MaEditor {
...
     void sortSequences(const MultipleAlignment::SortType &sortType, const MultipleAlignment::Order &sortOrder);
 
+    /** Forces complete re-computation of the active collapse model based on the current MSA editor state. */
+    void updateCollapseModel();
+
+    void setRowOrderMode(MaEditorRowOrderMode mode) override;
+
+    /**
+     * Returns modifiable instance of the 'freeModeMasterMarkersSet'.
+     * TODO: change signature to return const reference.
```
DmitriiSukhomlinov: Where did you get this `TODO` from?

DmitriiSukhomlinov (dismissed):

src/corelibs/U2View/src/ov_msa/MSAEditor.cpp
yalgaer: I fixed this TODO and moved the method into MultipleAlignmentRowData

src/corelibs/U2View/src/ov_msa/MSAEditor.h
yalgaer: I added it after reading existing code that works with freeModeMasterMarkerSet - this set should be protected in some way and not public.
I fixed the issue now and removed the TODO.

PR #446, opened 2021-06-30
Fix for clang warnings

yalgaer:
See Teamcity -> Code Analysis -> Clang warnings


src/plugins_3rdparty/sitecon/src/SiteconAlgorithm.cpp
```diff
@@ -88,29 +86,29 @@ QVector<PositionStats> SiteconAlgorithm::calculateDispersionAndAverage(const Mul
                 dispersion+=(average - v)*(average - v);
             }
             dispersion/=(N-1);
-            float sdeviation= sqrt(dispersion);
+            qreal sdeviation= sqrt(dispersion);
 
-            posResult.append(DiStat(p, sdeviation, average));
+            posResult.append(DiStat(p, (float)sdeviation, (float)average));
```
DmitriiSukhomlinov: You replaced `float` with `qreal`, but here you added an implicit type conversion to `float`. Could you choose one of them?

DmitriiSukhomlinov (dismissed):

src/plugins_3rdparty/sitecon/src/SiteconAlgorithm.cpp
yalgaer: I did the change you asked and it lead to the whole SiteconAlgorithm and sitecon plugin update to avoid conversions in other places. This plugin was also not formatted properly for some reason - I reformatted it too.

PR #453, opened 2021-07-07
UGENE-7059. Help button in Exporting Alignment without Chromatograms …

yalgaer:


DmitriiSukhomlinov (approved):

DmitriiSukhomlinov:
I think, one-line commits like this could be merged without PR

PR #457, opened 2021-07-09
UGENE-6869. The annotation symbol isn't selected in the 'Translate selection' mode

yalgaer:


smlnkcv (approved):

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewSingleLineRenderer.cpp
```diff
@@ -564,27 +564,40 @@ void DetViewSingleLineRenderer::drawSequenceSelection(QPainter &p, const QSize &
...
-                    highlight(p, U2Region(r.endPos() - translLen, translLen), complTransLine, canvasSize, visibleRange);
+                int complTransLine = posToComplTransLine(selectedRegion.endPos());
+                if (complTransLine >= 0 && visibleSelectedRegion.length >= 3) {
+                    qint64 frameEnd = visibleSelectedRegion.endPos();
+                    qint64 frameLength = visibleSelectedRegion.length;
+                    if (selectedRegion.startPos == visibleSelectedRegion.startPos) {
+                        frameLength -= (selectedRegion.endPos() - visibleSelectedRegion.startPos) % 3;
+                    }
+                    U2Region frameRegion(frameEnd - frameLength, frameLength);
+                    highlight(p, frameRegion, complTransLine, canvasSize, visibleRange);
```
DmitriiSukhomlinov, line 600: This part of code is VERY similar to the code above - the separate function begs here. I suppose, lambda-function is the best solution

DmitriiSukhomlinov (changes requested):
See comment

src/corelibs/U2View/src/ov_sequence/view_rendering/DetViewSingleLineRenderer.cpp
yalgaer: The code computes frame in the opposite direction. While half of lines are really the same, another half is VERY different.

DmitriiSukhomlinov (approved):

PR #460, opened 2021-07-13
UGENE-7342. Move MaEditorSelection management into a separate controller out of MaEditorSequenceArea

yalgaer:
No behavior changes are expected as the result of this change (API changes only).

### DIFF
--- a/src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSelection.h
+++ b/src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSelection.h
@@ -28,6 +28,9 @@
 
 namespace U2 {
 
+class MaEditor;
+class McaEditor;
+
 /** Data model for selection in MSA editor: list of visual non-overlapping rectangles. */
 class U2VIEW_EXPORT MaEditorSelection {
 public:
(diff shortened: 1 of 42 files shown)


src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSelection.h
```diff
@@ -67,6 +73,55 @@ class U2VIEW_EXPORT MaEditorSelection {
...
+    /** Signal emitted every time selection is changed. */
+    void si_selectionChanged(const MaEditorSelection &current, const MaEditorSelection &prev);
+
+public slots:
+    /** Sets selection to empty selection. Emits signal that selection is changed. */
+    virtual void clearSelection();
+
+protected:
+    /** MSA/MCA editor instance. Never null. */
+    MaEditor *const editor;
```
DmitriiSukhomlinov: The fact that `McaEditorSelectionController` has two pointers to the same editor looks a bit strange. Maybe, it's better to make this one private?

DmitriiSukhomlinov (changes requested):
See a comment

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSelection.h
yalgaer: Good idea. Done

PR #467, opened 2021-07-18
UGENE-7351. Add unicode support to Newick document format

yalgaer:


src/corelibs/U2Core/src/gobjects/PhyTreeObject.cpp
```diff
@@ -67,9 +67,9 @@ PhyTreeObject *PhyTreeObject::createInstance(const PhyTree &tree, const QString
 }
 
 void PhyTreeObject::commit(const PhyTree &tree, const U2EntityRef &treeRef, U2OpStatus &os) {
-    CHECK_EXT(nullptr != tree.data(), os.setError("NULL tree data"), );
-    QByteArray data = NewickPhyTreeSerializer::serialize(tree);
-    RawDataUdrSchema::writeContent(data, treeRef, os);
+    CHECK_EXT(tree.data() != nullptr, os.setError("NULL tree data"), );
+    QString data = NewickPhyTreeSerializer::serialize(tree);
+    RawDataUdrSchema::writeContent(data.toLocal8Bit(), treeRef, os);
```
DmitriiSukhomlinov: `NewickPhyTreeSerializer::serialize` returns `QByteArray` there's an implicit conversation to `QString` now, and then then you convert `QString` to `QByteArray` again explicitly (`data.toLocal8Bit()`). Replace it with the previous version

src/corelibs/U2Core/src/util/DatatypeSerializeUtils.cpp
```diff
@@ -38,7 +40,7 @@ const QString WMatrixSerializer::ID = "wm_1.14";
 const QString FMatrixSerializer::ID = "fm_1.14";
 
 #define CHECK_SIZE(size, result) \
-    if (offset + size > length) { \
+    if (offset + (size) > length) { \
```
DmitriiSukhomlinov, line 43: Why do we need these bracers?

DmitriiSukhomlinov (changes requested):
See comment

src/corelibs/U2Core/src/gobjects/PhyTreeObject.cpp
yalgaer: Didn't you notice this part of the patch: 
+ QString NewickPhyTreeSerializer::serialize(const PhyTree &tree)

I changed the tree serializer to produce a text but not QByteArray

src/corelibs/U2Core/src/util/DatatypeSerializeUtils.cpp
yalgaer: Arithmetic op on macro param

DmitriiSukhomlinov (dismissed):

tests/doc_format/newick/write/test_0008.xml
```diff
@@ -0,0 +1,12 @@
+<multi-test>
+
+    <!-- Check that a document with a unicode content can be read and written correctly.  -->
+    <load-document index="doc" url="newick/sample_utf16be.newick" io="local_file" format="newick"/>
```
evelinabis, line 4: Shouldn't this file be at github.com/ugeneunipro/ugene/tree/master/tests?

src/corelibs/U2Formats/src/NewickFormat.cpp
```diff
@@ -32,70 +33,50 @@
...
-        if ('\'' == data[i]) {
-            if (!quotedLabelStarted && i > 0 && (data[i - 1] == '(' || data[i - 1] == ',')) {
-                quotedLabelStarted = true;
-            } else if (quotedLabelStarted) {
-                quotedLabelStarted = false;
-            }
+    for (int i = 0; i < dataPrefix.size(); ++i) {
+        char ch = dataPrefix[i].toLatin1();
+        if (ch == '\'') {
+            char prevCh = dataPrefix[i - 1].toLatin1();
```
evelinabis: You removed check that `i` is greater than zero.

evelinabis (commented):
Newick file isn't detected by default, file preview doesn't work well (this problem isn't this issue).

src/corelibs/U2Formats/src/NewickFormat.cpp
yalgaer: This was good. Added a dedicated test that will not allow to happen this again.

tests/doc_format/newick/write/test_0008.xml
yalgaer: Do you mean data file? We keep them separate from the XML tests descriptors today in a dedicated SVN repo because the total data size is about 2-3Gb.

src/corelibs/U2Core/src/util/DatatypeSerializeUtils.cpp
```diff
@@ -203,36 +205,40 @@ enum ReadState { RS_NAME,
...
-    if (branches == 1 && (node->getName() == "" || node->getName() == "ROOT")) {
-        assert(node != node->getSecondNodeOfBranch(0));
-        packTreeNode(binary, node->getSecondNodeOfBranch(0));
+void packTreeNode(QString &resultText, const PhyNode *node, U2OpStatus &os) {
+    int branchCount = node->branchCount();
+    const QString &nodeName = node->getName();
+    if (branchCount == 1 && (nodeName.isEmpty() || nodeName == "ROOT")) {
+        const PhyNode *sibling = node->getSecondNodeOfBranch(0);
+        CHECK_EXT(node != sibling, os.setError(DatatypeSerializers::tr("Invalid tree topology")), );
+        packTreeNode(resultText, node->getSecondNodeOfBranch(0), os);
```
evelinabis: Probably you wanted to write here instead of `node->getSecondNodeOfBranch(0)` `sibling`.

src/corelibs/U2Core/src/util/DatatypeSerializeUtils.cpp
```diff
@@ -321,17 +328,17 @@ QList<PhyTree> NewickPhyTreeSerializer::parseTrees(IOAdapter *io, U2OpStatus &si
...
+            // Advance in state.
+            if (ch == '(') {    // A new child.
                 CHECK_EXT_BREAK(!nodeStack.isEmpty(), si.setError(DatatypeSerializers::tr("Tree node stack is empty")));
                 PhyNode *pn = new PhyNode();
                 PhyBranch *bd = PhyTreeData::addBranch(nodeStack.top(), pn, 0);
                 nodeStack.push(pn);
                 branchStack.push(bd);
                 state = RS_NAME;
-            } else if (c == ':') {    //weight start
+            } else if (latin1CharCode == ':') {    // Weight start.
```
evelinabis: Everywhere you compared `ch`, but here `latin1CharCode`.

evelinabis (commented):
"Select Correct Document Format" dialog still pops up when opening trees. Please, add GUI test that opens newick/quoted-1-node.newick.

src/corelibs/U2Core/src/util/DatatypeSerializeUtils.cpp
yalgaer: Right 'latin1CharCode' is used only as an index.

src/corelibs/U2Core/src/util/DatatypeSerializeUtils.cpp
yalgaer: Correct. It will be better this way

yalgaer:
> "Select Correct Document Format" dialog still pops up when opening trees. Please, add GUI test that opens newick/quoted-1-node.newick.

There are 2 ways to open files in UGENE:
1) File->Open: will automatically use the best found format
2) File->OpenAs: will show format detection scores and ask a user to make the selection.

Looks like you described the expected behavior for 2) and there is no issue here. I checked 1) too and it opens 'newick/quoted-1-node.newick' with no dialogs.

In general, any change in document formats must be testable by XML tests, otherwise it is a separate UX issue, that should to be filed/described separately.

EvelinaBiserova:
Now the dialog doesn't appear, ok.
`NewickFormat::checkRawTextData` is called only through GUI tests, test_0005 doesn't check it.

evelinabis (commented):
Now the dialog doesn't appear, ok.
`NewickFormat::checkRawTextData` is called only through GUI tests, test_0005 doesn't check it.

yalgaer:
> Now the dialog doesn't appear, ok.
> `NewickFormat::checkRawTextData` is called only through GUI tests, test_0005 doesn't check it.

This is correct, but it means that we should add an XML instruction to check this similar to <load-document>. Like <check-format-detection-score>...
I will do it later as a separate feature (there are still ~20 formats to migrate) or you can contribute it by yourself.

PR #470, opened 2021-07-19
UGENE-7355. Support multi-selection on MaEditor sequence area

yalgaer:
Selection manipulation in Sequence Area. Please check bug for more details.


src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.cpp
```diff
@@ -876,7 +876,12 @@ void MaEditorSequenceArea::mousePressEvent(QMouseEvent *e) {
             double baseWidth = ui->getBaseWidthController()->getBaseWidth();
             double baseHeight = ui->getRowHeightController()->getSingleRowHeight();
             const MaEditorSelection &selection = editor->getSelection();
-            movableBorder = SelectionModificationHelper::getMovableSide(shape, globalMousePosition, selection.toRect(), QSize(baseWidth, baseHeight));
+            movableBorder = SelectionModificationHelper::NoMovableBorder;
+            // Selection resizing is supported only in single selection mode.
```
DmitriiSukhomlinov, line 880: Do we need an issue to this? Or there's no plans to change this behavior?

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.cpp
```diff
@@ -876,7 +876,12 @@ void MaEditorSequenceArea::mousePressEvent(QMouseEvent *e) {
...
+            if (selection.isSingleSelection()) {
+                const QRect &selectionRect = selection.getRectList().first();
+                movableBorder = SelectionModificationHelper::getMovableSide(shape, globalMousePosition, selectionRect, QSize(baseWidth, baseHeight));
+            }
             moveBorder(pos);
```
DmitriiSukhomlinov: Shouldn't this line be moved to the `if` statement above?

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.cpp
```diff
@@ -989,21 +994,26 @@ void MaEditorSequenceArea::mouseMoveEvent(QMouseEvent *event) {
...
 
 void MaEditorSequenceArea::setBorderCursor(const QPoint &p) {
-    const QPoint globalMousePos = ui->getScrollController()->getGlobalMousePosition(p);
+    QPoint globalMousePos = ui->getScrollController()->getGlobalMousePosition(p);
     const MaEditorSelection &selection = editor->getSelection();
-    setCursor(SelectionModificationHelper::getCursorShape(globalMousePos, selection.toRect(), ui->getBaseWidthController()->getBaseWidth(), ui->getRowHeightController()->getSingleRowHeight()));
+    int viewWidth = ui->getBaseWidthController()->getBaseWidth();
+    int viewHeight = ui->getRowHeightController()->getSingleRowHeight();
+    // Only single selection support resizing.
+    QRect resizableRect = selection.isSingleSelection() ? selection.getRectList().first() : QRect();
```
DmitriiSukhomlinov: Maybe just:
```
CHECK(selection.isSingleSelection(), );
QRect resizableRect = selection.getRectList().first();
```

src/corelibs/U2View/src/ov_msa/view_rendering/SequenceAreaRenderer.cpp
```diff
@@ -78,36 +78,43 @@ bool SequenceAreaRenderer::drawContent(QPainter &painter, const U2Region &column
...
+    const QList<QRect> selectionRects = selection.getRectList();
+    for (const QRect &selectionRect : qAsConst(selectionRects)) {
+        QRect screenRect = ui->getDrawHelper()->getScreenRect(selectionRect);
+        int viewWidth = ui->getSequenceArea()->width();
+        if (screenRect.right() < 0 || screenRect.left() > viewWidth) {
+            continue;    // Selection is out of the screen.
+        }
+
+        // Check that frame has enough space to be drawn on both sides.
+        if (screenRect.left() >= 0 && screenRect.left() < SELECTION_STROKE_WIDTH) {
```
DmitriiSukhomlinov, line 101: I don't like the way you text double comparasings (and comaprasings at all). You have both "<" and ">" signs, but, from my viewpoint, its better always use "<", because the abscissa increases from left to right and it's easier to imagine the mutual arrangement of variable's values if they increas from left to right.

I won't insist on this, but I will use my way

DmitriiSukhomlinov (dismissed):

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.cpp
yalgaer: I documented selection.toRect() method as deprecated and to be removed before the release. 

There are > 10 places that still use bounding box but not the real multi-selection. I'm going to address them all 1 by 1 and create bugs for each place.

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.cpp
yalgaer: moveBorder is safe to be called within 'non-movable' context. But you are right it can be moved inside 'if' too.  Fixed.

src/corelibs/U2View/src/ov_msa/view_rendering/MaEditorSequenceArea.cpp
yalgaer: Yes, this will be more readable. Fixed.

src/corelibs/U2View/src/ov_msa/view_rendering/SequenceAreaRenderer.cpp
yalgaer: I see,
0 < x && x < 10  is also right, but looks like a Yoda style and is not the way we speak in natural language.

I'm trying to write code that can be read in natural language in a straightforward way without any mental transformation involved.

PR #471, opened 2021-07-19
UGENE-6304. Rewrite MSF parser

yalgaer:
The reason for rewrite is a fix for UGENE-6304.

The old code is 
1) complex & buggy 
2) incorrect the way it makes assumptions about sequence reordering: it supposed to allow reordering but only stable reordering. According to the format this support is not needed at all.

The new code is very trivial and should not allow broken MSF to be parsed at all. This is good because we should not make broken data to look like a valid one.

Example of MSF format:
http://rothlab.ucdavis.edu/genhelp/chapter_2_using_sequences.html#_Using_Multiple_Sequence_Format_(MSF

### DIFF
--- a/src/corelibs/U2Formats/src/MSFFormat.cpp
+++ b/src/corelibs/U2Formats/src/MSFFormat.cpp
@@ -99,11 +99,13 @@ struct MsfRow {
 void MSFFormat::load(IOAdapterReader &reader, const U2DbiRef &dbiRef, QList<GObject *> &objects, const QVariantMap &hints, U2OpStatus &os) {
     QString objName = reader.getURL().baseFileName();
     MultipleSequenceAlignment al(objName);
+    int lineNumber = 0;    // Current line number from the object start. Used for error reporing.
 
     // Skip comments.
     int checkSum = -1;
     while (!os.isCoR() && checkSum < 0 && !reader.atEnd()) {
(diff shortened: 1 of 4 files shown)


src/corelibs/U2Formats/src/MSFFormat.cpp
```diff
@@ -160,64 +154,39 @@ void MSFFormat::load(IOAdapterReader &reader, const U2DbiRef &dbiRef, QList<GObj
...
     }
 
-    // read data
-    QRegExp coordsRegexp("^\\s+\\d+(\\s+\\d+)?\\s*$");
-    QRegExp onlySpacesRegexp("^\\s+$");
-
-    QMap<QString, int> processedDuplicatedNames;
-
+    // Read sequences.
+    QRegExp coordsRegexp("^\\d+(\\s+\\d+)?$");
```
DmitriiSukhomlinov, line 158: `(\\s+\\d+)?` - this means, that the second number either exists or not, but I haven't found an example of the "MSF" file where coordinate line has only one number. Maybe the following regexp will be a bit simpler : `^\\d+\\s+\\d+$`

yalgaer: We have an example like this, check test7.msf (check the very end of the file)
I do not know where it came from but it looks pretty real.

PR #479, opened 2021-07-25
UGENE-7348. Create notification report

evelinabis:
What if backbone is empty?

### DIFF
--- a/src/plugins/pcr_primer_design_for_dna_assembly/src/tasks/PCRPrimerDesignForDNAAssemblyTask.cpp
+++ b/src/plugins/pcr_primer_design_for_dna_assembly/src/tasks/PCRPrimerDesignForDNAAssemblyTask.cpp
@@ -189,8 +189,55 @@ QList<Task*> PCRPrimerDesignForDNAAssemblyTask::onSubTaskFinished(Task* subTask)
 }
 
 QString PCRPrimerDesignForDNAAssemblyTask::generateReport() const {
-    //TODO - report
-    return QString();
+    QString report("<style type=\"text/css\">"
+                   "p{font-size: 7pt;}"
+                   ".seq{font-size: 8pt;}"
+                   "</style>"
+                   "<br>"
+                   "<br>"
+                   "<h3>%1</h3>");
+    SAFE_POINT(!sequence.isEmpty(), tr("Empty sequence"), report.arg(tr("Error, see log.")))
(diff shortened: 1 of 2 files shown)


DmitriiSukhomlinov:
Screen Shot 2021-07-26 at 3 59 32 PM

There is a too smal font on macOS, please, increase the font size

DmitriiSukhomlinov (changes requested):
Everything is good, but fix the font size

DmitriiSukhomlinov:
@yalgaer please review

evelinabis:
I will increase font size, but I want to note that on Windows the picture is different, and font I configured looks larger. image

DmitriiSukhomlinov:
> I will increase font size, but I want to note that on Windows the picture is different, and font I configured looks larger. 

I don't think it's an unsolvable problem, look, for example, at the Sanger report for Windows:
image

and for macOS:
Screen Shot 2021-07-26 at 5 42 53 PM

Font sizes are pretty the same on both pictures.

src/plugins/pcr_primer_design_for_dna_assembly/src/tasks/PCRPrimerDesignForDNAAssemblyTask.cpp
```diff
@@ -189,8 +189,51 @@ QList<Task*> PCRPrimerDesignForDNAAssemblyTask::onSubTaskFinished(Task* subTask)
...
+    QString report("<br>"
+                   "<br>"
+                   "<h3>%1</h3>");
+    SAFE_POINT(!sequence.isEmpty(), tr("Empty sequence"), report.arg(tr("Error, see log.")))
+    if (aForward.isEmpty() && aReverse.isEmpty() && b1Forward.isEmpty() && b1Reverse.isEmpty() && b2Forward.isEmpty() &&
+        b2Reverse.isEmpty() && b3Forward.isEmpty() && b3Reverse.isEmpty()) {
+        return report.arg(tr("There are no primers that meet the specified parameters."));
+    }
+
+    auto getPairReport = [&](U2Region forward, U2Region reverse, const QString &primerName) -> QString {
```
DmitriiSukhomlinov: I think, it's better to create a separate function - this code part is too large for lambda

DmitriiSukhomlinov (changes requested):
Minor comment

DmitriiSukhomlinov (approved):

PR #491, opened 2021-08-02
UGENE-6797. Welcome page should not show links to non-available file or projects

yalgaer:


src/corelibs/U2Core/src/models/ProjectModel.h
```diff
@@ -66,6 +66,12 @@ class U2OpStatus;
 /// Service responsible for project loading / unloading
 class U2CORE_EXPORT ProjectLoader : public QObject {
 public:
+    /**
+     * Runs openWithProjectTask for the recent files/projects list item.
+     * If the the with the given URL does not exist asks if to remove if from the recent files list.
```
DmitriiSukhomlinov: If **the the with the** ...
if to remove **if** from

PR #492, opened 2021-08-03
UGENE-7367. Optimize 'Tools->Generate sequence' tool

yalgaer:
Please see the original bug for the details: https://ugene.dev/tracker/browse/UGENE-7367

The fix contains  FASTA write speed optimization (20x boost), random sequence generator optimization (5x boost), improved task progress visualization, performance counters (timers) and the first GUI tests for the feature that tests all these features at once.


src/corelibs/U2Formats/src/FastaFormat.cpp
```diff
@@ -321,15 +321,20 @@ static void saveSequenceObject(IOAdapterWriter &writer, const U2SequenceObject *
     writeHeaderToFile(writer, sequence->getSequenceName(), os);
     CHECK_OP(os, );
     qint64 sequenceLength = sequence->getSequenceLength();
-    for (int i = 0; i < sequenceLength; i += FastaFormat::FASTA_SEQUENCE_LINE_LENGTH) {
-        qint64 chunkSize = qMin((qint64)FastaFormat::FASTA_SEQUENCE_LINE_LENGTH, sequenceLength - i);
+    // Reading DBI line-by-line is very expensive. Read by bigger chunks and split them into lines in memory.
+    qint64 maxChunkSize = 1000 * FastaFormat::FASTA_SEQUENCE_LINE_LENGTH;
```
DmitriiSukhomlinov, line 325: What `1000` is?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -585,6 +586,55 @@ GUI_TEST_CLASS_DEFINITION(test_7293) {
...
+GUI_TEST_CLASS_DEFINITION(test_7367) {
+    // Generate a large sequence.
+    // Check that test does not time-outs and the generated sequence contains expected base distribution.
+
+    DNASequenceGeneratorDialogFillerModel model(sandBoxDir + "/test_7370.fa");
+    model.percentA = 10;
+    model.percentC = 20;
+    model.percentG = 30;
+    model.percentT = 40;
+    model.length = 100 * 1000 * 1000;
```
DmitriiSukhomlinov, line 598: `100 * 1024 * 1024`?

DmitriiSukhomlinov (approved):

src/corelibs/U2Formats/src/FastaFormat.cpp
yalgaer: This is a constant that makes the original chunk size (FASTA_SEQUENCE_LINE_LENGTH) several order of magnitude bigger, so the expensive per-chunk transactional code overhead becomes negligible. This can be any big number, there are only 2 limitations here: it must be a multiple of FASTA_SEQUENCE_LINE_LENGTH and small enough to fit into memory.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: This one can be any big number that will make the test slow or will make the saved size != read size . It has not relation to low-level powers of 2.

PR #494, opened 2021-08-04
UGENE-6504. Memory reallocating error when exporting big fasta to CLUSTALW

yalgaer:
There is no safe way today to handle "unlimited-size" MSA in UGENE today, because a lot of scenarios around the codebase load the whole MSA data into memory. This way fixing a single format (ClustalW or any other) will not help: MSA will be loaded into memory in other places.

The simplest way to fix the problem with MSA today is to restrict the size of the MSA that UGENE can handle and do not produce objects we can't handle: I set the limit to 100Mb and tested and even 2-3x values are safe.

During the testing of a large sequence (>2Gb) I found multiple issues in Sequence view and here it is a different story: huge sequences are normal and should be handled by UGENE correctly. I included only trivial int32->int64 fixes into this patch for the Sequence View to avoid SAFE_POINTS and make 2Gb+ sequences openable. Other fixes will be submitted as separate issues.


src/corelibs/U2Formats/src/tasks/ExportTasks.h
```diff
@@ -42,6 +42,13 @@ class U2FORMATS_EXPORT ExportAlignmentTask : public DocumentProviderTask {
 
     void run() override;
 
+    /**
+     * Maximum MSA size in bytes that can be loaded/stored by UGENE safely.
+     * UGENE loads a full MSA model into a memory in a lot of places across the codebase.
+     * This constant should be limit the maxim size of MSA produced by UGENE, so UGENE won't create objects it can't handle.
+     */
+    static constexpr qint64 MAX_SAFE_ALIGNMENT_SIZE_TO_EXPORT = 100 * 1000 * 1000;
```
DmitriiSukhomlinov, line 50: "100 * 1024 * 1024" is a bit more correct

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -635,6 +634,21 @@ GUI_TEST_CLASS_DEFINITION(test_7367) {
     CHECK_SET_ERR(percentT >= model.percentT - diff && percentT <= model.percentT + diff, "Invalid percent of T: " + QString::number(percentT));
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7368) {
+    // Generate a large sequence (>=100mb).
+    // Check that error dialog is shown when such a big sequence is exported as an alignment.
+    DNASequenceGeneratorDialogFillerModel model(sandBoxDir + "/test_7368.fa");
+    model.length = 100 * 1000 * 1000;
```
DmitriiSukhomlinov, line 641: `MAX_SAFE_ALIGNMENT_SIZE_TO_EXPORT` is a public variable, you can use it here

src/plugins/dna_export/src/ExportSequences2MSADialog.cpp
```diff
@@ -22,21 +22,18 @@
...
 
-#include <U2Gui/DialogUtils.h>
 #include <U2Gui/HelpButton.h>
 #include <U2Gui/SaveDocumentController.h>
 
 namespace U2 {
 
 ExportSequences2MSADialog::ExportSequences2MSADialog(QWidget *p, const QString &defaultUrl)
-    : QDialog(p),
+    : QDialog(p), Ui_ExportSequences2MSADialog(),
```
DmitriiSukhomlinov, line 36: This isn't necessary, but doesn't make any problems

DmitriiSukhomlinov (dismissed):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: Unfortunately one plugin can't depend on another plugin and the constant is declared in the dna_export plugin.

I think we need to create something like "UGENE Limits".h in U2Core. I will do it in a separate patch and will move this cosntant there.

src/corelibs/U2Formats/src/tasks/ExportTasks.h
yalgaer: When I see powers of 2 I start thinking about low-level platform limitation: how many values in int32 or like that.

In bioinformatics we do not use powers of two: 1kbp is equal to 1000bp - all our code is written that way (see rulers, coordinate formatters etc.)

Here this limitation is not a platform limitation, but just a fact that for some big values UGENE becomes very slow.
100M here is not bytes. We can only approximately estimate how much memory is needed to store an alignment with 100M characters: it can be 100Mb or 200Mb or even 10Gb

src/plugins/dna_export/src/ExportSequences2MSADialog.cpp
yalgaer: OK, looks like I need to disable this warning/inspection in my CLion

PR #501, opened 2021-08-05
UGENE-7269 Add file filters to Align sequences... and Align alignment... dialogs

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7269

### DIFF
--- a/src/corelibs/U2Gui/src/util/DialogUtils.h
+++ b/src/corelibs/U2Gui/src/util/DialogUtils.h
@@ -51,6 +51,14 @@ public:
 
     static QString prepareDocumentsFileFilterByObjType(const GObjectType &t, bool any);
 
+    /**
+     * Create a list, separated by ";;" of the supported formats.
+     * @types the list of types, which format should be supported.
+     * @any contain the "All files" filter if true.
+     * @return the ";;" separated list of filters.
+     */
+    static QString prepareDocumentsFileFilterByObjTypes(const QList<GObjectType>& types, bool any);
+
     static void setWizardMinimumSize(QWizard *wizard, const QSize &minimumSize = QSize());
 };
 
--- a/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
+++ b/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
@@ -100,9 +100,8 @@ void AlignSequencesToAlignmentAction::sl_activate() {
         QString f1 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, false);
         QString f2 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
         QString filter = f2 + "\n" + f1;
-        auto selectedFilter = DialogUtils::prepareDocumentsFileFilter(BaseDocumentFormats::FASTA, false);
         LastUsedDirHelper lod;
-        QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter, &selectedFilter);
+        QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter);
 
         if (!urls.isEmpty()) {
             lod.url = urls.first();
--- a/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
+++ b/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
@@ -97,9 +97,7 @@ void AlignSequencesToAlignmentAction::sl_activate() {
             AppContext::getTaskScheduler()->registerTopLevelTask(task);
         }
     } else {
-        QString f1 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, false);
-        QString f2 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
-        QString filter = f2 + "\n" + f1;
+        QString filter = DialogUtils::prepareDocumentsFileFilterByObjTypes({ GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, GObjectTypes::SEQUENCE }, true);
         LastUsedDirHelper lod;
         QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter);
 
--- a/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
+++ b/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
@@ -98,8 +98,9 @@ void AlignSequencesToAlignmentAction::sl_activate() {
         }
     } else {
         QString filter = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
+        auto selectedFilted = DialogUtils::prepareDocumentsFileFilter(BaseDocumentFormats::FASTA, false);
         LastUsedDirHelper lod;
-        QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter);
+        QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter, &selectedFilted);
 
         if (!urls.isEmpty()) {
             lod.url = urls.first();
--- a/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
+++ b/src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
@@ -97,10 +97,12 @@ void AlignSequencesToAlignmentAction::sl_activate() {
             AppContext::getTaskScheduler()->registerTopLevelTask(task);
         }
     } else {
-        QString filter = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
-        auto selectedFilted = DialogUtils::prepareDocumentsFileFilter(BaseDocumentFormats::FASTA, false);
+        QString f1 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, false);
+        QString f2 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
+        QString filter = f2 + "\n" + f1;
+        auto selectedFilter = DialogUtils::prepareDocumentsFileFilter(BaseDocumentFormats::FASTA, false);
         LastUsedDirHelper lod;
-        QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter, &selectedFilted);
+        QStringList urls = U2FileDialog::getOpenFileNames(nullptr, tr("Open file with sequences"), lod.dir, filter, &selectedFilter);
 
         if (!urls.isEmpty()) {
             lod.url = urls.first();
--- a/src/plugins/GUITestBase/src/GTUtilsWorkflowDesigner.cpp
+++ b/src/plugins/GUITestBase/src/GTUtilsWorkflowDesigner.cpp
@@ -951,7 +951,7 @@ void GTUtilsWorkflowDesigner::setDatasetInputFile(GUITestOpStatus &os, const QSt
 
     GTFileDialogUtils::TextInput t = pastePath ? GTFileDialogUtils::CopyPaste : GTFileDialogUtils::Typing;
 
-    GTFileDialogUtils *ob = new GTFileDialogUtils(os, filePath, GTGlobals::UseMouse, GTFileDialogUtils::Open, t);
+    GTFileDialogUtils *ob = new GTFileDialogUtils(os, filePath, GTGlobals::UseMouse, GTFileDialogUtils::Open, QString(), t);
     GTUtilsDialog::waitForDialog(os, ob);
 
     GTWidget::click(os, addFileButton);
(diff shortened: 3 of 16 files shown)


src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
```diff
@@ -98,8 +98,9 @@ void AlignSequencesToAlignmentAction::sl_activate() {
         }
     } else {
         QString filter = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
+        auto selectedFilted = DialogUtils::prepareDocumentsFileFilter(BaseDocumentFormats::FASTA, false);
```
yalgaer: typo

yalgaer (dismissed):

yalgaer (changes requested):
The code looks good but for some reason too many (22) GUI tests have failed. Please check.

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
DmitriiSukhomlinov: Fixed

yalgaer:
Now I see the reason why the tests failed and changed my mind about "default filter selection".

The default "File Open Dialog" filter when multiple options are available should be the least restrictive one: so there will be no empty file list when a user has GenBank or Clustal files but not FASTA files in a folder.

I've checked all usages of "File Open Dialog" across UGENE (about 100 or them) and see that we have only 1 place where we use 'selectedFilter': this is a recently added dialog in trimmomatic tool. 

Once 'All Files' option is a legit filter - it should be used in UGENE by default. If it may break something - it should not be used at all. 

My suggestion about this PR:
1) Add MSA filter to the dialog and use correct API for that (DocumentFormatConstraints). This is a good change.
2) Keep your filter selector implementation in GTFileDialogUtils and add a default value (empty string) to it. This also may be a useful change for some future use-case.
3) Do not add any pre-selected filters to File Open Dialog at all.

I would even remove 'selectedFilter' option from our DialogUtils API - it is hard to imagine any good use-case for it.
The 'default selected filter' can always be the last in the list of filters:
a) If there is only 1 filter in the list: this filter is the last and is automatically selected.
b) If we allow "All Files (*.*)" - All files filter is the last and is automatically selected.
c) If we do not allow "All Files (*.*)" but have multiple filters, the last one should be a join of them. Like a "Sequences (*.fa, *.gb, *.fq)"
(this part about selectedFilter is out of scope of this PR, but if you do it I will approve).

yalgaer:
Closing since there is no activity. Please re-open if any work is planned.

DmitriiSukhomlinov:
> Now I see the reason why the tests failed and changed my mind about "default filter selection".
> 
> The default "File Open Dialog" filter when multiple options are available should be the least restrictive one: so there will be no empty file list when a user has GenBank or Clustal files but not FASTA files in a folder.
> 
> I've checked all usages of "File Open Dialog" across UGENE (about 100 or them) and see that we have only 1 place where we use 'selectedFilter': this is a recently added dialog in trimmomatic tool.
> 
> Once 'All Files' option is a legit filter - it should be used in UGENE by default. If it may break something - it should not be used at all.
> 
> My suggestion about this PR:
> 
> 1. Add MSA filter to the dialog and use correct API for that (DocumentFormatConstraints). This is a good change.
> 2. Keep your filter selector implementation in GTFileDialogUtils and add a default value (empty string) to it. This also may be a useful change for some future use-case.
> 3. Do not add any pre-selected filters to File Open Dialog at all.
> 
> I would even remove 'selectedFilter' option from our DialogUtils API - it is hard to imagine any good use-case for it.
> The 'default selected filter' can always be the last in the list of filters:
> a) If there is only 1 filter in the list: this filter is the last and is automatically selected.
> b) If we allow "All Files (_._)" - All files filter is the last and is automatically selected.
> c) If we do not allow "All Files (_._)" but have multiple filters, the last one should be a join of them. Like a "Sequences (*.fa, *.gb, *.fq)"
> (this part about selectedFilter is out of scope of this PR, but if you do it I will approve).

Filters are usually have the alphabetical order and the "All files" filter is located on the first place. It's not really alphabetically, because "ABIF" should be located earlier, but, "All files" is some kind of a special case and it looks better in the beginning:
image
And, I suppose, it looks a bit better than this:
image

So, my suggestion is not to change the location of "All files"

yalgaer:
> So, my suggestion is not to change the location of "All files"

I agree that "All files" on top (the first item in the list) is a good solution. At least other tools I checked on Linux (QtCreator, text editor) work like this.

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
```diff
@@ -97,7 +97,9 @@ void AlignSequencesToAlignmentAction::sl_activate() {
             AppContext::getTaskScheduler()->registerTopLevelTask(task);
         }
     } else {
-        QString filter = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
+        QString f1 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, false);
+        QString f2 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
```
yalgaer: Is the result list sorted or not? May be pass both types (list of types) to the prepareDocumentsFileFilterByObjType ?

src/plugins/GUITestBase/src/GTUtilsWorkflowDesigner.cpp
```diff
@@ -951,7 +951,7 @@ void GTUtilsWorkflowDesigner::setDatasetInputFile(GUITestOpStatus &os, const QSt
 
     GTFileDialogUtils::TextInput t = pastePath ? GTFileDialogUtils::CopyPaste : GTFileDialogUtils::Typing;
 
-    GTFileDialogUtils *ob = new GTFileDialogUtils(os, filePath, GTGlobals::UseMouse, GTFileDialogUtils::Open, t);
+    GTFileDialogUtils *ob = new GTFileDialogUtils(os, filePath, GTGlobals::UseMouse, GTFileDialogUtils::Open, QString(), t);
```
yalgaer: Can you move the filter parameter to the last in the parameters list please, so we won't be required to write a default value for it (looks like we have no single use of it today) ?

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.cpp
```diff
@@ -100,6 +104,8 @@ void GTFileDialogUtils::commonScenario() {
     GTGlobals::sleep(300);
     setViewMode(Detail);
     GTGlobals::sleep(300);
+    applyFilter();
```
yalgaer, line 106: Sleep only if applyFilter is present (may be move sleep inside of applyFIlter)

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.cpp
```diff
@@ -323,6 +329,19 @@ void GTFileDialogUtils::setViewMode(ViewMode v) {
 }
 #undef GT_METHOD_NAME
 
+#define GT_METHOD_NAME "applyFilter"
+void GTFileDialogUtils::applyFilter() {
+    if (filter.isEmpty()) {
+        return;
+    }
+
+    QComboBox* comboBox = fileDialog->findChild<QComboBox*>(FILE_TYPE_COMBO_BOX);
```
yalgaer: Use GTWidget::findComboBox

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.h
```diff
@@ -60,11 +60,13 @@ class GTFileDialogUtils : public Filler {
     void selectFile();
     void clickButton(Button);
     void setViewMode(ViewMode);
+    void applyFilter();
 
     QWidget *fileDialog;
     QString path, fileName;
     Button button;
     GTGlobals::UseMethod method;
+    QString filter;
```
yalgaer: Please document all new public fields: what is the format of this field?

yalgaer (changes requested):

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
DmitriiSukhomlinov: > Is the result list sorted or not?

The results are not sorted, they appeared in the combo box in the order they have been added

> May be pass both types (list of types) to the prepareDocumentsFileFilterByObjType ?

Sorry, don't gent the idea. You mean, to pass the argument like this?

`DialogUtils::prepareDocumentsFileFilterByObjType( { GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, GObjectTypes::SEQUENCE }, true);`

src/plugins/GUITestBase/src/GTUtilsWorkflowDesigner.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.cpp
DmitriiSukhomlinov: This function doesn't exist, but I added it

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
yalgaer: Yes, so the filters are sorted and the (*.* All) is the first.

DmitriiSukhomlinov:
XML: [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/XmlLinux/229448), [macOS](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/XmlMacOsX86_64/229450), [Windows](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/XmlWindows/229446).
GUI: [Linux](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/GuiLinux/229452?expandedTest=build%3A%28id%3A229452%29%2Cid%3A14420). There is some test failed, but, I suppose, it's some unstable failure. To be sure, I run it one more time [here](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/GuiLinux/229474).

DmitriiSukhomlinov:
Do I need to do anything else in this PR?

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.cpp
```diff
@@ -323,6 +328,18 @@ void GTFileDialogUtils::setViewMode(ViewMode v) {
 }
 #undef GT_METHOD_NAME
 
+#define GT_METHOD_NAME "applyFilter"
+void GTFileDialogUtils::applyFilter() {
+    if (filter.isEmpty()) {
+        return;
+    }
+
+    GTGlobals::sleep(300);
```
yalgaer, line 337: Do we really need this sleep? findComboBox will wait up to 30 seconds for the widget to appear

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
yalgaer: After this change we have 2 FASTA items in the file filter lis
fasta-2x
t

yalgaer (changes requested):

src/plugins/GUITestBase/src/base_dialogs/GTFileDialog.cpp
DmitriiSukhomlinov: I have added this sleep because the set of the similar functions (like `setViewMode` or `selectFile`) have sleep-wrappers like this. I do not know either this sleep is required or not. I can remove it if you want.

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
DmitriiSukhomlinov: Hm, yes, this problem really exists. `QFileDialog` doesn't remove duplicate filters on it's own, unfortunately. I can suggest you 2 solutions:
1. The local one. I will remove duplicates from every place in this commit, where I have added a double-filters like this:
    ```
    QString f1 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::MULTIPLE_SEQUENCE_ALIGNMENT, false);
    QString f2 = DialogUtils::prepareDocumentsFileFilterByObjType(GObjectTypes::SEQUENCE, true);
    QString filter = f2 + "\n" + f1;
    ```

2. The global one. I will add a new function `QString removeDuplicatedFilters(const QString &oldFilter)` to `U2FileDialog` and call this function in `U2FileDialog::getOpenFileName`, `U2FileDialog::getOpenFileNames` and `U2FileDialog::getSaveFileName` before `QFileDialog` creates the dialog.

Which one do I need to use?

DmitriiSukhomlinov:
I haven't commited anything, just added some comments - read them and give the answer about what way do I need to resolve your concerns in.

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
yalgaer: This is what I suggested in the first comment: a single function that accepts a list of object types.

yalgaer (changes requested):

src/corelibs/U2View/src/ov_msa/align_to_alignment/AlignSequencesToAlignmentSupport.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
Fixed

src/corelibs/U2Gui/src/util/DialogUtils.h
```diff
@@ -51,6 +51,14 @@ class U2GUI_EXPORT DialogUtils : public QObject {
 
     static QString prepareDocumentsFileFilterByObjType(const GObjectType &t, bool any);
 
+    /**
+     * Create a list, separated by ";;" of the supported formats.
```
yalgaer, line 55: nit: start method names with a verb: Create -> Creates

yalgaer (approved):

PR #502, opened 2021-08-06
UGENE-6170. Correct names of document formats in two workflow elements

yalgaer:


src/corelibs/U2Designer/src/DelegateEditors.h
```diff
@@ -201,13 +202,31 @@ private slots:
     QVariantMap spinProperties;
 };
 
-class U2DESIGNER_EXPORT ComboBoxDelegate : public PropertyDelegate {
+/** Base class for all combo-box delegates. Includes item name formatter. */
+class U2DESIGNER_EXPORT ComboBoxBaseDelegate : public PropertyDelegate {
+    Q_OBJECT
+public:
+    ComboBoxBaseDelegate(QObject *parent = 0);
```
DmitriiSukhomlinov: `nullptr` is a bit better

DmitriiSukhomlinov (dismissed):

smlnkcv (commented):
1.  "Filter BAM/SAM files": "Output format" attribute visible value in the same way: format names should be used instead if format IDs. Now  they are "sam", "bam"

src/corelibs/U2Designer/src/DelegateEditors.h
yalgaer: Agree. Fixed.

yalgaer:
>     1. "Filter BAM/SAM files": "Output format" attribute visible value in the same way: format names should be used instead if format IDs. Now  they are "sam", "bam"

Fixed

PR #512, opened 2021-08-11
UGENE-7193 In silico PCR doesh't create Products table in case of ope…

rasputinkirill:
…ning file via "Merge sequence ..." option


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -458,6 +459,28 @@ GUI_TEST_CLASS_DEFINITION(test_7183) {
...
+    //3. Enter the primers: "AAA" and "CCC".
+    GTUtilsPcr::setPrimer(os, U2Strand::Direct, "AAA");
+    GTUtilsPcr::setPrimer(os, U2Strand::Complementary, "CCC");
+
+    //4. Click the find button.
+    GTWidget::click(os, GTWidget::findWidget(os, "findProductButton"));
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    //Expected: one result found
+    CHECK_SET_ERR(1 == GTUtilsPcr::productsCount(os), "Wrong results count 2");
```
yalgaer: 1. Do not use yoda style. Compiler will catch assignments and report them as errors (+ you can't even have an assignment here)
2. Log the invalid value: add it to the fail message.

src/plugins/pcr/src/InSilicoPcrTask.cpp
```diff
@@ -111,6 +111,13 @@ void InSilicoPcrTask::prepare() {
     CHECK_OP(stateInfo, );
     FindAlgorithmTaskSettings reverseSettings = getFindPatternSettings(U2Strand::Complementary);
     CHECK_OP(stateInfo, );
+    if (qgetenv(ENV_GUI_TEST) == "1") {
```
yalgaer: This way you may affect other tests
Could you do the following
1) Use some bigger max values - not 1 but a more reasonable 10 for example?
2) Use a dedicated variable: UGENE_MAX_RESULTS_FOR_PRIMERS_PER_STRAND. Set it in the first line of you GUI test.

src/plugins/pcr/src/InSilicoPcrTask.h
```diff
@@ -106,6 +106,9 @@ class InSilicoPcrTask : public Task {
     void updateSequenceByPrimer(const PrimerBind &primer, QByteArray &productSequence) const;
 
 private:
+    //it takes ~6 minutes to process matrix 50x50 results with default algorithm settings in release build
```
yalgaer, line 109: Why 6 minutes are good, but 1 hour is not? If it means "compute  fast and do not make users to wait" this limit should be much lower: 10-30 seconds

Is it possible to ask about this value (how long a user can wait)  in the related configuration dialog, so a user will see this value and will be able to adjust it?

yalgaer (changes requested):

src/plugins/pcr/src/InSilicoPcrTask.h
rasputinkirill: I just choose this value because with default values for this task and 3bp primers we wait 6 minutes and get result table with 706 results. As i think its more than enough for user to choose desired primer from results (or even decide what search parameters should be altered).
Anyway if you don't like my value feel free to suggest your.

yalgaer: I would suggest adding an UI input: "Limit search time: N minutes".
This way users will be aware about this limit and will know what to expect.

The default value for the parameter can be set from the 'UGENE_MAX_RESULTS_FOR_PRIMERS_PER_STRAND' env variable.

yalgaer: BTW, question: is it OK to show a table with 700 results to user? A set of this size looks unmanageable. What do you think?

src/plugins/pcr/src/InSilicoPcrTask.cpp
```diff
@@ -111,6 +111,13 @@ void InSilicoPcrTask::prepare() {
     CHECK_OP(stateInfo, );
     FindAlgorithmTaskSettings reverseSettings = getFindPatternSettings(U2Strand::Complementary);
     CHECK_OP(stateInfo, );
+    if (qgetenv("UGENE_MAX_RESULTS_FOR_PRIMERS_PER_STRAND") == "1") {
```
yalgaer: This should be a value: parse it as an integer.

yalgaer (changes requested):

yalgaer:
I have another idea how to solve the problem.
Primer of length 3 (like you use in the test) is a non-realistic data. Can the problem with too many results and long execution time be solved if you limit the minimum length of the primer by a bigger number?

rasputinkirill:
I used short patterns just for easier reproduction problem.
In the original scenario we are using typical-length primers on the extremely big sequence.

yalgaer:
> I used short patterns just for easier reproduction problem.
> In the original scenario we are using typical-length primers on the extremely big sequence.

Neither super-short patterns nor extremely large sequences are realistic for the primer design.
A good fix may be just a limit of the primer lengths and search region lengths. 
Could you please find such limits? What is a valid range for primers and what is a valid range for the final product?

src/plugins/pcr/src/InSilicoPcrTask.cpp
DmitriiSukhomlinov: The documentation should be update about this env. variable and about the default value `MAX_RESULTS_FOR_PRIMERS_PER_STRAND`

DmitriiSukhomlinov (approved):

src/plugins/pcr/src/InSilicoPcrTask.cpp
DmitriiSukhomlinov: Please, update and add a link here to docs

DmitriiSukhomlinov (changes requested):
See comments

src/plugins/pcr/src/InSilicoPcrTask.h
yalgaer: Ok, here is some hints for the final solution
https://www.google.com/search?q=pcr+product+length (I would suggest set a limit to 10k)
https://www.google.com/search?q=primer+length (my suggestion for the relaxed limit is 15-100)

With these numbers we do not need to solve the current problem the way you did at all: the problem won't exist.
I think you can close this PR and create a new one. Lets start from a clean and correct one.

smlnkcv (dismissed):

rasputinkirill:
According to the internet typical sizes: 
 - primer 18-30 bp
 - amplicon (product) 200-1000 bp, we got default 5000 length
 
But, anyway, user can run primer search on any sequence. If you want to limit it what value it should be?

yalgaer:
> According to the internet typical sizes:
> 
>     * primer 18-30 bp
> 
>     * amplicon (product) 200-1000 bp, we got default 5000 length
> 
> 
> But, anyway, user can run primer search on any sequence. If you want to limit it what value it should be?

Yes I think that if it is meaningless to search for primers of length 3 or 1000 we should not allow run such searches.

The same is applied for the product size: if the product size can't be > some number (like 5000) there is no need to search for a bigger products.

I would suggest to add these limits and do not run the search at all for invalid params.

src/plugins/pcr/src/InSilicoPcrTask.cpp
rasputinkirill: https://doc.ugene.net/wiki/pages/viewpage.action?pageId=69402659

src/plugins/pcr/src/InSilicoPcrTask.cpp
```diff
@@ -107,10 +107,24 @@ FindAlgorithmTaskSettings InSilicoPcrTask::getFindPatternSettings(U2Strand::Dire
 }
 
 void InSilicoPcrTask::prepare() {
+    if (!PrimerStatistics::validatePrimerMinLength(settings.forwardPrimer) || !PrimerStatistics::validatePrimerMinLength(settings.reversePrimer)) {
+        algoLog.details(tr("One of the given primers is too short. Task cancelled."));
+        stateInfo.setCanceled(true);
```
yalgaer, line 111: This should be failed, not canceled

+ Why don't we check these values in the dialog, before running the task?

rasputinkirill: What dialog do you mean?
Entered settings already checked in OP tab.

yalgaer: > Entered settings already checked in OP tab.

OP tab has no check for the length of the primers. It means that I can run primers search task even with primers like 'A'  and 'A' which is wrong (and will make UGENE hang).

Safety checks in tasks are good, but we should not even allow running primer search algorithm for invalid input.

src/corelibs/U2Core/src/datatype/primers/PrimerStatistics.h
```diff
@@ -39,9 +39,15 @@ class U2CORE_EXPORT PrimerStatistics : public QObject {
     static double getAnnealingTemperature(const QByteArray &product, const QByteArray &forwardPrimer, const QByteArray &reversePrimer);
 
     static bool validate(const QByteArray &primer);
+    static bool validatePrimerLength(const QByteArray &primer);
```
yalgaer, line 42: const?

src/corelibs/U2Core/src/datatype/primers/PrimerStatistics.h
```diff
@@ -39,9 +39,15 @@ class U2CORE_EXPORT PrimerStatistics : public QObject {
...
     static bool validate(QString primer);
 
     static QString getDoubleStringValue(double value);
+
+private:
+    //Length bounds of a typical primer
+    static constexpr int MAXIMUM_PRIMER_LENGTH = 30;
```
yalgaer: Give a user more space for experiments: I would suggest a range of 15-50

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -458,6 +460,59 @@ GUI_TEST_CLASS_DEFINITION(test_7183) {
     // Expected state: UGENE is not crash
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7193_1) {
+    qputenv("UGENE_MAX_RESULTS_FOR_PRIMERS_PER_STRAND", "30");
```
yalgaer: Once you limited primers lengths do you still need this trick with MAX_RESULTS?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -458,6 +460,59 @@ GUI_TEST_CLASS_DEFINITION(test_7183) {
...
+    GTWidget::click(os, GTWidget::findWidget(os, "OP_IN_SILICO_PCR"));
+
+    //3. Enter the primers: "GGAAAAAATGCTAAGGGC" and "CTGGGTTGAAAATTCTTT".
+    GTUtilsPcr::setPrimer(os, U2Strand::Direct, "GGAAAAAATGCTAAGGGC");
+    GTUtilsPcr::setPrimer(os, U2Strand::Complementary, "CTGGGTTGAAAATTCTTT");
+    //4. Set both mismatches to 9
+    GTUtilsPcr::setMismatches(os, U2Strand::Direct, 9);
+    GTUtilsPcr::setMismatches(os, U2Strand::Complementary, 9);
+    //5. Set 3' perfect match to 3
+    QSpinBox *perfectSpinBox = dynamic_cast<QSpinBox *>(GTWidget::findWidget(os, "perfectSpinBox"));
```
yalgaer: Use GTWidget::findSpinBox()

src/plugins/pcr/src/InSilicoPcrTask.cpp
```diff
@@ -49,9 +49,9 @@ bool InSilicoPcrProduct::isValid() const {
     return ta != Primer::INVALID_TM;
 }
 
-InSilicoPcrTask::InSilicoPcrTask(const InSilicoPcrTaskSettings &_settings)
-    : Task(tr("In Silico PCR"), TaskFlags(TaskFlag_ReportingIsSupported) | TaskFlag_ReportingIsEnabled | TaskFlags_FOSE_COSC),
-      settings(_settings), forwardSearch(nullptr), reverseSearch(nullptr), minProductSize(0) {
+InSilicoPcrTask::InSilicoPcrTask(const InSilicoPcrTaskSettings &settings)
+    : Task(tr("In Silico PCR"), TaskFlags(TaskFlag_ReportingIsSupported) | TaskFlag_ReportingIsEnabled | TaskFlag_FailOnSubtaskError),
+      settings(settings), forwardSearch(nullptr), reverseSearch(nullptr), minProductSize(0) {
```
yalgaer: keep the '_' variant of the name: do not have 2 same-name variables in the same scope

yalgaer (changes requested):
Please check inlined comments.

src/corelibs/U2Core/src/datatype/primers/PrimerStatistics.h
rasputinkirill: Modifiers not allowed on static member functions.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -458,6 +460,57 @@ GUI_TEST_CLASS_DEFINITION(test_7183) {
...
+
+    //2. Open the PCR OP.
+    GTWidget::click(os, GTWidget::findWidget(os, "OP_IN_SILICO_PCR"));
+
+    //3. Enter the primers: "AAA" and "CCC".
+    GTUtilsPcr::setPrimer(os, U2Strand::Direct, "AAA");
+    GTUtilsPcr::setPrimer(os, U2Strand::Complementary, "CCC");
+
+    //Expected state: there is a warning about forward primer length
+    QLabel *warningLabel = qobject_cast<QLabel *>(GTWidget::findWidget(os, "warningLabel"));
```
yalgaer: use GTWidget::findLabel, so you do not need the nullcheck.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -458,6 +460,57 @@ GUI_TEST_CLASS_DEFINITION(test_7183) {
...
+    CHECK_SET_ERR(warningLabel != nullptr, "Cannot find warningLabel");
+    CHECK_SET_ERR(warningLabel->text().contains("The forward primer length should be between"), "Incorrect warning message");
+
+    const GTLogTracer lt("One of the given do not fits acceptable length. Task cancelled.");
```
yalgaer: nit: remove const. We do not use them for local vars (example: in your code you do not use it for the QLabel above)

src/plugins/pcr/src/InSilicoPcrTask.cpp
yalgaer: Any updates for this part?

yalgaer (changes requested):
The code looks good (see minor comments), please ping me when ready

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: Done.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: Done.

rasputinkirill:
We still need this check in the task because it is possible to run task from workflow designer.
Primer length check in  OP Tab is made only to show warning in warning label which not block task run button and user able to run task with any primer length parameters.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: Do not see any changes here. Forgot to push the update?

yalgaer (changes requested):

rasputinkirill:
Test fixed.

yalgaer (approved):

PR #515, opened 2021-08-11
Fix 6759

ichebyki:
Fix 6759 - replace SAFE_POINT with checking index to -1 and return false if true


src/plugins/circular_view/src/CircularItems.cpp
```diff
@@ -436,14 +436,15 @@ bool CircularAnnotationLabel::tryPushClockwise() {
     const int engagedIndex = ra->engagedLabelPositionToLabel.key(this, -1);
     SAFE_POINT(-1 != engagedIndex, "Unexpected label position index", false);
     const int numberOfEngagedIndex = suitableLabelPositionIndexes.indexOf(engagedIndex);
-    SAFE_POINT(-1 != numberOfEngagedIndex, "Unexpected label position index", false);
 
-    if ((suitableLabelPositionIndexes.size() - 1) == numberOfEngagedIndex) {
+    if ((suitableLabelPositionIndexes.size() - 1) == numberOfEngagedIndex
+            || -1 == numberOfEngagedIndex) {
```
yalgaer: Do not use yoda style and add a comment why the index should be skipped as well as no index at all, so it will be easy to understand to anyone who reads this code in the future.

src/plugins/circular_view/src/CircularItems.cpp
```diff
@@ -456,14 +457,15 @@ bool CircularAnnotationLabel::tryPushCounterclockwise() {
     const int engagedIndex = ra->engagedLabelPositionToLabel.key(this, -1);
     SAFE_POINT(-1 != engagedIndex, "Unexpected label position index", false);
     const int numberOfEngagedIndex = suitableLabelPositionIndexes.indexOf(engagedIndex);
-    SAFE_POINT(-1 != numberOfEngagedIndex, "Unexpected label position index", false);
 
-    if (0 == numberOfEngagedIndex) {
+    if (0 == numberOfEngagedIndex
```
yalgaer: Same here: do not use yoda style (compiler already produces compile-time error if you will have an assignment instead of comparison in this case) + add a comment that describes what is going on here.

yalgaer (changes requested):

DmitriiSukhomlinov (changes requested):
Th GUI test is required

DmitriiSukhomlinov (commented):
The GUI test is still required

yalgaer (dismissed):

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -966,6 +966,10 @@ void AnnotatedDNAView::unregisterSplitWidget(ADVSplitWidget *splitWidget) {
     splitWidgets.removeOne(splitWidget);
 }
 
+QList<ADVSplitWidget *> AnnotatedDNAView::getSplitWidgets() {
```
yalgaer: This method is not needed. You can lookup any widget from GUI tests by object name.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -61,13 +61,17 @@
 
 #include <U2Gui/GUIUtils.h>
 
+#include <U2View/ADVSequenceObjectContext.h>
+#include <U2View/AnnotatedDNAView.h>
 #include <U2View/DetView.h>
 #include <U2View/McaEditorReferenceArea.h>
 
+#include "../../circular_view/src/CircularViewSplitter.h"
```
yalgaer: Cross-plugin deps are illegal ( a shared lib should be used)
But you do not no need this at all. The way you use splitter does not require any deps beyond qt.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,90 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+    CircularViewSplitter *splitter = static_cast<CircularViewSplitter *>(splitters[0]);
+    CHECK_SET_ERR(splitter != nullptr, QString("CircularViewSplitter is null"));
+
+    QList<QScrollBar *> scrolls = splitter->findChildren<QScrollBar *>("CircularViewSplitter_horScroll");
+    CHECK_SET_ERR(scrolls.size() == 1, QString("Too many scrolls"));
+    QScrollBar *horScroll = scrolls[0];
 
+    horScroll->setValue(horScroll->value() + 13);
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    GTGlobals::sleep(1000);
```
yalgaer: Please remove all sleeps from the test. Sleeps are legal only when they are part of the real logic. For example to separate 2 clicks from double click or wait until UGENE detects some FS changes.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,90 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+    horScroll->setValue(horScroll->value() + 26);
```
yalgaer: Never update QT UI model from non-UI thread.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,90 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+    AnnotatedDNAView *aview = ctx->getAnnotatedDNAView();
+    CHECK_SET_ERR(aview != nullptr, QString("No annoview for single sequence view. First check"));
+    QList<ADVSplitWidget *> splitters = aview->getSplitWidgets();
+    CHECK_SET_ERR(splitters.size() == 1,
+                  QString("No splitters for single sequence view. First check"));
+    CircularViewSplitter *splitter = static_cast<CircularViewSplitter *>(splitters[0]);
+    CHECK_SET_ERR(splitter != nullptr, QString("CircularViewSplitter is null"));
+
+    QList<QScrollBar *> scrolls = splitter->findChildren<QScrollBar *>("CircularViewSplitter_horScroll");
+    CHECK_SET_ERR(scrolls.size() == 1, QString("Too many scrolls"));
```
yalgaer: You do not need QString  wrapper here for the message

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,90 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+
+    //    Use context menu on annotation in tree view
+    GTUtilsDialog::waitForDialog(os,
+                                 new ProjectTreeItemSelectorDialogFiller(os,
+                                                                         "annotations.gb",
+                                                                         "Unknown features"));
+    GTUtilsDialog::waitForDialog(os, new CreateObjectRelationDialogFiller(os));
+    GTUtilsDialog::waitForDialog(os,
+                                 new PopupChooserByText(os,
+                                                        QStringList()
```
yalgaer: Use short C++11 syntax for lists like {"A", "B", "C"}

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,90 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+    //    4. Find splitter
+    //    5. Find scroll in splitter
+    //    6. Scroll to end, then to begin
+
+    GTFileDialog::openFileList(os,
+                               testDir + "_common_data/regression/6759/",
+                               QStringList() << "annotations.gb"
+                                             << "sequence.gb");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    GTUtilsSequenceView::checkSequenceViewWindowIsActive(os);
```
yalgaer: Usually you need either one check or another (the line before). Keep any you want, 2 checks are not needed.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,90 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+    CHECK_SET_ERR(ctx != nullptr, QString("No context for single sequence view."));
+    AnnotatedDNAView *aview = ctx->getAnnotatedDNAView();
+    CHECK_SET_ERR(aview != nullptr, QString("No annoview for single sequence view. First check"));
+    QList<ADVSplitWidget *> splitters = aview->getSplitWidgets();
+    CHECK_SET_ERR(splitters.size() == 1,
+                  QString("No splitters for single sequence view. First check"));
+    CircularViewSplitter *splitter = static_cast<CircularViewSplitter *>(splitters[0]);
+    CHECK_SET_ERR(splitter != nullptr, QString("CircularViewSplitter is null"));
+
+    QList<QScrollBar *> scrolls = splitter->findChildren<QScrollBar *>("CircularViewSplitter_horScroll");
```
yalgaer: Use GTWidget::find.. method which will check type and assert that the result exists.

yalgaer (changes requested):
The patch was good before the GUI test was added.
The way the test is written is not good by our current standards. 
If you have a strong feeling that GUI test is needed for this "no-op" change (in release mode) please update it.

DmitriiSukhomlinov (dismissed):
Approved, but run your GUI test from teamcity

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: done

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: done

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: done

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: done

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: done

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: No,
the scrolling can be optimized and therefore no scroll at all
Also, its may be to fast and human does not see what was happened

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
ichebyki: done

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
ichebyki: done

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
yalgaer: These tests are not for human watching, so sleeps can't be used in the tests just make it look better.
A test should be executed as fast as possible and a reliable as  possible.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5597,75 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+    //    1. Open sequence
+    //    2. Open annotation file
+    //    3. Add annotation file to sequence
+    //    4. Find splitter
+    //    5. Find scroll in splitter
+    //    6. Scroll to end, then to begin
+
+    GTFileDialog::openFileList(os,
+                               testDir + "_common_data/regression/6759/",
+                               QStringList() << "annotations.gb"
```
yalgaer: Open 2 files separately. We have an optimization for file-opening in UI tests so opening them separately (as in other tests) will take 1 second, while here each opening may take ~10-15 seconds

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5597,75 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+
+    GTUtilsSequenceView::openPopupMenuOnSequenceViewArea(os);
+    //    Check {add-> Objects with annotations} action
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    int seqNum = GTUtilsSequenceView::getSeqWidgetsNumber(os);
+    CHECK_SET_ERR(seqNum == 1, QString("Too many seqWidgets count").arg(seqNum));
+
+    QScrollBar *horScroll = GTWidget::findExactWidget<QScrollBar *>(os, "CircularViewSplitter_horScroll");
+    CHECK_SET_ERR(horScroll != nullptr, "Can't find circular view splitter's horScrollbar");
```
yalgaer: This check is not needed

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5597,75 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
 
+    // We use sleep as scrolling is executing too fast without sleep
+    horScroll->setValue(horScroll->value() + 13);
```
yalgaer: The problem remains: do not modify UI state from non-UI threads. This is a 'must' for any QT code (it is not thread safe)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
yalgaer: Looks like some problem with my diff viewer. Can't see any changes in diff

yalgaer (changes requested):
The original fix is 'no-op' in release mode. Will the test fail if I run it in debug with no fix applied?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5597,75 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+                                 new PopupChooserByText(os, {"Add", "Objects with annotations..."}));
+    //    On question "Found annotations that are out of sequence range, continue?" answer "Yes"
+    GTUtilsDialog::waitForDialog(os, new MessageBoxDialogFiller(os, QMessageBox::Yes));
+
+    GTUtilsSequenceView::openPopupMenuOnSequenceViewArea(os);
+    //    Check {add-> Objects with annotations} action
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    int seqNum = GTUtilsSequenceView::getSeqWidgetsNumber(os);
+    CHECK_SET_ERR(seqNum == 1, QString("Too many seqWidgets count").arg(seqNum));
```
yalgaer: .arg() with no %1

yalgaer (changes requested):
Now tests looks good except sleep methods. These methods should be removed.

If you want to wait until some task is finished use 'waitTaskFinished'

If you want to make QT a chance to process event queue (to render/deliver mouse events etc) use waitMainThread..

sleeps can be used only for something that is out of QT/UGENE control: double clicks, wait for file modification change(we do it by a timer inside) , etc.

ichebyki:
ok.....

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -5592,7 +5598,66 @@ GUI_TEST_CLASS_DEFINITION(test_6754) {
...
+
+    int seqNum = GTUtilsSequenceView::getSeqWidgetsNumber(os);
+    CHECK_SET_ERR(seqNum == 1, QString("Too many seqWidgets count: %1").arg(seqNum));
+
+    QScrollBar *horScroll = GTWidget::findExactWidget<QScrollBar *>(os, "CircularViewSplitter_horScroll");
+
+    // We use sleep as scrolling is executing too fast without sleep
+    // Also, we don't want to break different scrolls by some optimizations (if any)
+    GTScrollBar::moveSliderWithMouseToValue(os, horScroll, 13);
+    GTUtilsTaskTreeView::waitTaskFinished(os);
```
yalgaer: Which task are you waiting here? What is the cpp class name?

ichebyki: removed

yalgaer (approved):

DmitriiSukhomlinov (approved):

PR #517, opened 2021-08-11
UGENE-7366. Smith–Waterman with CUDA builds with errors

evelinabis:


src/plugins/smith_waterman/src/sw_cuda_cpp.cpp
```diff
@@ -202,8 +202,8 @@ QList<resType> calculateOnGPU(const char *seqLib, int seqLibLength, ScoreType *q
     cudaMalloc((void **)&g_directionsRec, sizeQ);
 
     if (U2::SmithWatermanSettings::MULTIPLE_ALIGNMENT == resultView) {
-        cudaError errorMatrix = cudaMalloc(reinterpret_cast<void **>(&g_directionsMatrix), directionMatrixSize);
```
yalgaer, line 205: Please do not remove, but add error checks to all cudaMallocs in this file
Otherwise it will be a crash.

yalgaer (changes requested):

.gitignore
```diff
@@ -26,6 +26,7 @@ src/*.so.*
 *.vcxproj*
 *.tmp
 *.sln
+/src/plugins/smith_waterman/vc*.pdb
```
DmitriiSukhomlinov: What file is this about? As far as I know, the all ".pdb" files should be moved to the build directory

evelinabis: This file is generated in ugene\src\plugins\smith_waterman\_tmp\obj\debug\vc142.pdb

DmitriiSukhomlinov (dismissed):

src/plugins/smith_waterman/src/sw_cuda_cpp.cpp
```diff
@@ -176,34 +177,65 @@ QList<resType> calculateOnGPU(const char *seqLib, int seqLibLength, ScoreType *q
...
+                                cudaMalloc((void**)&g_directionsRec, sizeQ) };
 
     if (U2::SmithWatermanSettings::MULTIPLE_ALIGNMENT == resultView) {
         cudaError errorMatrix = cudaMalloc(reinterpret_cast<void **>(&g_directionsMatrix), directionMatrixSize);
         cudaError errorBacktrace = cudaMalloc(reinterpret_cast<void **>(&g_backtraceBegins), backtraceBeginsSize);
+        errors << errorMatrix << errorBacktrace;
+    }
+
+    bool isError = false;
+    for (cudaError error : errors) {
```
yalgaer: nit: find_if can fit the block 211...217 into 1 line

yalgaer (changes requested):
All is good. Could you please merge with master and check that XML/GUI tests have no regression.
The latest run shows 2 XML tests failed

src/plugins/smith_waterman/src/sw_cuda_cpp.cpp
evelinabis: I really want to use std, but I still don't understand when I can and when [I can't](https://github.com/ugeneunipro/ugene/pull/171#discussion_r556203899). Focus on the number of lines: if less than `for`, then use std?

yalgaer: I see your point. The general suggestion is to select the most readable variant: maximum of your domain code and minimum of 'helper API boilerplate'.

You can compare the result in character count or line count, but in general I think you can agree that this code (from your old PR):
```cpp
std::transform(inputPaths.begin(), inputPaths.end(), inputPaths.begin(), [](const QString &path) {
    return QFileInfo(path).absoluteFilePath();
});
```

Looks less readable than the suggested one:
```cpp
for (QString& path: inputPaths) {
    path = QFileInfo(path).absoluteFilePath();
}
```

 C++ has very verbose lambda syntax. In other languages I would prefer using lambdas, not an old-style loop for the same code.

Example (JavaScript):
```cpp
inputPath = inputPath.map(path => new QFileInfo(path).absoluteFilePath());
```

Or even shorter (Kotlin):
```cpp
inputPath = inputPath.map{ QFileInfo(it).absoluteFilePath() }
```

Below is the example what I call 'helper API boilerplate'. The same code but with our logic removed.

```cpp
// Variant 1:
std::transform(.begin(), .end(), .begin(), [](const QString &path) {
    return path;
});

// Variant 2:
for (QString& path: ) {
    path = path;
}

// JavaScript:
= .map(path => new path);

// Kotlin:
= .map{ it }
````

File: src/plugins/smith_waterman/src/sw_cuda_cpp.cpp [hunk 4]
```diff
@@ -176,34 +178,58 @@ QList<resType> calculateOnGPU(const char *seqLib, int seqLibLength, ScoreType *q
...
+                                cudaMalloc((void **)&g_HdataRec, sizeQ),
+                                cudaMalloc((void **)&g_directionsRec, sizeQ) };
 
     if (U2::SmithWatermanSettings::MULTIPLE_ALIGNMENT == resultView) {
         cudaError errorMatrix = cudaMalloc(reinterpret_cast<void **>(&g_directionsMatrix), directionMatrixSize);
         cudaError errorBacktrace = cudaMalloc(reinterpret_cast<void **>(&g_backtraceBegins), backtraceBeginsSize);
+        errors << errorMatrix << errorBacktrace;
+    }
+
+    if (std::find_if(errors.constBegin(), errors.constEnd(), [](const cudaError error) {
```
yalgaer, line 212: nit: remove const from param or use const type& to make it more reasonable (we do not use method-local const across the code)

yalgaer (commented):
I see XML tests finished correctly. Need only GUI tests & will approve

PR #520, opened 2021-08-12
Add a custom launcher option (like crash handler) to GUI tests runner

yalgaer:
I'm experimenting with a custom crash handler for GUI tests to have a stack trace for crashed GUI tests.

The option I added allows adding a custom runner for every GUI test launch that can dump the trace when ugeneui crashes.


src/plugins/GUITestBase/src/harness/GUITestLauncher.cpp
```diff
@@ -366,13 +365,19 @@ QString GUITestLauncher::runTest(const QString &testName, const int timeout) {
...
 
-    QString path = QCoreApplication::applicationFilePath();
+    QString ugeneUiPath = QCoreApplication::applicationFilePath();
     QStringList arguments = getTestProcessArguments(testName);
 
     // ~QProcess is killing the process, will not return until the process is terminated.
     QProcess process;
     process.setProcessEnvironment(environment);
-    process.start(path, arguments);
+    QString ugeneTestRunWrapper = qgetenv("UGENE_TEST_RUN_WRAPPER");
```
DmitriiSukhomlinov, line 375: I think, all these env variables should be moved to WIKI. It's very difficult to search them every time in the code and a, actually, I don't even remember about at least a half of the existing variables

yalgaer: I would put all these options into a separate C++ header file with a detailed documentation.

PR #521, opened 2021-08-13
Support multi-selection in Export MSA sequences

yalgaer:


DmitriiSukhomlinov:
What issue is this?

DmitriiSukhomlinov (changes requested):
What issue is this?

yalgaer:
> What issue is this?

The following method is deprecated and all its usages should be reworked. There are about 50 more usages of this method that are not reworked yet. None of them is a blocker for the feature completion (as asked by VB) and I will fix them all in 20-30 patches during the next month or two. This patch is N5 in the series.

```
/**
     * Returns selection state as a rect.
     * The returned rect is a bounding rect for all rects in the 'rectList'.
     * If 'rectList' is empty the method returns an empty (0, 0, 0, 0) rect.
     * Warning: this method is unsafe and will be removed. Use getRectList() to deal with each individual selection rect correctly.
     */
    QRect toRect() const;
```

PR #523, opened 2021-08-13
Fix all -Wshadow=local warnings in QSpec library

yalgaer:
Originally I wanted to fix 'shadows' across the whole UGENE code but 
1) there are too many of such warnings for a single PR
2) QSpec places with these warnings is a big copy-paste chunk that needed its own cleanup.

So I limited my changes to QSpec package only.  In GUITestBase I only updated spelling/includes in the files affected by QSpec changes


src/libs_3rdparty/QSpec/src/primitives/GTAction.cpp
```diff
@@ -18,118 +18,62 @@
...
-            }
-        }
-        GT_CHECK_RESULT(list.count() < 2, QString("There are %1 actions with this text").arg(list.count()), NULL);
-        if (options.failIfNotFound) {
-            GT_CHECK_RESULT(list.count() != 0, "action not found", NULL);
-            return list.takeFirst();
-        }
-        return NULL;
+QAction *GTAction::findAction(GUITestOpStatus &os, const QString &objectName, QObject *parent, const GTGlobals::FindOptions &options) {
+    QList<QAction *> actions = GTWidget::findChildren<QAction>(os, parent, [objectName](QAction *action) -> bool { return action->objectName() == objectName; });
```
DmitriiSukhomlinov, line 61: `QList<QAction *> actions = GTWidget::findChildren<QAction>(os, parent, [&objectName](QAction *action) -> bool { return action->objectName() == objectName; });`

DmitriiSukhomlinov: `[&objectName]`

DmitriiSukhomlinov: nit: you don't need to write a lambda return type if this type is a primitive one(like `bool`, `int`, etc)

DmitriiSukhomlinov: The same about the all similar cases below

src/libs_3rdparty/QSpec/src/primitives/GTMainWindow.cpp
```diff
@@ -60,6 +60,18 @@ QList<QWidget *> GTMainWindow::getMainWindowsAsWidget(GUITestOpStatus &os) {
...
 
+#define GT_METHOD_NAME "checkTitle"
+void GTMainWindow::checkTitle(GUITestOpStatus &os, const QString &title) {
+    QList<QWidget *> mainWindowWidgets = getMainWindowsAsWidget(os);
+    for (const QWidget *widget : qAsConst(mainWindowWidgets)) {
+        if (widget->windowTitle() == title) {
+            return;
+        }
+    }
+    GT_CHECK_RESULT(false, QString("Can't find window with the title: '%1'").arg(title), );
```
DmitriiSukhomlinov: I suggest you change it a bit:

```
    QList<QWidget *> mainWindowWidgets = getMainWindowsAsWidget(os);
    bool found = false;
    for (const QWidget *widget : qAsConst(mainWindowWidgets)) {
        if (widget->windowTitle() == title) {
            found = true;
            break;
        }
    }
    GT_CHECK_RESULT(found, QString("Can't find window with the title: '%1'").arg(title), );
```

This is cuz `GT_CHECK_RESULT` texts to log the information about this check - either this function passed or failed there will be the info about it. In your code this function will be "logged" only if it's failed

src/libs_3rdparty/QSpec/src/primitives/GTWidget.h
```diff
@@ -148,6 +150,35 @@ class HI_EXPORT GTWidget {
...
+#define GT_METHOD_NAME "findChildren"
+    /**
+     * Finds all children of the 'parent' using 'findChildren' method and checkFn to check if the child is matched.
+     * If parent is null, find all child in all main windows.
+     * Note: this function does not wait until any child is found and returns immediately.
+     */
+    template<class ChildType>
+    static QList<ChildType *> findChildren(GUITestOpStatus &os, const QObject *parent, std::function<bool(ChildType *)> checkFn) {
+        QList<ChildType *> children;
+        auto checkAndAddToList = [&children, &checkFn](const QList<ChildType *> candidates) -> void {
```
DmitriiSukhomlinov: nit: not need to text `void` as a return value - the compiler will figure it out

src/libs_3rdparty/QSpec/src/primitives/GTMainWindow.cpp
DmitriiSukhomlinov: BTW, the original function below (from GTUtilsApp) has this check in any case

DmitriiSukhomlinov (changes requested):
See comments

src/libs_3rdparty/QSpec/src/primitives/GTWidget.h
yalgaer: I do not want a compiler to use some implicit type it can derive: I want the return to be explicit and and make the compiler fail anytime I return a different value.

So if I add a line into the method like "return 0" I want this issue to be caught by the compiler because the declared type is 'void'. This is the reason why I always add types to lambdas: "type system" is a good tool and it should not be ignored.

It may be easier to understand if you check an optional 'override' keyword. We can also skip it but for some good reason we don't

src/libs_3rdparty/QSpec/src/primitives/GTAction.cpp
yalgaer: objectName -> fixed
types on lambdas -> answered in the separate thread.

src/libs_3rdparty/QSpec/src/primitives/GTMainWindow.cpp
yalgaer: I see your point, but think there is a more readable solution. 
I added a GT_FAIL directive: unconditional failure.

An error message in the log + class + method name should be enough to provide the context

src/libs_3rdparty/QSpec/src/primitives/GTWidget.h
yalgaer: I read more about lambdas & code style recommendations for them and I want to say that now I agree, that is is OK to skip explicit types for a single statement lambdas. I will update my patch before submission.

PR #526, opened 2021-08-16
UGENE-7384. Refactor & fix graphs support in UGENE 

yalgaer:
The patch is very big because it is a global rewrite of badly designed multi-series/multi-threading graph calculation & controllers code which had dozen of bugs.

The design of the new solution can be improved too, but it will make the patch even bigger. I will do it incremental steps (other PRs).

To review the code I  suggest to check not the diff, but the code itself. Diff is not readable because it is mostly rewrite (~1000 lines of code were removed).

Alternatively it may be very useful to help me to test the build manually.  
I already see that it is more stable than the old code and produces more meaningful results, but of course it may have bugs.

I will add more GUI tests until tomorrow or before the patch is submitted.


src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
```diff
@@ -19,175 +19,116 @@
...
-        dataStartIdx = qMax((float)0, dataStartIdx);
-        dataEndIdx = qMin((float)data.size() - 1, dataEndIdx);
-        float nDataPointsToAverage = dataEndIdx - dataStartIdx;
-        float val = unknownVal;
-        if (nDataPointsToAverage >= ACCEPTABLE_FLOAT_PRESISION_LOSS) {
-            val = calculateAverage(data, dataStartIdx, nDataPointsToAverage);
-        }
-        results.append(val);
+float GSequenceGraphUtils::getMaxValue(float value1, float value2) {
+    return isUndefined(value1) ? value2 : isUndefined(value2) ? value1
```
DmitriiSukhomlinov, line 67: What if both of them are undefined?

yalgaer: The result will be undefined. Undefined values in viewPoints and visibleMin/Max must be checked and handled correctly.

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
```diff
@@ -197,622 +138,305 @@ void GSequenceGraphDrawer::sl_calculationTaskFinished() {
...
+
+    QPen graphPen(Qt::SolidLine);
+    graphPen.setColor(seriesColorByName.value(seriesColorByName.contains(graph->graphName) ? graph->graphName : DEFAULT_COLOR));
+    graphPen.setWidth(1);
+    p.setPen(graphPen);
+
+    int graphHeight = rect.height() - 2;
+    float valueToHeightScale = visibleMin == visibleMax ? 1 : (graphHeight - 1) / (visibleMax - visibleMin);
+
+    int prevX = INT_MAX;  // INT_MAX is a default value that means that prevX/Y is not defined yet.
```
DmitriiSukhomlinov, line 198: So is that the maximum `int` value? If yes, it's better to use `numeric_limits`

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
```diff
@@ -197,622 +138,305 @@ void GSequenceGraphDrawer::sl_calculationTaskFinished() {
...
     }
 
-    label->setValue(value);
-    if (commdata.enableCuttoff) {
-        if (value >= commdata.maxEdge) {
-            ycoordInRect = graphHeight;
-        } else if (value > commdata.minEdge && value < commdata.maxEdge) {
-            ycoordInRect = graphHeight / 2;
+    int labelY;
+    int graphHeight = rect.bottom() - rect.top() - 2;
```
DmitriiSukhomlinov, line 363: What's the magic number `2`?

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
yalgaer: Could you please share more details why to prefer one over another?

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
yalgaer: Same as in the original code: a space used to draw top/bottom lines.

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
```diff
@@ -823,188 +447,141 @@ void GSequenceGraphDrawer::showSettingsDialog() {
...
+    CHECK(!sequenceObject.isNull(), );
+
+    GTIMER(cvar, tvar, "GraphPointsUpdater::calculateAlgorithmPoints");
+    for (auto graphData : qAsConst(graphDataList)) {
+        try {
+            QVector<float> dataPoints;
+            graphData->algorithm->calculate(dataPoints, sequenceObject, graphData->window, graphData->step, stateInfo);
+            CHECK_OP(stateInfo, );
+            result << dataPoints;
+        } catch (const std::bad_alloc &) {
```
DmitriiSukhomlinov, line 477: `TaskScheduler` can catch this exception, not sure that it's necessary

yalgaer: It won't give the same level of details in the error message and won't execute other lines below the catch block.

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
DmitriiSukhomlinov: For the same reason we prefer to use `static constexpr` where it possible instead of `static const` - this is the compile-time computed variable (which means performance improving in particular and the most modern approach in general).

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.h
```diff
@@ -36,214 +36,214 @@ namespace U2 {
...
+     * */
+    static const float UNDEFINED_GRAPH_VALUE;
 };
 
+/** Map of colors by data series name. */
 typedef QMap<QString, QColor> ColorMap;
 
-class U2VIEW_EXPORT GSequenceGraphDrawer : public QObject {
+/** Stateful graph drawing controller. Holds current setting of the graph widget. */
+class U2VIEW_EXPORT GSequenceGraphDrawer : public QObject, private GSequenceGraphUtils {
```
DmitriiSukhomlinov, line 96: `GSequenceGraphUtils` contains `static` functions only, I don't get the idea of the inheritance.

src/corelibs/U2View/src/ov_sequence/GSequenceGraphView.cpp
```diff
@@ -94,82 +87,74 @@ GSequenceGraphView::GSequenceGraphView(QWidget *p, SequenceObjectContext *ctx, G
...
     update();
 }
 void GSequenceGraphView::mousePressEvent(QMouseEvent *me) {
     setFocus();
 
     if (me->modifiers() == Qt::ShiftModifier && me->button() == Qt::LeftButton) {
-        float pos = toRenderAreaPoint(me->pos()).x() / renderArea->getCurrentScale() + getVisibleRange().startPos;
-        addLabel(pos);
+        float sequencePos = toRenderAreaPoint(me->pos()).x() / renderArea->getCurrentScale() + getVisibleRange().startPos;
+        int capturingDistancePx = 4;
```
DmitriiSukhomlinov, line 100: What's the magic number? At least make in `static constexpr`

src/corelibs/U2View/src/ov_sequence/GSequenceGraphView.cpp
```diff
@@ -94,82 +87,74 @@ GSequenceGraphView::GSequenceGraphView(QWidget *p, SequenceObjectContext *ctx, G
...
-    }
+void GSequenceGraphView::leaveEvent(QEvent *) {
+    updateMovingLabels();
 }
-void GSequenceGraphView::getLabelPositions(QList<QVariant> &labelPositions) {
-    graphs.at(0)->graphLabels.getLabelPositions(labelPositions);
+
+void GSequenceGraphView::getSavedLabelsState(QList<QVariant> &savedLabels) {
+    // TODO: save/restore labels from all graphs.
+    graphs.at(0)->labels.getLabelPositions(savedLabels);
```
DmitriiSukhomlinov, line 126: Check that the first element exists?

src/corelibs/U2View/src/ov_sequence/GSequenceGraphView.cpp
```diff
@@ -247,69 +232,49 @@ void GSequenceGraphView::sl_onShowVisualProperties(bool) {
...
-                }
-            } else
-                graphDrawer->selectExtremumPoints(graph, graphRect, windowSize, getVisibleRange());
+void GSequenceGraphView::sl_showLocalMinMaxLabels() {
+    QVector<U2Region> regionsToAnnotate = getSequenceContext()->getSequenceSelection()->getSelectedRegions();
+    if (regionsToAnnotate.isEmpty()) {
+        regionsToAnnotate.append(getVisibleRange());
+    }
+    const QRect &graphRect = getGraphRenderArea()->getGraphRect();
+    for (QSharedPointer<GSequenceGraphData> &graph : graphs) {
```
DmitriiSukhomlinov: `for (const QSharedPointer<GSequenceGraphData> &graph : graphs) {`

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.cpp
yalgaer: Which of these 2 values is static const?

src/corelibs/U2View/src/ov_sequence/ADVGraphModel.h
yalgaer: It is a pattern to avoid adding a long class name  prefix before each utility method

src/corelibs/U2View/src/ov_sequence/GSequenceGraphView.cpp
yalgaer: it is not a magic number - it has a dedicated name

src/corelibs/U2View/src/ov_sequence/GSequenceGraphView.cpp
yalgaer: GSequenceView wont open with no graphs.

src/corelibs/U2View/src/ov_sequence/GSequenceGraphView.cpp
yalgaer: This one is good. Fixed

PR #541, opened 2021-08-23
Support multi-selection for Paste action in MSA editor

yalgaer:
I also fixed type of the msaIndex from qint64 from int everywhere in DBI. This way we differentiate it from rowId.

### DIFF
--- a/src/corelibs/U2Core/src/dbi/U2DbiPackUtils.cpp
+++ b/src/corelibs/U2Core/src/dbi/U2DbiPackUtils.cpp
@@ -140,7 +140,7 @@ bool U2DbiPackUtils::unpackGapDetails(const QByteArray &modDetails, qint64 &rowI
 
 QByteArray U2DbiPackUtils::packRowOrder(const QList<qint64> &rowIds) {
     QByteArray result;
-    foreach (qint64 rowId, rowIds) {
+    for (qint64 rowId : qAsConst(rowIds)) {
         if (!result.isEmpty()) {
             result += ",";
         }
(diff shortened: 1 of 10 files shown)


src/corelibs/U2Core/src/dbi/U2DbiPackUtils.cpp
```diff
@@ -295,15 +264,15 @@ QByteArray U2DbiPackUtils::packRow(qint64 posInMca, const U2McaRow &row) {
...
-    SAFE_POINT(8 == tokens.size(), QString("Invalid added row modDetails string '%1'").arg(QString(modDetails)), false);
+    SAFE_POINT(tokens.size() == 8, QString("Invalid added row modDetails string '%1'").arg(QString(modDetails)), false);
     {  // version
         SAFE_POINT(VERSION == tokens[0], QString("Invalid modDetails version '%1'").arg(tokens[0].data()), false);
     }
-    {  // posInMsa
+    {  // row index
         bool ok = false;
-        posInMca = tokens[1].toLongLong(&ok);
+        rowIndex = tokens[1].toInt(&ok);
```
DmitriiSukhomlinov, line 275: You may lose the data here if `tokens[1]` is large enough

src/corelibs/U2Formats/src/mysql_dbi/MysqlMsaDbi.cpp
```diff
@@ -423,7 +413,6 @@ void MysqlMsaDbi::removeRows(const U2DataId &msaId, const QList<qint64> &rowIds,
 
 void MysqlMsaDbi::removeRow(const U2DataId &msaId, qint64 rowId, U2OpStatus &os) {
     MysqlTransaction t(db, os);
-    Q_UNUSED(t);
```
DmitriiSukhomlinov, line 426: Why'd you removed it? It's really unused

src/corelibs/U2Core/src/dbi/U2DbiPackUtils.cpp
yalgaer: How the tokens[1] can be large enough here?

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Formats/src/mysql_dbi/MysqlMsaDbi.cpp
yalgaer: It is not unused because constructor is a function call with side effects. I think the problem was some time ago with old compilers and now it is not a problem anymore. 
Today if we have a really unused variable the build (on Linux) will fail (we use -Werror=unused-variable)

rasputinkirill:
* What the reason for massive remove of Q_UNUSED macro?
* Some of the edited 'ifs' have parameters order changed, may be leave them as they were?

yalgaer:
>     * What the reason for massive remove of Q_UNUSED macro?
All removed Q_UNUSED were wrong and misleading. Compiler does not think that these vars are unused otherwise we have a compile time error (we do not allow unused variables).

>     * Some of the edited 'ifs' have parameters order changed, may be leave them as they were?
I always remove Yoda style & deprecated QT features (like foreach) from the code I change. This style is inconsistent with the most of UGENE code and should be fixed one day (e.g we do not use 2 or 3 different code styles).

PR #548, opened 2021-08-24
UGENE-7389 Need to add in the Data->Samples example of Sanger file

smlnkcv:
Added reference and 20 reads


yalgaer (changes requested):
Opening a genbank file will make a sequence view opened.
Save with ABIF files.

A single "ugenedb" file that will be opened MCA editor should be enough here.

yalgaer (approved):

PR #550, opened 2021-08-25
UGENE-7234 Crash "In silico PCR" sample: QMutex: destroying locked mutex

rasputinkirill:
Added code is similar like in other workers, for example RemoteDBFetcherWorker.cpp:171.


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -487,6 +487,34 @@ GUI_TEST_CLASS_DEFINITION(test_7212) {
     GTUtilsTaskTreeView::waitTaskFinished(os);
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7234) {
+    class InSilicoWizardScenario : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus &os) {
```
yalgaer, line 493: nit: always use override with virtual method

yalgaer (approved):

PR #555, opened 2021-08-30
Replace UGENE_USE_BUNDLED_SQLITE with UGENE_USE_SYSTEM_SQLITE

yalgaer:
The next step in the build files cleanup series.

UGENE_USE_BUNDLED_SQLITE is replaced with UGENE_USE_SYSTEM_SQLITE so by default we do not need any flag at all. Before this patch we always had -DUGENE_USE_BUNDLED_SQLITE=1 included into every compiler line.

The UGENE_USE_SYSTEM_SQLITE option is needed by external Linux integrators that do not want to include our own sqlite package into their OS.


DmitriiSukhomlinov (changes requested):
That's pretty nice, but we should have some repository of the all these environmental variables. If I need it once it'd be very difficult for me to find this variable in the certain place in the code.
Do you store it somewhere?

yalgaer:
> That's pretty nice, but we should have some repository of the all these environmental variables. If I need it once it'd be very difficult for me to find this variable in the certain place in the code.
> Do you store it somewhere?

I agree it would be nice to have such document. I'm not sure if we have one today.

PR #556, opened 2021-08-30
UGENE-6872 Crash on workflow "Trim and Map Sanger reads" with incorrect data

rasputinkirill:
Added limit for matrixLength parameter. Because crash occurred while memory allocating.


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6064,6 +6064,39 @@ GUI_TEST_CLASS_DEFINITION(test_6862_1) {
     CHECK_SET_ERR(value == "cistrome.xml", QString("Motif database value: expected 'cistrome.xml', current: '%1'").arg(value))
 }
 
+GUI_TEST_CLASS_DEFINITION(test_6872) {
+    GTUtilsWorkflowDesigner::openWorkflowDesigner(os);
+
+    class Scenario : public CustomScenario {
```
yalgaer: nit: Give the class a meaningful name

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6064,6 +6064,39 @@ GUI_TEST_CLASS_DEFINITION(test_6862_1) {
...
+    public:
+        void run(HI::GUITestOpStatus &os) {
```
yalgaer: Add override to virtual functions

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6064,6 +6064,39 @@ GUI_TEST_CLASS_DEFINITION(test_6862_1) {
...
+            QWidget *wizard = GTWidget::getActiveModalWidget(os);
+            GTWidget::clickWindowTitle(os, wizard);
+
+            GTUtilsWizard::setParameter(os, "Reference", QFileInfo(dataDir + "samples/FASTA/human_T1.fa").absoluteFilePath());
+            GTUtilsWizard::clickButton(os, GTUtilsWizard::Next);
+
+            GTUtilsWizard::setInputFiles(os, {{dataDir + "samples/FASTA/human_T1.fa"}});
+            GTUtilsWizard::clickButton(os, GTUtilsWizard::Next);
+
+            const QString expectedRowNamingPolicy = "Sequence name from file";
```
yalgaer: Use consistent code style: do not use consts for locals (we do not use them across UGENE code today)

yalgaer (changes requested):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6089,8 +6089,12 @@ GUI_TEST_CLASS_DEFINITION(test_6872) {
             GTUtilsWizard::clickButton(os, GTUtilsWizard::Run);
         }
     };
+    //1. Open "Trim and Map Sanger reads" sample in workflow.
+    //2. Set human_T1.fa as input files on first and second wizard pages.
+    //3. Run schema.
+    //Expected state: workflow stopped work with "No read satisfy minimum similarity criteria" error message in the log.
     GTLogTracer l;
-    GTUtilsDialog::waitForDialog(os, new WizardFiller(os, "Map Sanger Reads to Reference", new Scenario));
+    GTUtilsDialog::waitForDialog(os, new WizardFiller(os, "Map Sanger Reads to Reference", new FillTrimAndMapWizardWithHumanT1));
```
yalgaer: nit: add braces to the constructor. (not a blocker)

yalgaer (changes requested):
I reviewed the code more accurately and here what I have found:

1) You need to check for out of memory error in the place where this out-of-memory error happens. Check the malloc result in the same file, no need to build a heuristic that won't work for all situations anyway.

2) Check that other SW algorithm impls are not affected by the same issue.
For example you fixed the standard SW algorithm code, but the SSE2 variant has absolutely the same code that you missed in the fix

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -138,6 +138,8 @@ bool SmithWatermanAlgorithm::isValidParams() {
         return false;
     if (gapOpen >= 0 || gapExtension >= 0)
         return false;
+    if (matrixLength > 100000)
```
DmitriiSukhomlinov: Add quotes {}

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -117,7 +117,7 @@ void SmithWatermanAlgorithm::launch(const SMatrix &_substitutionMatrix,
                                     int _minScore,
                                     SmithWatermanSettings::SWResultView _resultView) {
     setValues(_substitutionMatrix, _patternSeq, _searchSeq, _gapOpen, _gapExtension, _minScore, _resultView);
-    if (isValidParams() && calculateMatrixLength()) {
+    if (calculateMatrixLength() && isValidParams()) {
```
DmitriiSukhomlinov: Is there a reason to swap them?

rasputinkirill: Yep, we cant validate parameter if it wasn't calculated yet.

DmitriiSukhomlinov (changes requested):
See comments

rasputinkirill:
> I reviewed the code more accurately and here what I have found:
> 
> 1. You need to check for out of memory error in the place where this out-of-memory error happens. Check the malloc result in the same file, no need to build a heuristic that won't work for all situations anyway.
> 2. Check that other SW algorithm impls are not affected by the same issue.
>    For example you fixed the standard SW algorithm code, but the SSE2 variant has absolutely the same code that you missed in the fix

1. Actually we got exception not about allocating but "Access violation writing location". According to the internet this kind of exceptions cant be handled. This appear here SmithWatermanAlgorithm.cpp:253. As i understand the code this exception was thrown when we tried to expand `unsigned char *dir` so i called it "memory allocating" to simplify description. So as this exception impossible to handle i decided to find acceptable value for `matrixLength` to avoid this situation.
2.  This exception also appear if you will try to search human_t1 pattern in  itself, and with the same algorithm options as in workflow worker, you should change 'Save results' from 'Annotations' to 'Multiple alignment'. I've already added this condition to isValidParams() method.

yalgaer:
>     1. Actually we got exception not about allocating but "Access violation writing location". According to the internet this kind of exceptions cant be handled. This appear here SmithWatermanAlgorithm.cpp:253. As i understand the code this exception was thrown when we tried to expand `unsigned char *dir` so i called it "memory allocating" to simplify description. So as this exception impossible to handle i decided to find acceptable value for `matrixLength` to avoid this situation.

The 'dir' must always within matrix bounds according to the algorithm or you need to fix the algorithm.

I believe the  correct solution for this problem is a 1 line fix in the "malloc" line. 
Your variant is only suitable as an extra optimization, but not as a bullet-proof solution.

yalgaer (changes requested):
Fix the malloc line. Don't rely on workarounds when the correct solution is available.

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -132,12 +132,30 @@ void SmithWatermanAlgorithm::launch(const SMatrix &_substitutionMatrix,
...
+        int pat_n = patternSeq.length();
+        unsigned int dirn = (4 + pat_n + 3) >> 2;
+        memory = pat_n * 2 * sizeof(int) + pat_n * 0x80 + matrixLength * dirn;
+        quint_memory = (quint64)pat_n * 2 * sizeof(int) + (quint64)pat_n * 0x80 + (quint64)matrixLength * (quint64)dirn;
+    } else if (resultView == SmithWatermanSettings::ANNOTATIONS) {
+        int pat_n = patternSeq.length();
+        memory = pat_n * 3 * sizeof(int) + pat_n * 0x80;
+        quint_memory = (quint64)pat_n * 3 * sizeof(int) + (quint64)pat_n * (quint64)0x80;
+    }
+    if (quint_memory != memory) {
```
yalgaer: This check must be done in the same function with a malloc.
Repeating a complex computation in a separate place is incorrect: what if malloc code will change (will have +1 in the memory allocation size).

All other changes in parameters validation code are not related to the original issue: I would suggest removing them (keeping the original code) or filing another bug with a new problem description.

yalgaer (changes requested):
See comments.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -6064,6 +6064,43 @@ GUI_TEST_CLASS_DEFINITION(test_6862_1) {
...
 }
 
+GUI_TEST_CLASS_DEFINITION(test_6872) {
+    GTUtilsWorkflowDesigner::openWorkflowDesigner(os);
+
+    class FillTrimAndMapWizardWithHumanT1 : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus &os) override {
+            QWidget *wizard = GTWidget::getActiveModalWidget(os);
+            GTWidget::clickWindowTitle(os, wizard);
```
yalgaer: q: why do we need to click on the window title?

src/plugins/external_tool_support/src/blast_plus/align_worker_subtasks/BlastReadsSubTask.h
```diff
@@ -122,6 +123,7 @@ class BlastAndSwReadTask : public Task {
     DbiDataStorage *storage;
 
     BlastNPlusSupportTask *blastTask;
+    AbstractAlignmentTask *alignTask;
```
yalgaer: nit: init will nullptr in-place

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
```diff
@@ -174,8 +174,9 @@ void SmithWatermanAlgorithm::calculateMatrixForMultipleAlignmentResult() {
     n = pat_n * 2;
     unsigned int dirn = (4 + pat_n + 3) >> 2;
     unsigned int memory = n * sizeof(int) + pat_n * 0x80 + matrixLength * dirn;
-    int *buf, *matrix = (int *)malloc(memory);
-    if (matrix == nullptr) {
+    quint64 quint_memory = (quint64)n * sizeof(int) + (quint64)pat_n * (quint64)0x80 + (quint64)matrixLength * (quint64)dirn;
+    int* buf, * matrix = (int*)malloc(memory);
```
yalgaer: Does malloc accept 64-bit value?
I would suggest to mark algorithm as failed if required memory size is > some 32-bit constant like 1Gb.

yalgaer (changes requested):

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.cpp
rasputinkirill: image
image
As i can see on my MSVC size_t is unsigned int64. So it should be acceptable.

I don't think this is good idea to add any constant limits to allocating values. If malloc couldn't allocate desireble amount of memory following code will handle this situation. In that way we got flexible algorithm which can work on systems with different RAM volume.

yalgaer: if you expect that malloc works correctly with 64bit values and this scale is good for SW algorithm why do you have the  'quint_memory != memory' check below?

rasputinkirill:
I decided not to introduce a separate kind of error handling for matrix memory limit exceeding. I think the current information about the lack of memory will be enough for the user to adjust the parameters for running the algorithm.

rasputinkirill:
I used existing memory consumption calculator.

src/plugins/smith_waterman/src/SWAlgorithmTask.cpp
```diff
@@ -713,43 +713,52 @@ void PairwiseAlignmentSmithWatermanTask::setupTask() {
...
+                                                                            maxScore,
+                                                                            SmithWatermanSettings::MULTIPLE_ALIGNMENT);
             break;
         default:
             assert(0);
     }
-
-    t = new SequenceWalkerTask(c, this, tr("Smith Waterman2 SequenceWalker"));
-    addSubTask(t);
+    if (neededRam > SmithWatermanAlgorithm::MEMORY_SIZE_LIMIT) {
```
yalgaer: This is good, but is not enough. The resource level checks are good to avoid "bad" task starts, but we still need a check in place where memory is allocated.
Could you please add 2 more checks
1) Before the 'malloc': estimated memory size < 1Gb
2) After the 'malloc': the returned value is not nullptr.

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.h
```diff
@@ -48,6 +48,7 @@ class SmithWatermanAlgorithm {
     static const char UP;
     static const char LEFT;
     static const char DIAG;
+    static constexpr unsigned int MEMORY_SIZE_LIMIT = 1024;  //1GB
```
yalgaer: nit: MEMORY_SIZE_LIMIT -> .MEMORY_SIZE_LIMIT_MB

yalgaer (changes requested):

src/plugins/smith_waterman/src/SWAlgorithmTask.cpp
rasputinkirill: Done.

src/plugins/smith_waterman/src/SmithWatermanAlgorithm.h
rasputinkirill: Done.

yalgaer (approved):

DmitriiSukhomlinov (approved):

PR #561, opened 2021-09-01
 UGENE-7058: Remove deprecarted 3D view on Windows & Linux

yalgaer:
I checked non-deprecated 3D impl on all Linux & Windows and found no issues. 

I think it is a good time now to make this change because we will have 3+ months for testing before the next release.

+ I also fixed our GUI testing on Windows and now 3D structures are rendered correctly (no error labels is shown). Because of it I updated GUI tests.


evelinabis (changes requested):
Do you want 3D models to be displayed on Macs and modern machines? We took BALL a long time ago, which was using OpenGL version about 2.1. After that, OpenGL changed a lot, I'm not sure if there is backward compatibility of the 4th OpenGL with the 2nd (you have to look on the Internet). Our sources have not changed. Whether the BALL code has changed from the developers - you need to figure it out.
So, I have the following:
 * If I have an old Linux - almost probably 3D will be displayed there.
 * If I have a new Linux - I need the computer to support the old OpenGL functions. Whether he will do it, I do not know.
 * If I have Windows - there may not be OpenGL, but there may be Direct.
 * If I have a Mac - there may not be OpenGL, but there may be  Metal.

A cross-platform solution to this question is [QOpenGLClasses](https://doc.qt.io/qt-5/qglwidget.html#:~:text=Note%3A%20This%20class%20is%20part%20of%20the%20legacy%20Qt%20OpenGL%20module%20and%2C%20like%20the%20other%20QGL%20classes%2C%20should%20be%20avoided%20in%20the%20new%20applications.%20Instead%2C%20starting%20from%20Qt%205.4%2C%20prefer%20using%20QOpenGLWidget%20and%20the%20QOpenGL%20classes.). But in order to use them, it is not enough to write
glCreateShader (...)
You need to call the function from the class https://doc.qt.io/qt-5/qopenglfunctions.html#glCreateShader. Old features like [gluCylinder](https://github.com/ugeneunipro/ugene/blob/110dc5107b4f576bde2e150043e76a5d4ccd489f/src/plugins/biostruct3d_view/src/GraphicUtils.cpp#L94) are not on this list. Your changes do nothing at best.

Sorry for English. Old Linux means computer with old OpenGL supporting.

yalgaer:
>Do you want 3D models to be displayed on Macs and modern machines?...

The long story short:
The migration done many years ago from the deprecated package to the new package was not 100% perfect from the code qualify and Qt guidance point of views and I'm aware about these issues.

I can describe them in details  and I also fixed the problem of rendering on Windows test machines, but this is out of scope of this PR. I see that the new state is better then before: we do not test "it does not work" label anymore.

Some facts/answers to your questions:
1) The oldest Linux we officially support is Ubuntu 16.04 and our new OpenGL package works there. We do not support older versions of OSes because of other problems not related to this plugin.

2) The code I used was in production for many years for Mac OS. I checked that it works for all other OSes I have (Linux, Windows) understood all limitations and know how to fix them. These will be a separate big fix for all platforms at once in the common codebase and this is not a regression by any means: the old code/OpenGL API had no better compatibility and that was the reason why Qt officially replaced it with the new version of dynamically resolved OpenGL functions.

This way I think/hope the patch has no regressions and even fixes rendering on our Windows test machines: this is good.

We have enough time to address other issues before the release: this is also good. You can join with your patches and code.
Following the official Qt guidance to support different platforms this is the way we should follow. At least I know no other good option here.

evelinabis:
I don't think the openGL versions depend on the operating system version. Rather from machine. I'm not saying I know it for sure

yalgaer:
> I don't think the openGL versions depend on the operating system version. Rather from machine. I'm not saying I know it for sure

The supported subset of OpenGL is provided by the video card drivers provider. The worst case situation is an automatic fallback to the software emulation. The problem here is that software emulation usually is not installed with the OS by default, but Qt has it. This is how the new Qt API works when used correctly (and yes, we do not use it correctly).

Considering the set of OpenGL features UGENE uses today we should never fail with "Internal error" screen: we should use software rendering fallback automatically provided by Qt & its opengl32sw library.

PR #566, opened 2021-09-02
Make SAFE_POINTS crash release version of UGENE in GUI test mode

yalgaer:
12 XML + 5 GUI tests failed after this change. 
I will have to fix that tests first, before this patch can be submitted.


src/corelibs/U2Core/src/globals/U2SafePoints.h
```diff
@@ -75,7 +87,7 @@
 */
 #define SAFE_POINT_EXT(condition, extraOp, result) \
     if (Q_UNLIKELY(!(condition))) { \
-        assert(condition); \
+        U2::U2SafePoints::fail(); \
```
DmitriiSukhomlinov: I suggest to add the following message:
`QString("Trying to recover from the unknown error at %1:%2").arg(__FILE__).arg(__LINE__))`
It would be better than an empty line

DmitriiSukhomlinov (dismissed):

src/corelibs/U2Core/src/globals/U2SafePoints.h
yalgaer: Good idea. We can actually print the original condition here too. Fixed

DmitriiSukhomlinov (approved):

PR #581, opened 2021-09-10
Fix SAFE_POINTs in pcr GUI tests

yalgaer:
These SAFE_POINTs were found by the https://github.com/ugeneunipro/ugene/pull/566 and blocks it from being submitted

The real fix is inside  "ExtractProductTask::run()" method. Everything else is no-op cleanup I did during debugging the issue.

The problem was with outputUrl that is never set when the task is run from WD

### DIFF
--- a/src/plugins/pcr/src/ExtractProductTask.cpp
+++ b/src/plugins/pcr/src/ExtractProductTask.cpp
@@ -61,9 +61,10 @@ QString ExtractProductTask::getProductName(const QString &sequenceName, qint64 s
         .arg(endPos);
 }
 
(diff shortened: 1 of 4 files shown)


src/plugins/pcr/src/ExtractProductTask.cpp
```diff
@@ -78,35 +79,35 @@ DNASequence ExtractProductTask::getProductSequence() {
 }
 
 DNASequence ExtractProductTask::extractTargetSequence() {
-    DNASequence result("", "");
```
yalgaer, line 81: reason for the rename: there was a conflict with a class field here

DmitriiSukhomlinov (approved):

PR #584, opened 2021-09-24
Logging and counter for UGENE crash report №24291

rasputinkirill:


DmitriiSukhomlinov (dismissed):

src/corelibs/U2Lang/src/support/WorkflowDebugMessageParser.cpp
```diff
@@ -60,6 +61,7 @@ void WorkflowDebugMessageParser::setSourceData(const QQueue<Message> &initSource
         QVariantMap data = message.getData().toMap();
         foreach (const QString &key, data.keys()) {
             if (key.contains(MESSAGE_PATH_DELIMETER)) {
+                coreLog.trace(QObject::tr("WorkflowDebugMessageParser.cpp:64 'key' = %1").arg(key));
```
yalgaer: nit: 
1) tr is not needed in trace() like this
2) use `__LINE__` when you need line number. Or remove the line number: the message is already unique for devs.

src/plugins/workflow_designer/src/WorkflowViewController.cpp
```diff
@@ -1399,6 +1399,10 @@ void WorkflowView::localHostLaunch() {
         GCounter::increment(meta.name, "WDSample:run");
     }
 
+    if (WorkflowSettings::isDebuggerEnabled()) {
+        GCOUNTER(cvar, "Worklow started with enabled debugger");
```
yalgaer: Call GCounter::increment() directly. See the code above,

yalgaer (changes requested):
See comments

yalgaer (approved):

PR #586, opened 2021-10-04
UGENE-7432. Implement GUI tests

evelinabis:


src/corelibs/U2Gui/src/options_panel/GroupOptionsWidget.h
```diff
@@ -69,7 +69,7 @@ class U2GUI_EXPORT GroupOptionsWidget : public QWidget {
     QVBoxLayout *mainLayout;
 
     static const int TITLE_HEIGHT = 30;
-    static const int MIN_WIDGET_WIDTH = 440;
+    static const int MIN_WIDGET_WIDTH = 480;  // Minimum for the tab of sequence viewer "PCR Primer Design".
```
yalgaer: This will change width for all tabs in all views (assembly, sequence, msa, mca, tree)  regardless if pcr plugin is loaded or not.
This should be a dynamic per-tab variable set by the plugin.

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.h
```diff
@@ -0,0 +1,173 @@
...
+        InsertTo insertTo     = InsertTo::Backbone5;
+        int      backbone5Len = 0,
+                 backbone3Len = 0;
+    };
+    // Primer search areas for insert.
+    struct SearchArea {
+        SearchArea(U2Range<int> r, bool selectManually = false);
+
+        U2Range<int> region;
+        bool         selectManually = false;
```
yalgaer: Please reformat the file with the clangformat style

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.h
```diff
@@ -0,0 +1,173 @@
...
+            Yes
+        };
+        BadBackboneFiller(GUITestOpStatus& os, Button btn);
+        BadBackboneFiller(GUITestOpStatus& os, CustomScenario* scenario);
+    };
+
+    static void openTab(GUITestOpStatus& os);
+
+    // Sets forward (if direction=Direct) or reverse (if direction=Complementary) user primer.
+    static void setUserPrimer(GUITestOpStatus& os, const QString& primer, U2Strand::Direction direction);
```
yalgaer: Use const & for all non-standard types (U2Strand::Direction)

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+#define GT_METHOD_NAME "scrollToWidget"
+static void scrollToWidget(GUITestOpStatus& os, const QWidget* scrollTo, const QWidget* tab) {
+    GT_CHECK(scrollTo != nullptr, "Can't scroll to nullptr")
+    const auto isWidgetFullyVisible = [](const QWidget* w) {
+        const QRegion visibleRegion = w->visibleRegion();
+        return !visibleRegion.isEmpty() && visibleRegion.boundingRect().height() == w->height();
+    };
+    const auto moveMouseAndScroll = [](QWidget* moveMouseTo, bool scrollUp) {
+        GTMouseDriver::moveTo(GTWidget::getWidgetCenter(moveMouseTo));
+        const int scrollCoef = 20;
```
yalgaer: Dont use consts for locals (you do not use them consistently in this CL and UGENE does not use them across the code too)

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+        GTMouseDriver::scroll(scrollUp ? scrollCoef : -scrollCoef);
+        GTGlobals::sleep(200);
+    };
+
+    if (isWidgetFullyVisible(scrollTo)) {
+        return;  // Nothing to do.
+    }
+
+    tab = (tab == nullptr ? getTabWgt(os) : tab);
```
yalgaer: nit: braces are not needed

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+/**
+ * Checks if row index is correct for table. Indices are numbered starting from 0. If the table is empty, returns false,
+ * if the index is non-negative and less than the number of rows in the table, returns true. The table should not be
+ * nullptr.
+ */
+#define GT_METHOD_NAME "isTableIndexCorrect"
+static bool isTableIndexCorrect(GUITestOpStatus& os, int ind, const QTableWidget* table) {
+    GT_CHECK_RESULT(table != nullptr, "Table is nullptr", false)
+    const int tableRowCount = table->rowCount();
+    GT_CHECK_RESULT(0 <= ind && ind < tableRowCount, QString("Invalid index of '%1': expected 0<=i<%2, current i=%3").
```
yalgaer: Don't use Yoda style. Write code in the same order as you read it.

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+    GTWidget::click(os, btn);
+}
+
+/**
+ * Searches for 2 paired spinboxes and sets their values according to range. Parent is needed to refine the search for
+ * spinboxes. Doesn't open the tab.
+ */
+static void setRange(GUITestOpStatus& os, const QString& minSbName, const QString& maxSbName, U2Range<int> range,
+                     const QWidget* parent) {
+    if (QSpinBox* min = GTWidget::findSpinBox(os, minSbName, parent)) {
```
yalgaer: min is never nullptr here

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+    scrollToWidget(os, parent, tab);
+    setRange(os, "sbMinRequireGibbs",         "sbMaxRequireGibbs",         params.gibbsFreeEnergy, parent);
+    setRange(os, "spMinRequireMeltingTeml",   "spMaxRequireMeltingTeml",   params.meltingPoint,    parent);
+    setRange(os, "spMinRequireOverlapLength", "spMaxRequireOverlapLength", params.overlapLength,   parent);
+}
+
+void GTUtilsPcrPrimerDesign::setParametersToExcludeInWholePrimers(GUITestOpStatus& os,
+                                                                  ParametersToExcludeInWholePrimers params) {
+    const QWidget* tab = getTabWgt(os);
+    QWidget* const parent = GTWidget::findWidget(os, "wgtParameters2ExcludeInWholePrimers", tab);
```
yalgaer: remove const  "spam" from this line

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+    return GTWidget::findWidget(os, "wgtGenerateSequence", tab == nullptr ? getTabWgt(os) : tab);
+}
+
+// Return the generated sequence table. Doesn't open the tab. Returns nullptr in case of an error.
+static QTableWidget* getGenSeqTable(GUITestOpStatus& os, const QWidget* tab) {
+    return GTWidget::findExactWidget<QTableWidget*>(os, "twGeneratedSequences", getGenSeqWgt(os, tab));
+}
+
+// Return the table of result sequences. Doesn't open the tab. Returns nullptr in case of an error.
+static QTableWidget* getResultTable(GUITestOpStatus& os, const QWidget* tab) {
```
yalgaer: In case of an error sets the error into "os" or some other state holder structure. Never return nullptr from getters like in all other selector methods. Use "flags" to adjust the behaviour

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+    }
+    clickButtonInGenSeqWgt(os, buttonText);
+}
+
+void GTUtilsPcrPrimerDesign::setParametersOfPrimingSequences(GUITestOpStatus& os,
+                                                             const ParametersOfPrimingSequences& params) {
+    const QWidget* tab = getTabWgt(os);
+    const QWidget* parent = GTWidget::findWidget(os, "wgtParametersOfPrimingSequences", tab);
+    scrollToWidget(os, parent, tab);
+    setRange(os, "sbMinRequireGibbs",         "sbMaxRequireGibbs",         params.gibbsFreeEnergy, parent);
```
yalgaer: Reformat the file according to the clangformat

yalgaer (changes requested):

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
```diff
@@ -0,0 +1,367 @@
...
+#include "primitives/GTWidget.h"
+
+namespace U2 {
+using namespace HI;
+
+#define GT_CLASS_NAME "GTUtilsPcrPrimerDesign"
+
+static const QString badBackboneDialogName = "UnwantedStructuresInBackboneDialog";
+
+// Parent of most widgets used in this class. Doesn't open the tab. Returns nullptr in case of an error.
```
DmitriiSukhomlinov, line 48: `A parent of the most part of widgets...`

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
DmitriiSukhomlinov: I disagree. This kind of the order texting is much more clear and understandable. This is kinda a ray, the largest numbers are righter. The analogue of this sign:

`0 <= ind < tableRowCount`

This king of style makes it easy to imagine a picture like this:
image

DmitriiSukhomlinov (approved):

src/plugins/GUITestBase/src/GTUtilsPcrPrimerDesignForDnaAssembly.cpp
evelinabis: > Never return nullptr from getters like in all other selector methods.

And what to return if widget is not found?

src/corelibs/U2Gui/src/options_panel/OptionsPanelWidget.cpp
```diff
@@ -42,12 +42,17 @@ OptionsScrollArea::OptionsScrollArea(QWidget *parent)
                   " }");
     setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Preferred);
     setMaximumWidth(GroupOptionsWidget::getMaxWidgetWidth());
+    setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
```
yalgaer: Do we really need this?

If a user has a large font set by default all our tabs won't fit to the pre-defined tab width and will not be accessible for a such user at all.

You can reproduce this problem by setting "Text Scaling Factor" to 2x on Linux desktop.

src/corelibs/U2Gui/src/options_panel/OptionsPanelWidget.cpp
```diff
@@ -164,6 +169,9 @@ GroupOptionsWidget *OptionsPanelWidget::createOptionsWidget(const QString &group
 
     optionsWidgets.insert(0, groupWidget);
 
+    optionsScrollArea->setMinWidth(mainWidget->minimumWidth() + 20);
```
yalgaer: nit: add a comment what is '20'

src/corelibs/U2Gui/src/options_panel/OptionsPanelWidget.h
```diff
@@ -42,6 +42,12 @@ class OptionsScrollArea : public QScrollArea {
 
     /** Ensures that the scroll area would have an appropriate width */
     virtual QSize sizeHint() const;
+
+    /** Set the minimum width of a child. */
```
yalgaer: nit: Set -> Sets

yalgaer: QWidget already has "setMinimumWidth" and "setMinWidth" sounds ambiguous. Could you please name it more explicitly, like setMinimumWidthHint or similar.

yalgaer (changes requested):
Looks like a patch for the minWidth is very small. I can review it as a part of the current PR but would suggest forking it into an independent patch targeted directly to master.

In this case as an option panel test candidate for a different size we can select any existing tab that can benefit from size change.

evelinabis:
#601 #605

PR #590, opened 2021-10-11
UGENE-7448 Can't export sequence of selected annotations as amino

DmitriiSukhomlinov:


src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/ExportSequencesDialogFiller.cpp
```diff
@@ -149,6 +149,12 @@ void ExportSequenceOfSelectedAnnotationsFiller::commonScenario() {
 
     GTGlobals::sleep(200);
 
+    QCheckBox* translateButton = qobject_cast<QCheckBox*>(GTWidget::findWidget(os, "translateButton"));
+    CHECK_SET_ERR(translateButton != nullptr, "translateButton not found!");
+    GTCheckBox::setChecked(os, translateButton, translate);
+
+    GTGlobals::sleep(200);
```
yalgaer: I think we do not need any sleep methods here. The old sleeps are not cleaned up (a legacy code) but the new ones should not be introduced for no reason.

src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/ExportSequencesDialogFiller.cpp
```diff
@@ -149,6 +149,12 @@ void ExportSequenceOfSelectedAnnotationsFiller::commonScenario() {
 
     GTGlobals::sleep(200);
 
+    QCheckBox* translateButton = qobject_cast<QCheckBox*>(GTWidget::findWidget(os, "translateButton"));
+    CHECK_SET_ERR(translateButton != nullptr, "translateButton not found!");
```
yalgaer: Use GTWidget::findCheckBox or findExactWidget and you do not need to check for nullptr

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -993,6 +993,34 @@ GUI_TEST_CLASS_DEFINITION(test_7438) {
...
+                                                      ExportSequenceOfSelectedAnnotationsFiller::Fasta,
+                                                      ExportSequenceOfSelectedAnnotationsFiller::SaveAsSeparate,
+                                                      0,
+                                                      true,
+                                                      false,
+                                                      GTGlobals::UseMouse,
+                                                      true));
+    GTUtilsDialog::waitForDialog(os, new PopupChooserByText(os, { "Export", "Export sequence of selected annotations..." }));
+    GTMouseDriver::click(Qt::RightButton);
+
```
yalgaer, line 1241: add waitTaskFinished, otherwise the test may be unstable

src/plugins/dna_export/src/ExportSequenceTask.cpp
```diff
@@ -580,7 +580,7 @@ U2Sequence ExportAnnotationSequenceSubTask::importAnnotatedSeq2Dbi(const SharedA
             CHECK_OP(os, U2Sequence());
             currentRegionLength += chunkContent.length();
         }
-        CHECK_EXT(currentRegionLength == annotatedRegion.length,
+        CHECK_EXT(currentRegionLength == (nullptr != ei.aminoTT ? annotatedRegion.length / 3 : annotatedRegion.length),
```
yalgaer: Do not use Yoda style code

yalgaer (changes requested):
The fix is good, minor comments only.

src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/ExportSequencesDialogFiller.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/ExportSequencesDialogFiller.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
DmitriiSukhomlinov: OK

src/plugins/dna_export/src/ExportSequenceTask.cpp
DmitriiSukhomlinov: What does that mean in this case?

yalgaer: change `nullptr != ei.aminoTT `->` ei.aminoTT != nullptr` so you can read the code from left to right as a normal text

smlnkcv (dismissed):

src/plugins/dna_export/src/ExportSequenceTask.cpp
DmitriiSukhomlinov: Done

smlnkcv (dismissed):

yalgaer (changes requested):
I checked the result build and found one more issue with amino translation support in the fixed method:
1) Open 5mb.fa.gz (see the JIRA issue)
2) Create an annotation covering the whole file.
3) Export translation of the annotation.
4) Compare the last 3 symbols of the annotation and the exported sequence: the symbols will not match:
- The last 3 selected amino symbols are: RAG
- The last 3 symbols in the exported file: GRG

For the UI test I would suggest to generate the original file and avoid keeping it in the SVN/git.

DmitriiSukhomlinov:
I fixed the problem (with the solution I suggested the best one in this case) and fixed  the GUI test.

The above-mentioned problem has appeared because of the weird system of exporting chank-by-chunk, where the chunk has a size, which isn't divided by 3 without remainder (4194305 - the remainder  of the division by 3 is 2). There is an offset has appeared because of this remainder and translation has been moved. The chunk size is 4194303 now (The remainder of the division by 3 is 0).

yalgaer (changes requested):
Any test for the chunked mode that was fixed in the last commit?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -993,6 +993,38 @@ GUI_TEST_CLASS_DEFINITION(test_7438) {
     CHECK_SET_ERR(selectedRect.bottom() == 17, "Illegal end of the selection: " + QString::number(selectedRect.bottom()));
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7448) {
+    QString filePath = sandBoxDir + "test_7448.fa";
+    GTFile::copy(os, dataDir + "samples/Genbank/murine.gb", filePath);
```
yalgaer: Why do you want to copy the file in the test?

DmitriiSukhomlinov: My bad, fixed

DmitriiSukhomlinov:
Fixed, look again

yalgaer (changes requested):
I tested the patch and found the following problem:

1) Open 5mb.fa file from the GUI test.

2) Create a whole sequence annotation on the reverse complementary strand: CDS

3) Copy and paste somewhere translation of the annotation: (Ctrl-T): the sequence starts with TPA and ends with ILD: same as you see in the Sequence Viewer. This is the correct result.

4) Try "Export sequence of selected annotations" + amino form: the exported sequence is started with VYR and ends with GLV.

DmitriiSukhomlinov:
I have spent a bit time to figure this problem out. It happens when the reverse-complement region should be exported and the exported sequence is split into several chunks. In this case, we need to get the last chunk, add to it the chunk before last and so on, but we have the algorithm which exports chunks from the first one in any case.

This is not my case, I do not want to do this in this issue.

yalgaer:
> This is not my case, I do not want to do this in this issue.

Here is the problem:
1. Without this fix UGENE produces an error message and does not provide any result to user.
2. With this fix the UGENE does not produce any error message and silently provides incorrect result to user.

Do you think that this is acceptable?

yalgaer (changes requested):
See the comment above.

DmitriiSukhomlinov:
> > This is not my case, I do not want to do this in this issue.
> 
> Here is the problem:
> 
> 1. Without this fix UGENE produces an error message and does not provide any result to user.
> 2. With this fix the UGENE does not produce any error message and silently provides incorrect result to user.
> 
> Do you think that this is acceptable?

I have fixed this problem. See the code

src/plugins/dna_export/src/ExportSequenceTask.cpp
```diff
@@ -553,18 +553,31 @@ U2Sequence ExportAnnotationSequenceSubTask::importAnnotatedSeq2Dbi(const SharedA
     importer.startSequence(os, resultDbiRef, U2ObjectDbi::ROOT_FOLDER, ad->name, false);
     CHECK_OP(os, U2Sequence());
 
-    foreach (const U2Region &annotatedRegion, ad->location->regions) {
+    bool isReverseComplement = ad->getStrand().isCompementary() && nullptr != ei.complTT;
```
yalgaer: 1. rename to "isReverseComplementAndTranslate"
2. do not use Yoda style

src/plugins/dna_export/src/ExportSequenceTask.cpp
```diff
@@ -553,18 +553,31 @@ U2Sequence ExportAnnotationSequenceSubTask::importAnnotatedSeq2Dbi(const SharedA
...
+        // owervise the translation wiil be shifted on this not-skipped characters.
+        qint64 offset = 0;
+        if (isReverseComplement) {
+            offset = annotatedRegion.length % 3;
+        }
+        qint64 currentChunkSize = MAX_CHUNK_LENGTH;
+        for (qint64 pos = annotatedRegion.startPos; pos < annotatedRegion.endPos(); pos += currentChunkSize) {
+            currentChunkSize = qMin(MAX_CHUNK_LENGTH, annotatedRegion.endPos() - pos);
+            if (chunkContentList.isEmpty() && offset != 0) {
+                currentChunkSize = offset;
```
yalgaer: offset is always between 0 and 2: the chunk size of length <=2 can't be translated, so the code when currentChunkSize= offset is no-op. Why not start with the `pos = annotatedRegion.startPos+offset` to avoid the complexity inside of the loop?

src/plugins/dna_export/src/ExportSequenceTask.cpp
```diff
@@ -553,18 +553,31 @@ U2Sequence ExportAnnotationSequenceSubTask::importAnnotatedSeq2Dbi(const SharedA
     importer.startSequence(os, resultDbiRef, U2ObjectDbi::ROOT_FOLDER, ad->name, false);
     CHECK_OP(os, U2Sequence());
 
-    foreach (const U2Region &annotatedRegion, ad->location->regions) {
+    bool isReverseComplement = ad->getStrand().isCompementary() && nullptr != ei.complTT;
+    for (const U2Region &annotatedRegion : qAsConst(ad->location->regions)) {
         qint64 currentRegionLength = 0;
-        for (qint64 pos = annotatedRegion.startPos; pos < annotatedRegion.endPos(); pos += MAX_CHUNK_LENGTH) {
-            const qint64 currentChunkSize = qMin(MAX_CHUNK_LENGTH, annotatedRegion.endPos() - pos);
+        QList<QByteArray> chunkContentList;
```
yalgaer: The whole idea behind MAX_CHUNK_SIZE is to avoid keeping  large memory blocks in RAM and flush these blocks into DB (importer.addBlock) independently. This way the algorithm never uses memory above MAX_CHUNK_SIZE bytes

With your impl the algorithm will require as much of memory as the annotated sequence size: up to int64.

yalgaer (changes requested):

src/plugins/dna_export/src/ExportSequenceTask.cpp
DmitriiSukhomlinov: You lose first 1 or 2 bases

src/plugins/dna_export/src/ExportSequenceTask.cpp
DmitriiSukhomlinov: Yep, I get it

src/plugins/dna_export/src/ExportSequenceTask.cpp
DmitriiSukhomlinov: 1. Well, actually, "reverse-complement" and "translate" - two different options. You can "reverse-complement", but not "translate" and vice-versa.
2. OK

src/plugins/dna_export/src/ExportSequenceTask.cpp
```diff
@@ -553,18 +553,65 @@ U2Sequence ExportAnnotationSequenceSubTask::importAnnotatedSeq2Dbi(const SharedA
...
+        if (isReverseComplement) {
+            offset = annotatedRegion.length % 3;
+            pos = annotationEndPos - MAX_CHUNK_LENGTH;
+            if (pos < annotatedRegion.startPos + offset) {
+                pos = annotatedRegion.startPos + offset;
+            }
+        } else {
+            pos = annotatedRegion.startPos;
+        }
+        auto condition = [&]() -> bool {
```
yalgaer, line 578: nit: You share external state with a big 'condition' & 'change' functions that internally have a clear 
"if (isReverseComplement) else" code separation.
I would suggest running 2 loops: direct and reverse instead to make the code easier to read.

yalgaer (changes requested):
I checked the code in runtime and found that the following scenario does not work correctly:

1) Open human_T1.fa
2) Create an annotation with the following location: 10..15
3) Export it either as nucleic or amino

The following error occurs:

`'Opening view for document: [unknown]' task failed: Subtask {Adding document to project: [unknown]} is failed: Subtask {Export annotations} is failed: Subtask {Extract annotation regions} is failed: Sequences of the selected annotations can't be exported. At least one of the annotations is out of boundariesDocument was removed`

This is a regression introduced in the patch, now even non-translated sequence can't be exported.

src/plugins/dna_export/src/ExportSequenceTask.cpp
DmitriiSukhomlinov: I have been thinking about the way to solve this problem and I like the lambda-solutions more, I don't see a point in replacing it

DmitriiSukhomlinov:
The GUI problem's fixed.  There are several tests already exists on this problem, so I haven't written another one:
 - GUITest_common_scenarios_project_sequence_exporting_test_0002
 - GUITest_common_scenarios_project_sequence_exporting_test_0003
 - GUITest_regression_scenarios_test_6628_4

yalgaer (changes requested):
I checked the patch and found the following bug:

- Open humanT1.fa
- Create an annotation: CDS join(10..16,18..20)
- Check the highlighted text on the screen:  SP and S
- Copy annotation translation with Ctrl+T: clipboard buffer has SPS
- Now export translation of the annotation and you will see SPV but not SPS.

DmitriiSukhomlinov:
> I checked the patch and found the following bug:
> 
> * Open humanT1.fa
> * Create an annotation: CDS join(10..16,18..20)
> * Check the highlighted text on the screen:  SP and S
> * Copy annotation translation with Ctrl+T: clipboard buffer has SPS
> * Now export translation of the annotation and you will see SPV but not SPS.

The problem has appeared in the "Copy annotation amino acid" feature (which has been implemented by you, by the way), not in the "Export sequence of selected annotations" feature:
image

In this case, I need to fix everything in UGENE about it. I suppose, it's enough for this issue - I fixed the all cases and created 3 GUI tests. Please, stop torturing me just because you want and can do it.

yalgaer:
> In this case, I need to fix everything in UGENE about it. I suppose, it's enough for this issue - I fixed the all cases and created 3 GUI tests. Please, stop torturing me just because you want and can do it.

This is not a bug in the original code: check the annotation highlighting for joins and copy-annotation-translation feature: these 2 features work synchronously.

Here is the explanation (you can find it in the Genbank/EMBL manual for joins):
Joins are usually used to show that 2 regions (exons, coding sequences) are separated as a result of splicing.

Before these regions are translated into amino they are joined.  You can't translate them independently: it may break a translation frame.

I'm not asking you to fix any code outside of your PR at all, but only to implement the feature you decided to implement correctly.

DmitriiSukhomlinov:
> 

OK, I see, that, as usual, you are right by default and won't listen to anyone but yourself. My suggestion to split two code-independent issues are rejected just because you want. I got tired of this. Either close the PR or approve it.

yalgaer:
I think that my comments/bug scenarios above have nothing personal, I only reported issues found during the testing.
In the opposite I see a lot of negative personal messages.

I do not know how to split the issue: all we need is a simple and correct solution (10-15 lines of code):
- The current state: the feature does not work.
- The fixed state: the feature works.

Anything between, like "the feature mostly works but does not always produce a correct/reliable result" is not acceptable.

I'm closing the issue, because the current patch produces incompatible results with the current UGENE functionality: translation highlighting or copying into a clipboard.

PR #592, opened 2021-10-14
UGENE-7451. Crash on a right-click on a recent files item

yalgaer:


src/ugeneui/src/welcome_page/WelcomePageMdiController.cpp
```diff
@@ -71,11 +73,15 @@ void WelcomePageMdiController::sl_onMdiClose(MWMDIWindow *mdi) {
 }
 
 void WelcomePageMdiController::sl_onRecentChanged() {
```
DmitriiSukhomlinov, line 75: This function is defined as a slot, but it's not connected to anything. It's just called as a simple function (line 67). It this case, this fix looks a bit weird. You need either use it's slot possibilities  or make it a simple function.
1. Use this function as a slot.
Create some signal to connect it to:
```
signals:
    void si_onRecentChanged():
```
Make a connection in the constructor:
```
    connect(this, SIGNAL(si_onRecentChanged()), this, SLOT(sl_onRecentChanged()));
```
Emit a signal on the line 67:
```
    emit si_onRecentChanged():
```
And remove the all stuff from this function you have added in this commit, cuz the connection provides an asynchronous  call.

2. Leave the function as you have done it, but move it from `private slots` to `private` and rename (remove `sl_`).

I like the first option better, cuz `QTimer::singleShot(0` looks a bit strange for me.

DmitriiSukhomlinov (changes requested):
Look at the comment.

src/ugeneui/src/welcome_page/WelcomePageMdiController.cpp
yalgaer: > This function is defined as a slot, but it's not connected to anything. 

Main.cpp:
QObject::connect(pli, SIGNAL(si_recentListChanged()), wpc, SLOT(sl_onRecentChanged()));
?

PR #593, opened 2021-10-16
UGENE-7321. Increase width of option panel to make the tab fully visible

evelinabis:


DmitriiSukhomlinov:
And again - merge it to the `pcr_primer_design` branch, not to the `UGENE-7432`, then rebase `UGENE-7432` with `pcr_primer_design`

evelinabis:
#602

PR #600, opened 2021-10-21
UGENE-7460. Disable MSA overview widget by default for very large ali…

yalgaer:
…gnments


src/corelibs/U2View/src/ov_msa/MSAEditor.cpp
```diff
@@ -520,6 +520,20 @@ QWidget *MSAEditor::createWidget() {
     return ui;
 }
 
+void MSAEditor::initActions() {
+    MaEditor::initActions();
+
+    // Disable overview for very large alignments by default.
+    // UGENE's MSA overview calculation is too slow and having the overview ON when multiple large files
+    // are opened can lead to 100% CPU block.
+    quint64 maxAlignmentSizeToHaveOverviewByDefault = 1000 * 1000;
```
DmitriiSukhomlinov: `static constexpr`

PR #606, opened 2021-11-02
UGENE-7444. After reopening the PCR Primer Design tab, the "Select manually" button is pressed, but should not be

evelinabis:


src/plugins/pcr_primer_design_for_dna_assembly/src/PCRPrimerDesignForDNAAssemblyOPWidget.cpp
```diff
@@ -133,6 +133,9 @@ PCRPrimerDesignForDNAAssemblyOPWidget::PCRPrimerDesignForDNAAssemblyOPWidget(Ann
     connect(tbRightAreaSelectManually, &QAbstractButton::clicked, this, &PCRPrimerDesignForDNAAssemblyOPWidget::sl_selectManually);
     U2WidgetStateStorage::restoreWidgetState(savableWidget);
 
+    tbLeftAreaSelectManually->setChecked(false);
+    tbRightAreaSelectManually->setChecked(false);
```
DmitriiSukhomlinov: This solutions works, but it's a crutch. The root of the problem isn't here.

`PCRPrimerDesignForDNAAssemblyOPSavableTab` is a class to save state of the tab's elements when the tab closes and restore these states when you open the tab again (`getChildValue` saves the state on closing, `setChildValue` sets the state on opening). There is a possibility to set a custom behavior for some widgets if you want them saving a bit differently. You need to avoid saving state in case of `tbLeftAreaSelectManually` and `tbRightAreaSelectManually`.

DmitriiSukhomlinov (changes requested):
A major comment

src/plugins/pcr_primer_design_for_dna_assembly/src/PCRPrimerDesignForDNAAssemblyOPSavableTab.cpp
```diff
@@ -49,6 +49,8 @@ QVariant PCRPrimerDesignForDNAAssemblyOPSavableTab::getChildValue(const QString&
         return QVariant::fromValue<ResultTableData>(productTable->getPCRPrimerProductTableData());
     } else if (userPrimerLineEdit != nullptr) {
         return QVariant::fromValue<UserPimerLineEditResult>(userPrimerLineEdit->getData());
+    } else if (childId.endsWith("AreaSelectManually")) {
```
DmitriiSukhomlinov: Argh, don't think, that I am too picky or trying to pick on you, but it's not good again, If someone someday for some reason change the widget IDs, this code will stop working and it'll be very difficult to find that it's because the part of ID which's been changed is used here. This is the case when the best solution is already suggested and you don't need to invent something new. Just do as it's done several lines above - use `qobject_cast`

DmitriiSukhomlinov (changes requested):
Again, the major comment

src/plugins/pcr_primer_design_for_dna_assembly/src/PCRPrimerDesignForDNAAssemblyOPSavableTab.cpp
```diff
@@ -45,10 +45,13 @@ PCRPrimerDesignForDNAAssemblyOPSavableTab::~PCRPrimerDesignForDNAAssemblyOPSavab
 QVariant PCRPrimerDesignForDNAAssemblyOPSavableTab::getChildValue(const QString& childId) const {
     ResultTable* productTable = qobject_cast<ResultTable*>(getChildWidgetById(childId));
     UserPimerLineEdit* userPrimerLineEdit = qobject_cast<UserPimerLineEdit*>(getChildWidgetById(childId));
+    QToolButton* toolButton = qobject_cast<QToolButton*>(getChildWidgetById(childId));
     if (productTable != nullptr) {
         return QVariant::fromValue<ResultTableData>(productTable->getPCRPrimerProductTableData());
     } else if (userPrimerLineEdit != nullptr) {
         return QVariant::fromValue<UserPimerLineEditResult>(userPrimerLineEdit->getData());
+    } else if (toolButton != nullptr) {
+        return {};
```
DmitriiSukhomlinov, line 54: `{}` -  remove

DmitriiSukhomlinov (approved):

PR #608, opened 2021-11-03
UGENE-7388. SAFE_POINT if select Copy (custom format) action on aligm…

yalgaer:
…ent with gaps


smlnkcv (dismissed):
There is no SAFE_POINT after fix

DmitriiSukhomlinov (dismissed):
lgtm

PR #610, opened 2021-11-08
UGENE-7468. Migrate Stockholm format to the correct Unicode support

yalgaer:


src/corelibs/U2Formats/src/StockholmFormat.cpp
```diff
@@ -94,491 +75,331 @@ enum AnnotationTag {
...
-    }
-    virtual ~Annotation() {
-    }
-};
-//#=GF annotations
-struct FileAnnotation : public Annotation {
-    FileAnnotation(AnnotationTag t, QString v)
-        : Annotation(FILE_ANNOTATION, t, v) {
+/** Internal representation of the annotation used internally during parsing. */
+class StockholmAnnotation {
```
DmitriiSukhomlinov: Please, do something about thisc class - split declarations/definitions and move to separate files

src/corelibs/U2Formats/src/StockholmFormat.cpp
```diff
@@ -94,491 +75,331 @@ enum AnnotationTag {
...
-            assert(SS_CONS == ann->tag || RF == ann->tag);
-            Annotation *nAnn = findAnnotation(ann_list, COLUMN_ANNOTATION, ann->tag);
-            if (nullptr != nAnn) {
-                nAnn->val.append(ann->val);
-                delete ann;
+/**
+ * A list of StockholmAnnotation with an adjusted 'add' method.
+ * Auto-deletes all annotations when destroyed.
+ */
+class AnnotationBank {
```
DmitriiSukhomlinov, line 100: The same

src/corelibs/U2Formats/src/StockholmFormat.cpp
```diff
@@ -94,491 +75,331 @@ enum AnnotationTag {
...
-    virtual ~Annotation() {
-    }
-};
-//#=GF annotations
-struct FileAnnotation : public Annotation {
-    FileAnnotation(AnnotationTag t, QString v)
-        : Annotation(FILE_ANNOTATION, t, v) {
+/** Internal representation of the annotation used internally during parsing. */
+class StockholmAnnotation {
+public:
```
DmitriiSukhomlinov: If it's fully-public class, so, it's deffinetely a structure. I would suggest use this type in this case

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Formats/src/StockholmFormat.cpp
yalgaer: Why? There is nothing bad in local API defined locally. 

All modern languages I know use this style and the archaic C/C++ way of declaration/definition separation is a solution for a different problem: the one we do not have in this case.

src/corelibs/U2Formats/src/StockholmFormat.cpp
yalgaer: The same.

src/corelibs/U2Formats/src/StockholmFormat.cpp
yalgaer: Fixed: made it a struct.

PR #613, opened 2021-11-09
UGENE-7390 Crash if try to execute on Windows workflow example with Spades

DmitriiSukhomlinov:
This problem appeared in [this](https://github.com/ugeneunipro/ugene/commit/7585dd663100422191f08e7d2fda5ffa569eebd6#diff-9050ea6dec59fdab16a88a27315f37a6afb65eb950d52fcf532a297822477656R184) commit. For some reason which, I suppose, @rasputinkirill do not remember already, hi changed `a->getId()` to `a->getProto()->getId()` which isn't correct - it's possible, that the prototype ID isn't the same that the actor ID. 

I couldn't find any review of the above-mentioned comment, so, I think, this mistake could be unnoticeable.


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -861,6 +863,68 @@ GUI_TEST_CLASS_DEFINITION(test_7388) {
...
+    GTUtilsDialog::waitForDialog(os, new AppSettingsDialogFiller(os, new SetSpades()));
+    GTMenu::clickMainMenuItem(os, { "Settings", "Preferences..." }, GTGlobals::UseMouse);
+
+    // 2. Open WD
+   GTUtilsWorkflowDesigner::openWorkflowDesigner(os);
+
+    using TrimmomaticAddSettings = QPair<TrimmomaticDialogFiller::TrimmomaticSteps, QMap<TrimmomaticDialogFiller::TrimmomaticValues, QVariant>>;
+    QList<TrimmomaticAddSettings> steps;
+    steps.append(TrimmomaticAddSettings(TrimmomaticDialogFiller::TrimmomaticSteps::ILLUMINACLIP, {} ));
+    TrimmomaticDialogFiller* filler = new TrimmomaticDialogFiller(os, steps);
```
yalgaer: nit: type is typed twice in the line (use auto)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -861,6 +863,68 @@ GUI_TEST_CLASS_DEFINITION(test_7388) {
...
+    // 3. Open the "De novo assemble Illumina SE reads" sample
+    // 4. Set "human_T1.fa" as input
+    // 5. Click "Next"
+    // 6. Set the "ILLUMINACLIP" Trimmomatic step
+    // 7. Click "Next"
+    // 8. Click "Next"
+    // 9. Click "Apply"
+    class ProcessWizard : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus& os) {
```
yalgaer: nit: add override

yalgaer (commented):
Looks good, will approve after all tests are finished.

smlnkcv (dismissed):
No crash if select any path to the Spades in the External tool.
Error in WD: [ERROR][21:06] External tool "SPAdes" is invalid. UGENE may not support this version of the tool or a wrong path to the tools is selected

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
DmitriiSukhomlinov: The line's been removed

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
DmitriiSukhomlinov: OK

yalgaer (approved):

PR #621, opened 2021-11-14
UGENE-7263. Remove deprecated NGS tools from release bundle

yalgaer:
This package removes tools from the release bundle only. We are still keeping the tools in the codebase & build them on every commit.

The subset of tools was selected using the statistics analysis made 3 month ago: 0 active users, 0 active tests, 0 active plans.
Once we get a supporter and plans for any of these tools we can return them back.

+ Note:  no tools were removed from Windows (the most popular distribution). The change affects Linux and Mac only (there is a problem with code-signing of some  very old tools on Mac).


DmitriiSukhomlinov (changes requested):
Woooow! I hardly disagree about the removing metagenomics from UGENE! Just about several weeks ago a user connected to us and did work with these plugins! We shall not remove them!

PR #622, opened 2021-11-15
UGENE-7472. Build tree dialog does not check if output location is wr…

yalgaer:
…itable


src/corelibs/U2Gui/src/util/SaveDocumentController.cpp
```diff
@@ -143,6 +142,17 @@ DocumentFormatId SaveDocumentController::getFormatIdToSave() const {
     return formatsInfo.getIdByName(currentFormat);
 }
 
+QString SaveDocumentController::getValidatedSaveFilePath(U2OpStatus &os) const {
+    QString fileName = getSaveFileName();
+    CHECK_EXT(!fileName.isEmpty(), os.setError(tr("Output file name is empty")), "");
+
+    CHECK_EXT(FileAndDirectoryUtils::canWriteToPath(fileName),
+              os.setError(tr("File location is not writable: %1").arg(fileName)),
```
DmitriiSukhomlinov: You could use `L10N::errorOpeningFileWrite`

src/corelibs/U2View/src/phyltree/CreatePhyTreeDialogController.cpp
```diff
@@ -171,16 +166,19 @@ bool CreatePhyTreeDialogController::checkMemory() {
 
 void CreatePhyTreeDialogController::initSaveController(const MultipleSequenceAlignmentObject *msaObject) {
     SaveDocumentControllerConfig config;
-    config.defaultFileName = GUrlUtils::getNewLocalUrlByExtension(msaObject->getDocument()->getURLString(), msaObject->getGObjectName(), ".nwk", "");
+    GUrl msaDocumentUrl = msaObject->getDocument()->getURL();
+    QString saveDirPath = QFileInfo(msaDocumentUrl.getURLString()).absolutePath();
+    if (!FileAndDirectoryUtils::canWriteToPath(saveDirPath)) {
+        saveDirPath = AppContext::getAppSettings()->getUserAppsSettings()->getDefaultDataDirPath();
```
DmitriiSukhomlinov, line 181: I don't like that you've changed the result directory without saying a word to user. Maybe, it's better ask him either them agree to save to the default data dir or not?

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2View/src/phyltree/CreatePhyTreeDialogController.cpp
yalgaer: I see you point, lets check what better alternatives do we have:

1) Ask every time a user clicks "Build Tree"
In this case we will end-up always asking a user who opens files from some public/shared resource which is read-only for the user. This is what we usually try to avoid (remember all that 'roll-file-name' utilities, etc).
or
2) Work as before: set an incorrect folder and show the error each time user clicks "Build"?

The "DataDir" in UGENE is a special folder (like  "My Documents" ) designed to be used by UGENE as the default place to store results. The current solution looks like a valid trick to me: do not set a folder we know is invalid by default and do not interrupt user workflow by asking about a folder in 100% of cases.

src/corelibs/U2Gui/src/util/SaveDocumentController.cpp
yalgaer: Fixed

PR #625, opened 2021-11-16
UGENE-7465 [Wd] Crash on execution workflow sample "MUSCLE alignment"

rasputinkirill:


src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
```diff
@@ -1804,6 +1804,7 @@ void GUITestBasePlugin::registerTests(UGUITestBase *guiTestBase) {
     REGISTER_TEST(GUITest_regression_scenarios::test_7451);
     REGISTER_TEST(GUITest_regression_scenarios::test_7456);
     REGISTER_TEST(GUITest_regression_scenarios::test_7460);
+    REGISTER_TEST_T(GUITest_regression_scenarios::test_7465, minutes(14));
```
yalgaer: If the task fails with out-of-memory error why do you need such a big timeout?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1280,6 +1281,31 @@ GUI_TEST_CLASS_DEFINITION(test_7460) {
     GTUtilsTaskTreeView::waitTaskFinished(os, 10000);  // Check that there is no long-running active tasks.
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7465) {
+    //1. Open workflow sample "Align sequences with MUSCLE"
+    //Expected state: wizard has appeared.
+    class AlignSequencesWithMuscleWizardFiller : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus &os) {
```
yalgaer: add override

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1280,6 +1281,31 @@ GUI_TEST_CLASS_DEFINITION(test_7460) {
...
+
+            //2. Set file with many (~1200) sequences as input file and run workflow
+            GTUtilsWizard::setInputFiles(os, {{QFileInfo(testDir + "_common_data/fastq/lymph.fastq").absoluteFilePath()}});
+            //GTUtilsWizard::setParameter(os, "Reference", QFileInfo(testDir + "_common_data/fastq/lymph.fastq").absoluteFilePath());
+            GTUtilsWizard::clickButton(os, GTUtilsWizard::Next);
+            GTUtilsWizard::clickButton(os, GTUtilsWizard::Run);
+        }
+    };
+    GTUtilsDialog::waitForDialog(os, new WizardFiller(os, "Align Sequences with MUSCLE Wizard", new AlignSequencesWithMuscleWizardFiller));
+    GTUtilsWorkflowDesigner::openWorkflowDesigner(os);
```
yalgaer, line 1453: Do you really need a workflow or even a GUI test? Once we test an algorithm only why not using XML?

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
```diff
@@ -64,16 +64,18 @@ int MuscleParallelTask::estimateMemoryUsageInMb(const MultipleSequenceAlignment
     }
     std::sort(rowsLengths.begin(), rowsLengths.end(), std::greater<int>());
 
-    qint64 usedBytes = 0;
+    quint64 usedBytes = 0;
```
yalgaer: Don't use unsigned types. It gives more problem than solutions

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
```diff
@@ -64,16 +64,18 @@ int MuscleParallelTask::estimateMemoryUsageInMb(const MultipleSequenceAlignment
...
     int availableThreads = workpool->nThreads;
     for (int i = 0; i < rowsLengths.size() && availableThreads > 0; i++) {
         for (int j = 0; j < rowsLengths.size() && availableThreads > 0; j++, availableThreads--) {
             usedBytes += qint64((rowsLengths[i] + 1025)) * (rowsLengths[j] + 1025);
         }
     }
     const int maxInt = std::numeric_limits<int>::max();
+    quint64 m_DistsUsedBytes = sizeof(float) * rowsLengths.size() * rowsLengths.size();
     usedBytes = usedBytes >= 0 ? usedBytes : maxInt;
```
yalgaer: you changed usedBytes to unsigned, yes?

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
```diff
@@ -64,16 +64,18 @@ int MuscleParallelTask::estimateMemoryUsageInMb(const MultipleSequenceAlignment
...
-    qint64 usedBytes = 0;
+    quint64 usedBytes = 0;
     int availableThreads = workpool->nThreads;
     for (int i = 0; i < rowsLengths.size() && availableThreads > 0; i++) {
         for (int j = 0; j < rowsLengths.size() && availableThreads > 0; j++, availableThreads--) {
             usedBytes += qint64((rowsLengths[i] + 1025)) * (rowsLengths[j] + 1025);
         }
     }
     const int maxInt = std::numeric_limits<int>::max();
+    quint64 m_DistsUsedBytes = sizeof(float) * rowsLengths.size() * rowsLengths.size();
```
yalgaer: Do not use Hungarian notation.

yalgaer (changes requested):

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
rasputinkirill: Opening this file is the most time consuming operation. After that Muscle task measure memory cnsumption.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: I think it is possible to make this test as XML, but i dont see any advantages between this options, XML won't be faster.

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
rasputinkirill: What kind of problem? I think more naturally to use unsigned because memory consumption value can't be negative value.

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
rasputinkirill: Yes.

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
rasputinkirill: Tried to save original variable name, moved it to comment.

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
yalgaer: We can optimize opening of the file, because the opening itself is out of the test scope.
How to optimize it:
1) Use a FASTA file. Such files have a very good parser
or
2) Use UGENEDB file. Such  files are opened instantly.
or
3) Generate a FASTA file and open it as an alignment.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: You said "Opening this file is the most time consuming operation"
Is it because of the slow FASTQ or a slow UI?

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
yalgaer: In this case usedBytes >= 0 is always true? Right?

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
yalgaer: See the problem in your code below. 
As another example you can guess other popular modern languages that separate signed and unsigned type. I think this is an artifact inherited from the old low-level C language and it should be avoided to have no extra type conversions in the code. 
Another example: see QT collections or strings: they all use signed types for size/capacity, etc.

yalgaer (changes requested):
Added comments and suggestions

yalgaer:
Closing as "inactive". Please re-open when you are ready.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: Because of slow FASTQ reading.

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
rasputinkirill: 1st option choosen
2nd is impossible - ugene can't convert msa object to ugenedb
3rd i didn't  try, but as i see it is the same as 1st, because we need to load FASTA

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: done

yalgaer (changes requested):
Please add per-comment answers what was fixed and what was not

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
rasputinkirill: quint64 is back

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
rasputinkirill: This method part is rewritten.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1437,6 +1437,30 @@ GUI_TEST_CLASS_DEFINITION(test_7463) {
...
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7465) {
+    //1. Open workflow sample "Align sequences with MUSCLE"
+    //Expected state: wizard has appeared.
+    class AlignSequencesWithMuscleWizardFiller : public CustomScenario {
+    public:
+        void run(HI::GUITestOpStatus &os) override {
+            QWidget *wizard = GTWidget::getActiveModalWidget(os);
+            GTWidget::clickWindowTitle(os, wizard);
```
yalgaer: nit: is this line really needed? the widget is already active?

src/plugins_3rdparty/umuscle/src/MuscleParallel.cpp
```diff
@@ -71,9 +71,10 @@ int MuscleParallelTask::estimateMemoryUsageInMb(const MultipleSequenceAlignment
             usedBytes += qint64((rowsLengths[i] + 1025)) * (rowsLengths[j] + 1025);
         }
     }
-    const int maxInt = std::numeric_limits<int>::max();
-    usedBytes = usedBytes >= 0 ? usedBytes : maxInt;
-    return qMin(usedBytes / 1024 / 1024, qint64(maxInt));
+    //Check memory consumption for m_Dists in distfunc.cpp
+    qint64 mDistsUsedBytes = sizeof(float) * rowsLengths.size() * rowsLengths.size();
```
yalgaer: int * int * int = int.
You case to the int64 is too late here

yalgaer (changes requested):

yalgaer (approved):

PR #641, opened 2021-12-01
Refactor MSA Editor Align menu: use sections for different algorithm …

yalgaer:
This is the next patch needed for UGENE-7499.

With UGENE-7499 we will have several new alignment algorithms, so we need 
1) structure them in the menu  
2) avoid boiler-plate code in "Align" menu construction in plugins that register new algorithms.


src/corelibs/U2Gui/src/ObjectViewModel.h
```diff
@@ -246,6 +264,13 @@ protected slots:
     bool closing;
     QList<GObjectViewObjectHandler *> objectHandlers;
     OptionsPanel *optionsPanel;
+    QList<GObjectViewActionsProvider *> actionsProviders;
+};
+
+class U2GUI_EXPORT GObjectViewActionsProvider {
```
DmitriiSukhomlinov, line 270: Why do you need a separate class here?

DmitriiSukhomlinov (changes requested):
See the comment

src/corelibs/U2Gui/src/ObjectViewModel.h
yalgaer: Separate from the GObjectViewWindowContext?

I think the reason is the same why GObjectViewObjectHandler separated form  GObjectViewWindowContext   - to separate a clean functionality instead of passing a bloated GObjectViewWindowContext . 

The GObjectViewActionsProvider should not be necessary an instance of the GObjectViewWindowContext.

PR #645, opened 2021-12-07
UGENE-7505. Support sequence start/end centering by double click in MSA editor

yalgaer:
Request of the Vector Best.


src/corelibs/U2View/src/ov_msa/MaEditorNameList.cpp
```diff
@@ -275,10 +275,30 @@ void MaEditorNameList::paintEvent(QPaintEvent *) {
...
+    }
+    QWidget::mouseDoubleClickEvent(e);
+}
+
 void MaEditorNameList::keyPressEvent(QKeyEvent *e) {
     int key = e->key();
-    bool isShiftPressed = e->modifiers().testFlag(Qt::ShiftModifier);
+    Qt::KeyboardModifiers modifiers = e->modifiers();
+
+    if (modifiers == Qt::NoModifier && (key == Qt::Key_Space || key == Qt::Key_Enter || key == Qt::Key_Return)) {
```
DmitriiSukhomlinov: You have a `switch`-`case` below, it'd be better if you move it there

DmitriiSukhomlinov (changes requested):
See comment

src/corelibs/U2View/src/ov_msa/MaEditorNameList.cpp
yalgaer: Fixed.

PR #662, opened 2021-12-16
Cleanup GUI tests registration code. Re-enable some tests.

yalgaer:
What was done:
1) Some ignored/commented tests were re-enabled
2) Commented tests with no implementation were removed.
3) Test registration macro variants were reduced


DmitriiSukhomlinov (changes requested):
I'm sorry, but the code looks like you changed nothing but some renaming, which wasn't necessary to do at all. I see no sense in renaming `NOT_FOR_WINDOWS` to `LINUX_AND_MAC`

yalgaer:
> I'm sorry, but the code looks like you changed nothing but some renaming, which wasn't necessary to do at all. I see no sense in renaming `NOT_FOR_WINDOWS` to `LINUX_AND_MAC`

This is only a part of the item 3: Test registration macro variants were reduced.
Please check for changes  described in 1 and 2 also.

PR #663, opened 2021-12-17
UGENE-7392. Crash while reading annotations from emb file in WD

evelinabis:
`SAFE_POINT` is triggered at [doc->removeObject](https://github.com/ugeneunipro/ugene/pull/663/files#diff-15e3cf9a2861fb63a8c4be50d2ff5ef06e8678f7a9c6f5041546a1e04481db88L254). Because of this, `go` remains `doc`'s child. `go` is then removed twice: [explicitly](https://github.com/ugeneunipro/ugene/pull/663/files#diff-15e3cf9a2861fb63a8c4be50d2ff5ef06e8678f7a9c6f5041546a1e04481db88L256) and in the `doc`'s destructor as a `doc` child.


yalgaer (changes requested):
There is a problem with the current solution: a memory leak.

Objects in the document are never removed now. These can be non-annotation objects (sequence objects) or even annotation objects that were not re-used because of the 'merge' mode (a new annotation object was created to merge the old ones).

evelinabis:
> There is a problem with the current solution: a memory leak.
> 
> Objects in the document are never removed now. These can be non-annotation objects (sequence objects) or even annotation objects that were not re-used because of the 'merge' mode (a new annotation object was created to merge the old ones).

The `GObject` destructor is called for all objects in the document

evelinabis:
I have not found any memory leaks associated with my task. Please specify where the my leak is.

yalgaer:
> I have not found any memory leaks associated with my task. Please specify where the my leak is.

Here how it works:

1) You load a document with annotations: this document may contain sequences, annotation, other kind of objects.
There are 2 kind of objects created in this case a) in-memory objects and database (DB) records.

2) You re-use the annotation object only but set 'documentOwnsDbiResources' to false. This flag turns off DB resource management for all kind of objects, not only annotations.

As the result the document is removed and no DB resources are cleaned up: the sequence data in DB read together with the annotations will never be cleared.

yalgaer (changes requested):
Please check that there is not DB-level data leak after this patch.

yalgaer:
Is there any work planned for this issue? I'm going to close this PR as inactive after 14 days of inactivity. Feel free to re-open it when you resume.

evelinabis:
> Is there any work planned for this issue?

Yes.

> I'm going to close this PR as inactive after 14 days of inactivity. Feel free to re-open it when you resume.

In the near future I will write an answer about DB-memory leaks, please do not close the pull request, I remember about it.

evelinabis:
#### Possible implementations of the "Reading annotations" task:

1. [Reading into another db](https://github.com/ugeneunipro/ugene/blob/77781d4a79156441cd7a814f37a900a2ad75ea26/src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp#L234):
  1.1. read a document into session db;
  1.2. copy annotation tables from the session to the workflow db;
  1.3. the document will be deleted, the data in the session db will be deleted, the data in the workflow db will remain.
  1.4. Note: this option is different from other readers ([LoadSeqTask](https://github.com/ugeneunipro/ugene/blob/9ab0006505c6338e7e7cd548c248a607725b1fa5/src/plugins/workflow_designer/src/library/GenericReadWorker.cpp#L348), [LoadMSATask](https://github.com/ugeneunipro/ugene/blob/9ab0006505c6338e7e7cd548c248a607725b1fa5/src/plugins/workflow_designer/src/library/GenericReadWorker.cpp#L247), [ReadVariationTask](https://github.com/ugeneunipro/ugene/blob/9ab0006505c6338e7e7cd548c248a607725b1fa5/src/plugins/workflow_designer/src/library/ReadVariationWorker.cpp#L155)).
2. Removing the required annotations from the document:
  2.1. read the document into the workflow db;
  2.2. remove the required annotation tables from the document;
  2.3. the document will be deleted with all its objects and their data in the db, except for the annotation tables;
  2.4. annotation tables will remain in the workflow db until they are deleted with it at the end of the workflow.
  2.5. Note: this is the current version of the master branch and it is not possible because it crashes.
3. The document doesn't own its db resources:
  3.1. read the document into workflow db;
  3.2. `doc->setDocumentOwnsDbiResources(false)`;
  3.3. the document will be deleted, all data will remain in the db until it is deleted with it at the end of the workflow.
  3.4. Note: this is the current solution; unnecessary objects such as sequences will not be deleted immediately.
4. Removing unnecessary objects from the db:
  4.1. read the document into workflow db;
  4.2. `doc->setDocumentOwnsDbiResources(false)`;
  4.3. remove non-annotation objects from the db;
  4.4. the document will be deleted, all data will remain in the db until it is deleted with it at the end of the workflow.
  4.5. Note: I don't see a nice way to delete objects in the db without touching the document (as in option 2). However, it is possible.
5. Clone annotations:
  5.1. read the document into workflow db;
  5.2. clone annotations;
  5.3. the document will be deleted with all its objects and their data in the db, the copied annotation tables will remain until they are deleted with it at the end of the workflow.

The optimal variant by execution time is 3, by memory is 4. Which option to implement? If the choice is mine, I will choose 4.

yalgaer:
> The optimal variant by execution time is 3, by memory is 4. Which option to implement? If the choice is mine, I will choose 4.

AFAIU the problem is with the optimization when there is only 1 annotation object and we want to reuse it without copying. If we always copy it (like in a general case with merge or multiple objects) the problem does not exist.

I would suggest to remove the optimization now and follow the "mergeAnnotations && annsObjList.size() > 1)" route.

Deleting data at the end of workflow is not a good idea: a workflow should be able to process unlimited streams of data.

evelinabis:
> follow the "mergeAnnotations && annsObjList.size() > 1)" route.

The resulting file of this variant differs from the master branch variant.
Code (instead of existing loop):
```
DbiDataStorage *storage = context->getDataStorage();
auto &results_ = results;
bool isMerge = mergeAnnotations && annsObjList.size() > 1;

auto saveAnnots = [storage, &m, &results_](const QList<SharedAnnotationData> &annots, const QString &name) {
    const SharedDbiDataHandler tableId = storage->putAnnotationTable(annots, name);
    m[BaseSlots::ANNOTATION_TABLE_SLOT().getId()] = qVariantFromValue<SharedDbiDataHandler>(tableId);
    results_.append(m);
};

for (GObject *go : qAsConst(annsObjList)) {
    if (auto annsObj = dynamic_cast<AnnotationTableObject *>(go)) {
        if (!mergeAnnotations) {
            dataList.clear();
        }
        foreach (Annotation *a, annsObj->getAnnotations()) {
            dataList << a->getData();
        }
        if (!isMerge) {
            saveAnnots(dataList, go->getGObjectName());
        }
    }
}
if (isMerge) {
    saveAnnots(dataList, mergedAnnTableName);
}
```
Test:

- _common_data/cmdline/read-write/write/ann-ann.uwl
- input: _common_data/cmdline/read-write/write/ann1.gb
- mode: "Merge annotation tables from file".

Result of the master branch:
image
The result of the code above:
image

yalgaer:
You have an explicit group qualifier in "Two":
```
misc_feature    complement(10..12)
                     /ugene_name="Two"
                     /ugene_group="One"
```
So in the annotation view the annotation should be shown in the 'Two' group. Is the group qualifier preserved?

evelinabis:
> You have an explicit group qualifier in "Two":
> 
> ```
> misc_feature    complement(10..12)
>                      /ugene_name="Two"
>                      /ugene_group="One"
> ```
> 
> So in the annotation view the annotation should be shown in the 'Two' group. Is the group qualifier preserved?

Please clarify the question.

Upd. The group qualifier is not preserved in the scenario I described above if I run the current code from branch 7392.

yalgaer:
> Upd. The group qualifier is not preserved in the scenario I described above if I run the current code from branch 7392.

When we merge annotations all qualifiers from file should be preserved.
As far as I understand this is the current behavior and we should keep it.
So for the annotations:
```
     misc_feature    2..6
                     /ugene_name="One"
     misc_feature    complement(10..12)
                     /ugene_name="Two"
                     /ugene_group="One"
```
There should be 1 group created with 2 annotations inside.

yalgaer (changes requested):
Please check the commend and add a test for the scenario that causes the problem with groups.

You can add a check to your existing test that original document (or annotations in the original document) is equal to the result document.

yalgaer:
Closing as inactive. Please reopen when updated.

evelinabis:
Annotation tables are cloned in the database: clones remain in the database and go to the next workers, cloned tables and sequences from the document are cleared when `run` exits. This is for the case when the merge is not needed. When a merge is needed, the original implementation did not take into account groups, my implementation takes groups into account. Need testing functions to check the correctness of this, and test coverage.

src/plugins/CoreTests/src/AnnotationTableObjectTest.h
```diff
@@ -225,6 +225,17 @@ class GTest_CheckAnnotationsNamesInTwoObjects : public XmlTest {
     QString secondDocContextName;
 };
 
+// Compares all annotation tables in two documents. The order of tables is taken into account, the order of annotations
```
yalgaer: nit: 1 object == 1 annotation table. It is better to say "Compares annotations in two objects"

src/plugins/CoreTests/src/AnnotationTableObjectTest.cpp
```diff
@@ -885,6 +885,101 @@ Task::ReportResult GTest_CheckAnnotationsNamesInTwoObjects::report() {
...
+    }
+
+    secondDocContextName = el.attribute(VALUE_ATTR);
+    if (secondDocContextName.isEmpty()) {
+        failMissingValue(VALUE_ATTR);
+    }
+}
+
+Task::ReportResult GTest_CheckAnnotationsInTwoObjects::report() {
+    auto doc = getContext<Document>(this, docContextName),
```
yalgaer: nit: initialize each variable independently.

src/plugins/CoreTests/src/AnnotationTableObjectTest.cpp
```diff
@@ -885,6 +885,101 @@ Task::ReportResult GTest_CheckAnnotationsNamesInTwoObjects::report() {
...
+            }
+        return tables;
+    };
+
+    QString docName = doc->getName(),
+            docName2 = doc2->getName();
+    QList<AnnotationTableObject *> tables = getAnnotationTables(doc),
+                                   tables2 = getAnnotationTables(doc2);  // The order of the tables matters, the name
+                                                                         // and other characteristics do not matter.
+    if (hasError())
```
yalgaer: guard all blocks with brackets {}

src/plugins/CoreTests/src/AnnotationTableObjectTest.cpp
```diff
@@ -885,6 +885,101 @@ Task::ReportResult GTest_CheckAnnotationsNamesInTwoObjects::report() {
...
+                tables.push_back(table);
+            } else {
+                setError(QString("Failed to cast object '%1' of document '%2' to an annotation table")
+                             .arg(obj->getGObjectName(), document->getName()));
+                return {};
+            }
+        return tables;
+    };
+
+    QString docName = doc->getName(),
```
yalgaer: nit: For a symmetry it is better to have docName1/docName2 names, not docName/docName2

src/plugins/CoreTests/src/AnnotationTableObjectTest.cpp
```diff
@@ -885,6 +885,101 @@ Task::ReportResult GTest_CheckAnnotationsNamesInTwoObjects::report() {
...
+                         .arg(anns.size())
+                         .arg(tableName2, docName2)
+                         .arg(anns2.size()));
+            return ReportResult_Finished;
+        }
+        for (Annotation *a : qAsConst(anns)) {
+            auto foundAnn = std::find_if(anns2.begin(), anns2.end(), [a](Annotation *a2) {
+                return *a->getData() == *(a2)->getData() &&
+                       a->getGroup()->getGroupPath() == a2->getGroup()->getGroupPath();  // Case sensitive.
+                                                                         // AnnotationData::operator== may contain bugs.
```
yalgaer: Lets document/create this bug and point it here. Looks like something important

src/plugins/CoreTests/src/AnnotationTableObjectTest.cpp
```diff
@@ -885,6 +885,101 @@ Task::ReportResult GTest_CheckAnnotationsNamesInTwoObjects::report() {
...
+        if (anns.size() != anns2.size()) {
+            setError(QString("Table '%1' ('%2') has %3 annotations, table '%4' ('%5') has %6")
+                         .arg(tableName, docName)
+                         .arg(anns.size())
+                         .arg(tableName2, docName2)
+                         .arg(anns2.size()));
+            return ReportResult_Finished;
+        }
+        for (Annotation *a : qAsConst(anns)) {
+            auto foundAnn = std::find_if(anns2.begin(), anns2.end(), [a](Annotation *a2) {
```
yalgaer: nit: do not save a lot on variable names for a better readability.

src/plugins/CoreTests/src/AnnotationTableObjectTest.h
```diff
@@ -225,6 +225,17 @@ class GTest_CheckAnnotationsNamesInTwoObjects : public XmlTest {
...
+// in tables is not taken into account, groups of annotations are taken into account.
+class GTest_CheckAnnotationsInTwoObjects : public XmlTest {
+    Q_OBJECT
+    SIMPLE_XML_TEST_BODY_WITH_FACTORY(GTest_CheckAnnotationsInTwoObjects, "compare-annotations-in-two-objects")
+
+    ReportResult report() override;
+    QString docContextName;
```
yalgaer: nit: firstDocContextName. Or even more readable doc1ContentName

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
```diff
@@ -54,6 +57,38 @@ namespace {
...
+
+    for (AnnotationTableObject *table : qAsConst(tables)) {
+        AnnotationGroup *root = table->getRootGroup();
+        SAFE_POINT(root != nullptr, L10N::nullPointerError(QString("root '%1'").arg(table->getGObjectName())), {});
+
+        QStringList groupPaths;
+        root->getSubgroupPaths(groupPaths);
+        for (const QString &groupPath : qAsConst(groupPaths)) {
+            AnnotationGroup *group = root->getSubgroup(groupPath, false);
+            SAFE_POINT(group != nullptr, L10N::nullPointerError(QString("group '%1'").arg(groupPath)), {});
```
yalgaer: No need to localize SAFE_POINT messages.

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
```diff
@@ -54,6 +57,38 @@ namespace {
...
+                                                                    const QList<AnnotationTableObject *> &tables,
+                                                                    const QString &mergedTableName) {
+    U2OpStatusImpl os;
+    DbiOperationsBlock operationBlock(dbiRef, os);
+    CHECK_OP(os, {});
+    auto mergedAnnotationTable = std::make_unique<AnnotationTableObject>(mergedTableName, dbiRef);
+    AnnotationGroup *mergedRoot = mergedAnnotationTable->getRootGroup();
+
+    for (AnnotationTableObject *table : qAsConst(tables)) {
+        AnnotationGroup *root = table->getRootGroup();
```
yalgaer: nit: root -> rootGroup

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
```diff
@@ -54,6 +57,38 @@ namespace {
...
 const QString ANN_TABLE_DEFAULT_NAME("Unknown features");
+
+static std::unique_ptr<AnnotationTableObject> mergeAnnotationTables(const U2DbiRef &dbiRef,
+                                                                    const QList<AnnotationTableObject *> &tables,
+                                                                    const QString &mergedTableName) {
+    U2OpStatusImpl os;
+    DbiOperationsBlock operationBlock(dbiRef, os);
+    CHECK_OP(os, {});
+    auto mergedAnnotationTable = std::make_unique<AnnotationTableObject>(mergedTableName, dbiRef);
+    AnnotationGroup *mergedRoot = mergedAnnotationTable->getRootGroup();
```
yalgaer: nit: mergedRoot-> mergedRootGroup

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
```diff
@@ -95,13 +130,16 @@ void ReadAnnotationsWorker::onTaskFinished(Task *task) {
 
 void ReadAnnotationsWorker::sl_datasetEnded() {
     CHECK(datasetData.size() > 0, );
-    QList<SharedAnnotationData> anns;
+    QList<AnnotationTableObject *> anns;
```
yalgaer: With GAutoDeleteList you do not need to worry about qDeleteAll

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
```diff
@@ -242,30 +281,34 @@ void ReadAnnotationsTask::run() {
...
+    // If "SPLIT" -- transfer (write to the result) tables of annotations from files as is.
+    // If "MERGE" and there is only one annotation table, transfer the table as is.
+    // If "MERGE_FILES", transfer the file tables as is and merge them into ReadAnnotationsWorker::sl_datasetEnded().
+    // Otherwise ("MERGE" with several tables of annotations in one file), merge these tables into one.
+    if (mergeAnnotations != ReadAnnotationsProto::MERGE || annsObjList.size() == 1) {
+        for (GObject *gObject : qAsConst(annsObjList)) {
+            auto annotationTable = dynamic_cast<AnnotationTableObject *>(gObject);
+            CHECK_EXT(annotationTable != nullptr, setError(L10N::nullPointerError("annotationTable")), );
+
+            QScopedPointer<AnnotationTableObject> annotationTableClone(qobject_cast<AnnotationTableObject *>(annotationTable->clone(dbiRef, stateInfo)));
```
yalgaer: Why do you want to clone here? Clone() creates another instance in DB and if we have only 1 Annotation Object we can just return it, right?

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
```diff
@@ -242,30 +281,34 @@ void ReadAnnotationsTask::run() {
...
-        m[BaseSlots::ANNOTATION_TABLE_SLOT().getId()] = qVariantFromValue<SharedDbiDataHandler>(tableId);
-        results.append(m);
+    QList<AnnotationTableObject *> annotationTables;
+    for (GObject *gObject : qAsConst(annsObjList)) {
+        auto annotationTable = dynamic_cast<AnnotationTableObject *>(gObject);
+        CHECK_EXT(annotationTable != nullptr, setError(L10N::nullPointerError("annotationTable")), );
+        annotationTables += annotationTable;
     }
+    std::unique_ptr<AnnotationTableObject> mergedTable = mergeAnnotationTables(dbiRef, annotationTables, mergedAnnTableName);
+    SharedDbiDataHandler tableId = context->getDataStorage()->putAnnotationTable(mergedTable.get());
```
yalgaer: Here you need to delete all original annotation tables

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.h
```diff
@@ -76,7 +76,7 @@ class ReadAnnotationsWorkerFactory : public DomainFactory {
 class ReadAnnotationsTask : public Task {
     Q_OBJECT
 public:
-    ReadAnnotationsTask(const QString &url, const QString &datasetName, WorkflowContext *context, bool mergeAnnotations, const QString &mergedAnnTableName = QString());
+    ReadAnnotationsTask(const QString &url, const QString &datasetName, WorkflowContext *context, ReadAnnotationsProto::Mode mergeAnnotations, const QString &mergedAnnTableName = QString());
```
yalgaer: use & for all non-native types

yalgaer (changes requested):
Only 2 important issues about memory management.

Other issues are style-only.

src/plugins/CoreTests/src/AnnotationTableObjectTest.h
evelinabis: There can be several annotation tables in one document. By analogy with other checkers, I called "check in two objects", but it is correct to say "check in two documents"

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
evelinabis: How to transfer a table to another worker? It will be removed with the document.

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
evelinabis: It will be removed with the document.

src/plugins/CoreTests/src/AnnotationTableObjectTest.cpp
evelinabis: If there are identical qualifiers in the annotation, then equality will not be a symmetric operation: {q1, q1} == {q1, q2} will return true, {q1, q2} == {q1, q1} will return false

evelinabis:
- Current changes: merging annotations in the "MERGE_FILES" mode has been moved to a separate task, taking into account big data and without freezing GUI.

- As a reminder: annotation tables and sequences read in the document will be deleted along with it. Extracting annotation tables from the document is not possible as it is the current bug [7392](https://ugene.dev/tracker/browse/UGENE-7392). It could be established that the document does not own all of its db resources, but then there would be a db memory leak. The current solution clones the required document annotation tables and the original objects are removed as expected along with the document. Some other readers call
`doc->setDocumentOwnsDbiResources(false)`, that is, the db memory may be leaking.

- Question 1. `new AnnotationTableObject(...)` I do not see a way to check the correctness of creating an annotation table.
Question 2. There are no differences between datasets in the "Merge all annotation tables from <ins>dataset</ins>" mode. In the GUI UGENE, you can set several datasets, files from all <ins>dataset</ins>**s** will be merged into one file. In the console version, I do not see the ability to set several datasets.
Question 3. In the "Merge all annotation tables from dataset" mode, `DATASET_SLOT` contains name of the last dataset. In my opinion, it is logical to write something like "Merged dataset".
- Notes on my implementation: the `MergeAnnotationsTask` does not calculate the amount of memory required, although in theory it can be calculated inside the `ReadAnnotationsTask`. Perhaps it would be better to gradually merge the read annotation tables, and not all at once after reading the datasets (memory savings).

yalgaer:
I need 1-2 days to checkout and review this patch carefully. Will update until Monday.

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
yalgaer: All workers in workflow use the same share database and pass data to each other via QSharedDataPointer<DbiDataHandler>

Here how you create your annotations:
`
hints[DocumentFormat::DBI_REF_HINT] = QVariant::fromValue<U2DbiRef>(context->getDataStorage()->getDbiRef());
QScopedPointer<Document> doc(format->loadDocument(iof, url, hints, stateInfo));
`
Here how you clone it several lines below:
`
go->clone(context->getDataStorage(), stateInfo)))
`

It is the same database already: no need to clone, just re-use the object you loaded.
You should also ensure that the document won't destroy these objects in the database during own destruction: remove the objects from the document.

yalgaer (changes requested):
Lets fix the basic issues with memory management first and discuss advanced stuff in a separate PR.

src/plugins/workflow_designer/src/library/ReadAnnotationsWorker.cpp
evelinabis: How to delete objects from the document?

yalgaer: ```bool Document::removeObject``` ? 
If the existing API is not enough you can extend it.

yalgaer:
Closing as inactive. Please reopen when is ready for the review

evelinabis:
yalgaer:
> it should be possible to remove an object from a document in a simple way without deleting it in the database. removeObject takes flags on what to do when an object is removed. If these flags are not enough, then you can add one more. But the idea is simple: nothing should prevent you from taking an entire object from a document.

src/corelibs/U2Core/src/models/DocumentModel.h
```diff
@@ -340,9 +340,10 @@ enum DocumentObjectRemovalMode {
     DocumentObjectRemovalMode_Deallocate,  // ordinary removal: both object and its DB representation are deallocated
     DocumentObjectRemovalMode_OnlyNotify,  // fake removal: neither object nor its DB representation are deallocated.
     // Only signals about removal are emitted. The object itself remains to belong to the document.
-    DocumentObjectRemovalMode_Release  // fake removal: the same as DocumentObjectRemovalMode_OnlyNotify.
+    DocumentObjectRemovalMode_Release,  // fake removal: the same as DocumentObjectRemovalMode_OnlyNotify.
     // Additionally, the object is excluded from the document's child objects list.
     // External code has to handle a proper object deallocation.
+    DocumentObjectRemovalMode_Silent  // sudo removes an object from the doc, removes an object, leaving it in the db.
```
yalgaer: The better name would be a 'Detach'. 
What does 'sudo' in the comment mean?

yalgaer (commented):
Looks good, please re-run tests to get clean results and I will approve.

src/corelibs/U2Core/src/models/DocumentModel.h
evelinabis: sudo as in a Unix shell. It means "delete no matter what".

yalgaer (approved):

PR #665, opened 2021-12-17
UGENE-7276. After sorting and aligning, selection has changed

yalgaer:


src/corelibs/U2Core/src/util/MSAUtils.h
```diff
@@ -107,9 +107,20 @@ class U2CORE_EXPORT MSAUtils : public QObject {
      *  The optional 'prefix' field may be used to detect indexed rows. In this case all rows with no given prefix will not be renamed, but passed as is.
      *
      *  The method returns 'true' if all rows with a correct prefix were renamed.
+     *
+     *  TODO: this method is deprecated and all usages of it will be removed. Use 'restoreOriginalRowProperties' instead.
```
DmitriiSukhomlinov, line 111: As far as you mean this function deprecated mark it with the [corresponding attribute](https://en.cppreference.com/w/cpp/language/attributes/deprecated)

DmitriiSukhomlinov (changes requested):
See the comment

src/corelibs/U2Core/src/util/MSAUtils.h
yalgaer: It is a good idea to use this attribute, so it will produce warnings in case if deprecated API is used.
These warning will fail the build: we have a 0-warning state on Linux today and any new warning is treated as an error.

If you have any idea how to solve this problem I agree to start using [[deprecated]] attribute in the UGENE code in all places.

yalgaer: As an alternative we an use a pre-defined "TODO/deprecated" format in comments that can be easy locatable in IDE

PR #672, opened 2021-12-21
Fix for GUI test 1986

yalgaer:
The problem is that NCBI has added more results for mouse and the top of them do not have any FASTA sequence at all. 
I switched the test to another organism.

### DIFF
--- a/src/plugins/GUITestBase/src/runnables/ugene/ugeneui/NCBISearchDialogFiller.cpp
+++ b/src/plugins/GUITestBase/src/runnables/ugene/ugeneui/NCBISearchDialogFiller.cpp
@@ -312,49 +312,61 @@ void NcbiSearchDialogFiller::waitTasksFinish() {
 #undef GT_CLASS_NAME
 
 #define GT_CLASS_NAME "GTUtilsDialog::NCBISearchDialogSimpleFiller"
+
+NCBISearchDialogSimpleFiller::NCBISearchDialogSimpleFiller(HI::GUITestOpStatus &os,
+                                                           const QString &_query,
+                                                           bool _doubleEnter,
+                                                           int _resultLimit,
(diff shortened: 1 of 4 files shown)


src/plugins/GUITestBase/src/runnables/ugene/ugeneui/NCBISearchDialogFiller.cpp
```diff
@@ -312,49 +312,61 @@ void NcbiSearchDialogFiller::waitTasksFinish() {
...
+        // Click the first result. Original behavior.
+        GTWidget::click(os, resultsTreeWidget, Qt::LeftButton, QPoint(10, 35));
+    } else {
+        auto item = GTTreeWidget::findItem(os, resultsTreeWidget, resultToLoad, nullptr, 0, GTGlobals::FindOptions(true, Qt::MatchContains));
+        GTTreeWidget::click(os, item);
+    }
 
-    QDialogButtonBox *box = qobject_cast<QDialogButtonBox *>(GTWidget::findWidget(os, "buttonBox", dialog));
-    GT_CHECK(box != nullptr, "buttonBox is NULL");
+    auto box = GTWidget::findExactWidget<QDialogButtonBox *>(os, "buttonBox", dialog);
```
DmitriiSukhomlinov: You may use `GTUtilsDialog::clickButtonBox(os, dialog, QDialogButtonBox::Ok);`

DmitriiSukhomlinov (dismissed):

src/plugins/GUITestBase/src/runnables/ugene/ugeneui/NCBISearchDialogFiller.cpp
yalgaer: Good idea. Fixed

PR #675, opened 2021-12-23
Update windows release script to include Qt* lib debug symbols

yalgaer:
The original QT dlls do not contain debug info, so we need to take this info from the pdb files.
 This way "Build" phase should add these files for the "Release" script to the bundle.


DmitriiSukhomlinov (changes requested):
Why do we need this debug symbols? For crash reports?

yalgaer:
> Why do we need this debug symbols? For crash reports?

Yes, each release build provides a set of debug symbols (a symbols.tar.gz file) that is later used by the crash report handler.

PR #677, opened 2021-12-25
UGENE-7510. Error if select 'gpu-blastp' parameter in 'Request to Loc…

yalgaer:
…al Blast Database' dialog

One more attempt to complete https://ugene.dev/tracker/browse/UGENE-945 correctly: removing of gpu-blast

I also renamed BLAST+ to BLAST (this is how the original authors call the tool for many years after dropping the original BLAST)


smlnkcv (approved):

DmitriiSukhomlinov (changes requested):
I don't like the fact you've renamed blast+ to blast, cuz if you want to get rid of all "blast + " mentions in UGENE, you need also rename the code stuff. There are still a lot of tasks, which are called `BlastPlus...`, variables and so on. If you want to use "BLAST" instead - rename all of them and everywhere.

But, I suppose, you won't listen to my opinion)

yalgaer:
> I don't like the fact you've renamed blast+ to blast, cuz if you want to get rid of all "blast + " mentions in UGENE, you need also rename the code stuff. There are still a lot of tasks, which are called `BlastPlus...`, variables and so on. If you want to use "BLAST" instead - rename all of them and everywhere.

You are correct, I will provide a separate PR with code renaming only next. This patch was about "gpu-blast" & UI visible artifacts only.

> But, I suppose, you won't listen to my opinion)

This is not right. The logic is simple: if you really did not believe you would not spend your time writing  the review above.

PR #682, opened 2021-12-27
UGENE-7514. Use the original Genbank annotation names in UGENE

yalgaer:
Problem description:

Originally, up to 2017 UGENE used standard naming convention used by Genbank/EMBL/DDBJ (https://www.insdc.org/files/feature_table.html#7.2).
So any Genbank or EMBL created by UGENE was strictly complient with the standard and could be submitted to any database.

Next, some "name beautification" was made that made UGENE annotation names looks good inside UGENE,  but meaningless for external sources. For example "misc_feature" has became "Misc. Feature". In some cases it would be even possible to have both: "misc_feature" (loaded from the original document) and "Misc. Feature" (added by UGENE into the same document).

This patch is the first in the series to fix how UGENE works with annotations: it returns original names to Genbank annotations, so user can see the correctly named annotations right in the UGENE.

The next patches I plan are:
1) Group feature keys into 2 sets "Standard" and "Extended" in UI. This way a user will see when he stays within the standard and when is not.

2)  Allow a user to edit "Extended" feature types: add/remove his own variants.

3) Change the way feature types are stored/handled. It does not need a special numeric field the feature key plus a dedicated type-qualifier is enough for that. This way features saved to and re-loaded from Genbank format will never loose their "types" which are not stored to Genbank today at all.


DmitriiSukhomlinov (changes requested):
I don't like that I've got no idea how it supposed to be in the very end - have no idea how "Standard" and "Extended" options may be look like, how the GUI will change, etc. This feature exists in your head only - you even didn't created the issue, so this code flies in the space, not connected to anything. 

Please, at least connect it to some issue.

src/corelibs/U2Core/src/datatype/U2FeatureType.cpp
```diff
@@ -55,379 +55,209 @@ U2FeatureType U2FeatureTypes::getTypeByName(const QString &visualName) {
 }
 
 QList<U2FeatureTypes::U2FeatureTypeInfo> U2FeatureTypes::initFeatureTypes() {
-    QList<U2FeatureTypeInfo> featureTypeInfos;
```
DmitriiSukhomlinov, line 58: The whole changes in this file are so meaningless and so bloated, so I see no sense of doing that

yalgaer:
> Please, at least connect it to some issue.
https://ugene.dev/tracker/browse/UGENE-7514

src/corelibs/U2Core/src/datatype/U2FeatureType.cpp
yalgaer: Could you please elaborate on this?
The change reduces number of lines by 2, groups standard annotations and adds links to the standard.

PR #684, opened 2021-12-27
UGENE-7406 Crash when execute workflow Read annotation->Write annotation with different formats

rasputinkirill:


yalgaer (changes requested):
The patch looks more like a workaround for the problem in some other place. 

The original SAFE_POINT asserts that there should not be an annotation with an empty list of regions, but not with a zero-length region. How the list of regions becomes empty?

I think there is nothing really bad in zero-regions: even a zero length annotation has a valid start-pos: a position in the sequence (like a cut-site, insertion point).

yalgaer:
More about 0 length locations from here: https://www.insdc.org/files/feature_table.html#3.4

`A site between two adjoining nucleotides, such as endonucleolytic cleavage 
site, is indicated by listing the two points separated by a carat (^). The 
permitted formats for this descriptor are n^n+1 (for example 55^56), or, for 
circular molecules, n^1, where "n" is the full length of the molecule, ie 
1000^1 for circular molecule with length 1000.`

It looks like we do not support this correctly today and I'm not asking you to support it as a part of this bug. May be should have some special location for 0 or unknown length sequences like `<1..>1`

smlnkcv (commented):
Execution without crash.

yalgaer:
Closing as inactive.
Please re-open when you have an update

PR #704, opened 2022-01-09
Replace non-main threads ->click(), ->setText(), ->setChecked() with …

yalgaer:
…correct gui tests API calls

What is in the patch: I reviewed all places with direct  "->click()", "->setChecked", "->setText" calls on QT widgets from non-GUI threads and moved these calls into GUI threads.

I also cleaned up a little code style around such code.

### DIFF
--- a/src/plugins/GUITestBase/src/GTUtilsProjectTreeView.cpp
+++ b/src/plugins/GUITestBase/src/GTUtilsProjectTreeView.cpp
@@ -192,14 +192,13 @@ void GTUtilsProjectTreeView::scrollToIndexAndMakeExpanded(HI::GUITestOpStatus &o
 
     class MainThreadActionExpand : public CustomScenario {
     public:
-        MainThreadActionExpand(QTreeView *treeView, const QModelIndex &index)
-            : CustomScenario(), treeView(treeView), index(index) {
+        MainThreadActionExpand(QTreeView *_treeView, const QModelIndex &_index)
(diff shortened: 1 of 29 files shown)


src/plugins/GUITestBase/src/GTUtilsProjectTreeView.cpp
```diff
@@ -192,14 +192,13 @@ void GTUtilsProjectTreeView::scrollToIndexAndMakeExpanded(HI::GUITestOpStatus &o
 
     class MainThreadActionExpand : public CustomScenario {
     public:
-        MainThreadActionExpand(QTreeView *treeView, const QModelIndex &index)
-            : CustomScenario(), treeView(treeView), index(index) {
+        MainThreadActionExpand(QTreeView *_treeView, const QModelIndex &_index)
+            : treeView(_treeView), index(_index) {
```
DmitriiSukhomlinov, line 196: Not sure it's good - you've removed the initialization of the base class. The all base class members are invalid now.

DmitriiSukhomlinov (changes requested):
Check the message

src/plugins/GUITestBase/src/GTUtilsProjectTreeView.cpp
yalgaer: I think we already discussed it some PRs ago.  Please read the standard: https://isocpp.org/std/the-standard.

PR #705, opened 2022-01-12
UGENE-7516. Exclude List in MSA editor

yalgaer:
This is a minimal implementation that covers all client needs. 
Please read the detailed description and my opinion about it in JIRA.


src/corelibs/U2Core/src/globals/Task.h
```diff
@@ -589,8 +589,16 @@ class U2CORE_EXPORT TaskScheduler : public QObject {
 
 signals:
     void si_noTasksInScheduler();
+
+    // TODO: remove this signal from TaskScheduler. It is not scheduler responsibliity to track global UGENE state.
```
DmitriiSukhomlinov, line 593: I suppose, it's much better just to rename this signal, cuz it's pretty useful, but has a strange name. It's emitted when the all top level tasks and their subtasks are finished their jobs, so you may name it something like `si_allTopLevelTasksAreDone`

src/corelibs/U2Gui/src/util/GUIUtils.cpp
```diff
@@ -236,4 +236,16 @@ void GUIUtils::showMessage(QWidget *widgetToPaintOn, QPainter &painter, const QS
     return;
 }
 
+void GUIUtils::insertActionAfter(QMenu *menu, QAction *insertionPointMarkerAction, QAction *actionToInsert) {
+    SAFE_POINT(menu != nullptr, "menu is null", );
```
DmitriiSukhomlinov, line 240: Use `U2::L10n::nullPointerError`

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+    }
+
+    connect(AppContext::getTaskScheduler(), &TaskScheduler::si_stateChanged, this, &MsaExcludeListWidget::trackMsaObjectSaveTask);
+
+    toolbarLayout->addStretch();
+
+    loadExcludeList(true);
+}
+
+MsaExcludeListWidget::~MsaExcludeListWidget() {
```
DmitriiSukhomlinov, line 285: Destructor isn't suppose to do some extra work, it's for destruction only. This stuff should be done by using a special function [destroyed](https://doc.qt.io/qt-5/qobject.html#destroyed)

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+    if (isDirty) {
+        runSaveTask(excludeListFilePath);
+    }
+}
+
+int MsaExcludeListWidget::addEntry(const DNASequence &sequence, int excludeListRowId) {
+    SAFE_POINT(sequence.alphabet != nullptr, "Sequence must be fully defined!", 0);  // By default MSA row sequences have no alphabet. Catch this kind of error ASAP.
```
DmitriiSukhomlinov, line 292: `L10n`

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+    item->setData(LIST_ITEM_DATA_ROW_ID, computedExcludeListRowId);
+    nameListView->addItem(item);
+    sequenceByExcludeListRowId[computedExcludeListRowId] = sequence;
+    isDirty = true;
+    return computedExcludeListRowId;
+}
+
+int MsaExcludeListWidget::addMsaRowEntry(const MultipleAlignmentRow &row, int excludeListRowId) {
+    DNASequence sequence = row->getUngappedSequence();
+    sequence.alphabet = msaEditor->getMaObject()->getAlphabet();  // TODO: MSA row must return a fully defined sequence with a correct alphabet.
```
DmitriiSukhomlinov, line 305: The sequence alphabet could be differ from the whole alignment alphabet, isn't it?

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.h
```diff
@@ -0,0 +1,201 @@
...
+    int addEntry(const DNASequence &sequence, int excludeListRowId = 0);
+
+    /** Adds a name list entry from MSA row. Fixes non-defined DNA sequence alphabet of MSA row. */
+    int addMsaRowEntry(const MultipleAlignmentRow &row, int excludeListRowId = 0);
+
+    /** Remove sequence entries from the Exclude List model. */
+    void removeEntries(const QList<QListWidgetItem *> &items);
+
+    /** Triggers context menu for the name list. */
+    void showNameListContextMenu(const QPoint &pos);
```
DmitriiSukhomlinov: Why do you need the argument if it's unused?

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+        excludeListRowIdsMovedToMsa << getExcludeListRowId(listItem);
+    }
+    removeEntries(selectedItems);
+
+    int versionBefore = msaObject->getObjectVersion();
+    int insertionIndex = -1;  // Append by default.
+    if (!msaEditor->getSelection().isEmpty() && msaEditor->getRowOrderMode() == MaEditorRowOrderMode::Original) {
+        insertionIndex = msaEditor->getSelection().getRectList().last().bottom() + 1;
+    }
+    AddSequenceObjectsToAlignmentTask task(msaObject, sequences, insertionIndex, true);
```
DmitriiSukhomlinov, line 435: Not safe at all. What if in the future somebody will add some meaningful changes to the `prepare` part, for example?

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+        QSet<qint64> msaRowIdsAfter = msaObject->getRowIds().toSet();
+        for (int i = 0; i < maBefore->getNumRows(); i++) {
+            const MultipleAlignmentRow &row = maBefore->getRow(i);
+            if (!msaRowIdsAfter.contains(row->getRowId())) {
+                msaRows << row;
+            }
+        }
+        SAFE_POINT(msaRows.size() == undoRedoContext.excludeListRowIdsDelta.size(), "Failed to map Exclude List rows to Msa rows", )
+        for (int i = 0; i < msaRows.size(); i++) {
+            auto &msaRow = msaRows[i];
```
DmitriiSukhomlinov: Maybe `const auto&`?

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+            int excludeListRowId = undoRedoContext.excludeListRowIdsDelta[i];
+            addMsaRowEntry(msaRow, excludeListRowId);
+        }
+    } else {  // Remove rows from Exclude List.
+        QSet<int> rowIdsToRemove = undoRedoContext.excludeListRowIdsDelta.toSet();
+        QList<QListWidgetItem *> listItemsToRemove;
+        for (int rowIndex = nameListView->count(); --rowIndex >= 0;) {
```
DmitriiSukhomlinov, line 484: I suppose, it would be more clear to have a for like this:
`for (int rowIndex = nameListView->count() - 1; rowIndex >= 0; rowIndex--) {`
Otherwise it's some kind of yoda-style, which you do not like so much.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+
+    if (task->hasError()) {
+        stateLabel->setText(tr("Error loading exclude list file: ").arg(task->getError()));
+    } else if (!task->isCanceled()) {
+        QList<GObject *> objects = task->getDocument()->findGObjectByType(GObjectTypes::SEQUENCE);
+        nameListView->clear();
+        U2OpStatus2Log os;
+        for (auto object : qAsConst(objects)) {
+            auto sequenceObject = qobject_cast<U2SequenceObject *>(object);
+            SAFE_POINT(sequenceObject != nullptr, "Not a sequence object: " + object->getGObjectName(), );
```
DmitriiSukhomlinov, line 615: `L10n`

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
```diff
@@ -0,0 +1,593 @@
...
+
+void MsaExcludeListWidget::trackMsaObjectSaveTask(Task *task) {
+    auto saveMsaObjectTask = qobject_cast<SaveDocumentTask *>(task);
+    CHECK(saveMsaObjectTask != nullptr && saveMsaObjectTask->getDocument() == msaEditor->getMaObject()->getDocument(), );
+    CHECK(saveMsaObjectTask->isFinished() && !saveMsaObjectTask->getStateInfo().isCoR(), );
+    runSaveTask(excludeListFilePath);
+}
+
+QSize MsaExcludeListWidget::sizeHint() const {
+    return {500, 200};
```
DmitriiSukhomlinov, line 645: Add some description to these numbers, it's not clear where are they come from

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.h
```diff
@@ -0,0 +1,201 @@
...
+
+class LoadDocumentTask;
+class MSAEditor;
+class MaModificationInfo;
+class MsaExcludeListWidget;
+class MultipleAlignment;
+class MultipleAlignmentRow;
+
+/** MSA editor built-in support for "MSA Exclude List" widget. */
+class MsaExcludeListContext : public GObjectViewWindowContext {
```
DmitriiSukhomlinov, line 50: I don't like two pretty large classes in the same file, the file is very bloated

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.h
```diff
@@ -0,0 +1,201 @@
...
+ * The widget contains a data model for Exclude List and a limited Undo/Redo support for MSA:
+ * the Undo-Redo is supported only for the life-time of the widget: when Exclude List is closed the Undo/Redo stack of the widget is reset.
+ */
+class MsaExcludeListWidget : public QWidget {
+    Q_OBJECT
+public:
+    MsaExcludeListWidget(QWidget *parent, MSAEditor *msaEditor, MsaExcludeListContext *viewContext);
+
+    /** Triggers auto-save of the current exclude list state. */
+    ~MsaExcludeListWidget();
```
DmitriiSukhomlinov, line 97: As it was already told, the destructor shouldn't trigger anything, it should just deallocate memory

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Core/src/globals/Task.h
yalgaer: The fix is out of scope of this PR. I have only documented the lame code and will address it separately

src/corelibs/U2Gui/src/util/GUIUtils.cpp
yalgaer: SAFE_POINTS should not be translated. They do not exist from the user's POV.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: Why do destructors exist at all in C++ in that case? 
Are they really called at all unless mentioned explicitly (like constructors)? 
A lot of QT classes have destructors (QObject/QWidget(s)/QModel(s)), a lot of std classes also have destructors.

The destructor usage may be safe or unsafe like any other function and this is a goal of any code review: to reveal these unsafe places

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: Please forget about L10n in SAFE_POINTS. Such code will not pass a code reviews.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: I only documented the problem and will work on a solution in a separate PR.
The bug is that any sequence in the DB must have a valid alphabet. 

It is perfectly OK if a sequence has an alphabet compatible with an alignment and is not OK otherwise.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.h
yalgaer: Because this function is bound as a callback  for the "void customContextMenuRequested(const QPoint &pos)"
Do not have a consistent style in UGENE where to use it and where not.  If you will search for customContextMenuRequested() in UGENE you will find in 90% of cases it has the parameter.

I removed the param to keep moving forward.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: This is why I put TODO here. Will be addressed next in the separate PR to keep this more focused. Tests guards against the problem before it is fixed.

Repeating TODOs in the review have no value, isn't it?

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: It won't change anything: the list is not shared (is local). Added 'const' to move forward.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: A simple reg-ex in UGENE code shows >100 backward loops like this. This notion is equal to and  shorten than what you suggested, so I would keep it. 
Compare:
```for (int rowIndex = nameListView->count(); --rowIndex >= 0; ) ```

```for (int rowIndex = nameListView->count() - 1; rowIndex >= 0; rowIndex--) ```

However I do not mind if you use the other variant.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: L10n in SAFE_POINT will not pass code review.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: Already did it in the header file. Please check.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.h
yalgaer: Ok. I will not mind if you split them, but see no problems in files of 200-500 lines of code. 
I would definitely start splitting after 1000 lines of code.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.h
yalgaer: Please read the comment above. Destructor is a tool and there is nothing wrong with it when it is used correctly.
BTW You ask me to use a signal that emitted from the destructor itself.

DmitriiSukhomlinov (changes requested):
I've resolved the all minor comments, also I agree do not make a holy war about destructors, but you can't just add a problem code with the mark "TODO: fix it", don't you think so? I don't like the fact you've ignored the my comments about it :-/

yalgaer:
> but you can't just add a problem code with the mark "TODO: fix it", don't you think so? I don't like the fact you've ignored the my comments about it :-/

Agree and I do not add TODOs only. Here is the fix for the one of them:
https://github.com/ugeneunipro/ugene/pull/706

Another one in the Task Scheduler will be fixed tomorrow.

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: Fix for the TODO: https://github.com/ugeneunipro/ugene/pull/706 (quite a big update that worth to be tested separately)

src/corelibs/U2View/src/ov_msa/exclude_list/MsaExcludeList.cpp
yalgaer: Another reason why "destroyed" is not suitable here:
"destroyed" should be used when one objects wants to do something on another object destruction. Here we have a different situation: we have only 1 object: the widget. Subscribing to self has no meaning: the parent data is already destroyed when "destroyed" signal is emitted.

PR #712, opened 2022-01-15
Optimize GTUtilsAnnotationsTreeView methods

yalgaer:


yalgaer:
Closing for a while: the patch revealed too many problems in GUI tests

PR #713, opened 2022-01-17
UGENE-7518. Optimize Msa Overview computation time

yalgaer:


src/corelibs/U2Core/src/datatype/msa/MultipleSequenceAlignmentRow.cpp
```diff
@@ -557,4 +566,54 @@ MultipleAlignmentData *MultipleSequenceAlignmentRowData::getMultipleAlignmentDat
...
+}
+
+/**
+ * Size of the gapped cache: 200 symbols.
+ * Optimized for chars per screen and makes getChar() ~200x faster thanMsaRowUtils::charAt for sequential access.
+ */
+static constexpr int GAPPED_CACHE_SIZE = 200;
+
+char MultipleSequenceAlignmentRowData::getCharFromCache(int gappedPosition) const {
+    if (gappedPosition >= gappedCacheOffset && gappedPosition < gappedCacheOffset + gappedSequenceCache.size()) {
```
DmitriiSukhomlinov, line 608: Don't like this way of inequalities ordering. You need to look for some time on it to understand, how parts of the inequality relates to each other. Much better is to use the only `<` and `<=` signs and write variables from smallest to largest:
`gappedCacheOffset <= gappedPosition && gappedPosition < gappedCacheOffset + gappedSequenceCache.size()`

Here you may see easily, that `gappedPosition` should be between `gappedCacheOffset` and `gappedCacheOffset + gappedSequenceCache.size()`

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
```diff
@@ -432,4 +432,80 @@ void MsaRowUtils::removeTrailingGapsFromModel(qint64 length, QVector<U2MsaGap> &
...
+    CHECK(region.length > 0, {});
+    CHECK(!sequence.isEmpty(), QByteArray(region.length, U2Msa::GAP_CHAR));
+    const char *coreSequence = sequence.constData();
+    int coreLength = sequence.length();
+    QByteArray result;
+    result.reserve(region.length);
+    int gapIndex = 0;  // Current gap index.
+    int corePos = 0;  // Current offset in core sequence.
+    int pos = 0;  // Current gapped position in the row.
+    int regionEnd = (int)region.endPos();  // Last (exclusive) global gapped position to include into the result.
```
DmitriiSukhomlinov, line 451: `qint64` to `int` - data loss is possible

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
```diff
@@ -432,4 +432,80 @@ void MsaRowUtils::removeTrailingGapsFromModel(qint64 length, QVector<U2MsaGap> &
...
+                int nChars;
+                int corePartOffset;
+                if (region.startPos <= pos) {  // Add a prefix or whole core sequence region.
+                    nChars = qMin(corePartLength, regionEnd - pos);
+                    corePartOffset = 0;
+                } else if (regionEnd > pos + corePartLength) {  // Add suffix of the core sequence region.
+                    nChars = pos + corePartLength - region.startPos;
+                    corePartOffset = corePartLength - nChars;
+                } else {  // Add middle part of the core sequence region.
+                    SAFE_POINT(region.startPos > pos && regionEnd <= pos + corePartLength, "Invalid middle region coordinates", {};)
```
DmitriiSukhomlinov, line 474: The same

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
```diff
@@ -432,4 +432,80 @@ void MsaRowUtils::removeTrailingGapsFromModel(qint64 length, QVector<U2MsaGap> &
...
+    int corePos = 0;  // Current offset in core sequence.
+    int pos = 0;  // Current gapped position in the row.
+    int regionEnd = (int)region.endPos();  // Last (exclusive) global gapped position to include into the result.
+    if (region.startPos < 0) {
+        result.append(-region.startPos, U2Msa::GAP_CHAR);
+    }
+    // Iterate interleaved gap & core sequence regions.
+    while (gapIndex <= gaps.length() && pos < regionEnd && corePos < coreLength) {
+        const U2MsaGap *gap = gapIndex == gaps.length() ? nullptr : &gaps[gapIndex];
+        bool isValidGapModel = gapIndex == 0 || gap == nullptr || gap->startPos > gaps[gapIndex - 1].endPos();
```
DmitriiSukhomlinov, line 458: Wait, `gapIndex` equals to 0 on the first iteration, `gapIndex - 1` is `-1`, so `gaps[-1]` could cause a lot of problems here. Why isn't that happening?

DmitriiSukhomlinov: Oh, I see, `gapIndex == 0` is here. But don't you want to use `U2MsaGap::isValid()` too?

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
```diff
@@ -432,4 +432,80 @@ void MsaRowUtils::removeTrailingGapsFromModel(qint64 length, QVector<U2MsaGap> &
...
+                           "Invalid sub-region in the core region",
+                           {});
+                result.append(coreSequence + corePos + corePartOffset, nChars);
+            }
+            corePos += corePartLength;
+            pos += corePartLength;
+            SAFE_POINT(corePos <= coreLength, "Core sequence overflow!", {});
+        } else {
+            // Processing gap region.
+            int gapPartLength = (int)gap->length;
```
DmitriiSukhomlinov, line 488: The same, data loss is possible

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
```diff
@@ -432,4 +432,80 @@ void MsaRowUtils::removeTrailingGapsFromModel(qint64 length, QVector<U2MsaGap> &
...
+                } else {  // Add middle part of the gap region.
+                    nGaps = region.length;
+                }
+                SAFE_POINT(nGaps <= gapPartLength && result.length() + nGaps <= region.length, "Invalid gap region", {});
+                result.append(nGaps, U2Msa::GAP_CHAR);
+            }
+            pos += gap->length;
+            gapIndex++;
+        }
+        SAFE_POINT(pos < region.startPos ? result.isEmpty() : result.length() <= pos - region.startPos, "Invalid result region after processing gap.", {});
```
DmitriiSukhomlinov, line 504: Sometimes you try to fill the only line with a number or expressions and it makes them difficult to read. There is no need of doing that, we do not have lines limitations! I think, it's better either to separate this stuff on several `SAFE_POINT`s or use the usual, very base and common, but so much pretty understandable `if-else`

src/libs_3rdparty/QSpec/src/primitives/GTWidget.cpp
```diff
@@ -226,18 +226,16 @@ QList<QLabel *> GTWidget::findLabelByText(GUITestOpStatus &os,
                                           QWidget *parentWidget,
                                           const GTGlobals::FindOptions &options) {
     QList<QLabel *> resultLabelList;
-    for (int time = 0; time < GT_OP_WAIT_MILLIS; time += GT_OP_CHECK_MILLIS) {
-        GTGlobals::sleep(time > 0 ? GT_OP_CHECK_MILLIS && resultLabelList.isEmpty() : 0);
+    for (int time = 0; time < GT_OP_WAIT_MILLIS && resultLabelList.isEmpty(); time += GT_OP_CHECK_MILLIS) {
```
DmitriiSukhomlinov, line 229: The same about lines. It would be so much better, if you make just the separate line like this:
`CHECK_CONTINUE(resultLabelList.isEmpty());`

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Core/src/datatype/msa/MultipleSequenceAlignmentRow.cpp
yalgaer: I do not agree. The checked subject (gappedPosition) should be always on the same side.

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
yalgaer: There is no data loss with 64->32 bit conversion in MSA. We do not support (even open) MSA objects with more than 100M chars in the row. It is a global for UGENE.

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
yalgaer: How can I use U2MsaGap::isValid() here to compare 2 sibling gaps? Can you post an example, please?

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
yalgaer: Same: no data loss. MSA is 32 bit dataset. We will crash if its bigger and have a guard on the moment of creation because of it.

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
yalgaer: You right, but I do it only for SAFE_POINTS. SAFE_POINTS are not checks (which is a part of "normal" logic) so I try to keeping them as short as possible, to give priority to the "real" code.

src/libs_3rdparty/QSpec/src/primitives/GTWidget.cpp
yalgaer: There are about 20 time-out waiters like this in our code. Almost all of them use this pattern.

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
yalgaer: Added more memory opts to the fix. Answered all question. Re-checking the tests now

src/corelibs/U2Core/src/datatype/msa/MultipleSequenceAlignmentRow.cpp
DmitriiSukhomlinov: Let us both do in a way we like

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
DmitriiSukhomlinov: I'm not talking about comparing, just validate the `U2MsaGap` object

DmitriiSukhomlinov (approved):
Just one minor comment, do it if you wish

src/corelibs/U2Core/src/util/MsaRowUtils.cpp
yalgaer: What kind of error can be here with gap? Negative length should be handled in the gap constructor, but not in the 100+ places where gaps are used.

src/corelibs/U2Core/src/datatype/msa/MultipleSequenceAlignmentRow.cpp
yalgaer: Ok, I never have problems like this in reviews: we do not have a strict code style about it.

PR #715, opened 2022-01-17
UGENE-7227 Assert in WD on saving sequence in "UGENE Database" format

rasputinkirill:


yalgaer (changes requested):
Nice to see one more patch from the "master of easy fixes" :)

Unfortunately the fix is wrong.

The simplest argument could be "Why extending GenbankWriter but not FastaWriter"?
Right, both of them should not be used to pre-process data stored in UGENEDB format.

Here is details about the correct fix:

1)  In UgeneDBWriter implement the required function independently from Genbank or any other format.
2) Do not handle any data types (like annotations) you have no tests for.
3) In the test check the saved result. If the workflow didn't run at all and has no errors in log the current test will pass.
4) optional: it would be nice if you support streaming for UGENEDB format (like in FASTA)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -555,6 +555,58 @@ GUI_TEST_CLASS_DEFINITION(test_7212) {
...
+    WorkflowProcessItem *readSeq = GTUtilsWorkflowDesigner::addElement(os, "Read Sequence", true);
+    WorkflowProcessItem *writeSeq = GTUtilsWorkflowDesigner::addElement(os, "Write Sequence", true);
+
+    GTUtilsWorkflowDesigner::connect(os, readSeq, writeSeq);
+
+    GTUtilsWorkflowDesigner::addInputFile(os, "Read Sequence", dataDir + "/samples/Genbank/murine.gb");
+    GTUtilsWorkflowDesigner::click(os, "Write Sequence");
+    GTUtilsWorkflowDesigner::setParameter(os, "Document format", "UGENE Database", GTUtilsWorkflowDesigner::comboValue);
+
+    const GTLogTracer logTracer;
```
yalgaer: Do not use consts randomly. We either use them everywhere consistently or do not use them at all (at least for locals)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -555,6 +555,58 @@ GUI_TEST_CLASS_DEFINITION(test_7212) {
...
+    GTUtilsWorkflowDesigner::runWorkflow(os);
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    CHECK_SET_ERR(!logTracer.hasErrors(), "Errors in log: " + logTracer.getJoinedErrorString());
+
+    const QStringList outputFiles = GTUtilsDashboard::getOutputFiles(os);
+    CHECK_SET_ERR(1 == outputFiles.size(), QString("Too many output files, exptected 1, got %1").arg(outputFiles.size()));
```
yalgaer: Do not use Yoda style for comparisons (here and below)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -555,6 +555,58 @@ GUI_TEST_CLASS_DEFINITION(test_7212) {
...
+
+    CHECK_SET_ERR(2 == documents.first().count(), QString("An incorrect objects count in '%1' document: expected 1, got %2").arg(documents.keys().first()).arg(documents.first().count()));
+
+    const QString expectedObjectName = "[s] NC_001363";
+    const QString actualObjectName = documents.first().first();
+    CHECK_SET_ERR(expectedObjectName == actualObjectName, QString("An unexpected object name: expected '%1', got '%2'").arg(expectedObjectName).arg(actualObjectName));
+    
+    const QString expectedObjectName2 = "[a] NC_001363 features";
+    const QString actualObjectName2 = documents.first()[1];
+    CHECK_SET_ERR(expectedObjectName2 == actualObjectName2, QString("An unexpected object name: expected '%1', got '%2'").arg(expectedObjectName2).arg(actualObjectName2));
```
yalgaer: Check not only the object name by the sequence parameters too: length and some content. You can hardcode some part of the content here to make the test shorter.

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,143 @@ bool MSAWriter::isStreamingSupport() const {
...
+    if (data.contains(BaseSlots::DNA_SEQUENCE_SLOT().getId())) {
+        SharedDbiDataHandler seqId = data[BaseSlots::DNA_SEQUENCE_SLOT().getId()].value<SharedDbiDataHandler>();
+        seqObj.reset(StorageUtils::getSequenceObject(context->getDataStorage(), seqId));
+        SAFE_POINT(nullptr != seqObj.data(), tr("UgeneDB writer: NULL sequence object"), );
+
+        U2OpStatusImpl os;
+        DNASequence seq = seqObj->getWholeSequence(os);
+        SAFE_POINT_OP(os, );
+        QMapIterator<QString, QVariant> it(seq.info);
+        while (it.hasNext()) {
```
yalgaer: Copy-paste from Genbank is good, but does it have any meaning for UGENEDB format here?

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,143 @@ bool MSAWriter::isStreamingSupport() const {
...
+            annotationName = getAnnotationName(seq.getName());
+        }
+
+        dna = qobject_cast<U2SequenceObject *>(doc->findGObjectByName(seq.getName()));
+        if (!dna && !seq.isNull()) {
+            dna = addSeqObject(doc, seq);
+        }
+    }
+
+    if (data.contains(BaseSlots::ANNOTATION_TABLE_SLOT().getId())) {
```
yalgaer, line 912: Do not add support for annotations with no dedicated XML tests

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,143 @@ bool MSAWriter::isStreamingSupport() const {
...
+    QList<GObject *> anObjList;
+    if (data.contains(BaseSlots::ANNOTATION_TABLE_SLOT().getId())) {
+        const QVariant &annsVar = data[BaseSlots::ANNOTATION_TABLE_SLOT().getId()];
+        const QList<SharedAnnotationData> atl = StorageUtils::getAnnotationTable(context->getDataStorage(), annsVar);
+
+        if (!atl.isEmpty()) {
+            if (annotationName.isEmpty()) {
+                annotationName = QString("unknown features %1").arg(entryNum);
+            }
+            AnnotationTableObject *att = new AnnotationTableObject(annotationName, context->getDataStorage()->getDbiRef());
```
yalgaer: do not use type twice in the line

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,143 @@ bool MSAWriter::isStreamingSupport() const {
...
+        }
+        if (!anObjList.isEmpty()) {
+            objectsMap[GObjectTypes::ANNOTATION_TABLE] = anObjList;
+        }
+    }
+    CHECK(!objectsMap.isEmpty(), );
+
+    format->storeEntry(io, objectsMap, os);
+
+    foreach (GObject *o, anObjList) {
```
yalgaer: On line 1003 you export this list content and now you are deleting it. Looks like a bug?

src/plugins/workflow_designer/src/library/DocWorkers.h
```diff
@@ -212,6 +212,24 @@ class GFFWriter : public BaseDocWriter {
...
+    }
+
+protected:
+    void data2doc(Document *, const QVariantMap &) override;
+    void storeEntry(IOAdapter *io, const QVariantMap &data, int entryNum) override;
+    bool hasDataToWrite(const QVariantMap &data) const override;
+    QSet<GObject *> getObjectsToWrite(const QVariantMap &data) const override;
+
+public:
+    static void data2document(Document *, const QVariantMap &, WorkflowContext *);
```
yalgaer: Add docs to all new public non-override methods

yalgaer (changes requested):

smlnkcv (commented):
There as no Assert after fix.

src/plugins/workflow_designer/src/library/DocWorkers.cpp
rasputinkirill: It is just memory cleaning after creation objects at line 982. I think it should be replaced with qDeleteAll().

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -710,6 +710,8 @@ void SeqWriter::data2doc(Document *doc, const QVariantMap &data) {
         RawSeqWriter::data2document(doc, data, context);
     } else if (BaseDocumentFormats::GFF == fid) {
         GFFWriter::data2document(doc, data, context);
+    }else if (BaseDocumentFormats::UGENEDB == fid) {
```
yalgaer: Do not use yoda style in new code (move the constant/expected value to the right)

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+
+void UgeneDBWriter::data2document(Document *doc, const QVariantMap &data, WorkflowContext *context) {
+    QScopedPointer<U2SequenceObject> seqObj(nullptr);
+    U2SequenceObject *dna = nullptr;
+    QString annotationName;
+
+    if (data.contains(BaseSlots::DNA_SEQUENCE_SLOT().getId())) {
+        SharedDbiDataHandler seqId = data[BaseSlots::DNA_SEQUENCE_SLOT().getId()].value<SharedDbiDataHandler>();
+        seqObj.reset(StorageUtils::getSequenceObject(context->getDataStorage(), seqId));
+        SAFE_POINT(nullptr != seqObj.data(), tr("UgeneDB writer: NULL sequence object"), );
```
yalgaer: 1) do not use yoda style.
2) do not use translation in SAFE_POINTS. If this is a user-facing log use CHECK

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+
+        if (seq.getName().isEmpty()) {
+            int num = doc->findGObjectByType(GObjectTypes::SEQUENCE).size();
+            seq.setName(QString("unknown sequence %1").arg(num));
+        } else {
+            annotationName = getAnnotationName(seq.getName());
+        }
+
+        dna = qobject_cast<U2SequenceObject *>(doc->findGObjectByName(seq.getName()));
+        if (!dna && !seq.isNull()) {
```
yalgaer: Why don't you add an input sequence to the document if the document already has another sequence?

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+            dna = addSeqObject(doc, seq);
+        }
+    }
+
+    if (data.contains(BaseSlots::ANNOTATION_TABLE_SLOT().getId())) {
+        const QVariant &annsVar = data[BaseSlots::ANNOTATION_TABLE_SLOT().getId()];
+        const QList<SharedAnnotationData> atl = StorageUtils::getAnnotationTable(context->getDataStorage(), annsVar);
```
yalgaer: Do not use "random" consts (or consts for locals in methods at all). 
Example below you write "QList<GObject *> relAnns" with no const at all. Having consts in one places and not others is inconsistent code style.

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+                                                                                          doc->getObjects(),
+                                                                                          UOF_LoadedOnly);
+                att = relAnns.isEmpty() ? nullptr : qobject_cast<AnnotationTableObject *>(relAnns.first());
+            }
+            if (!att) {
+                if (annotationName.isEmpty()) {
+                    int featuresNum = doc->findGObjectByType(GObjectTypes::ANNOTATION_TABLE).size();
+                    annotationName = QString("unknown features %1").arg(featuresNum);
+                }
+                att = qobject_cast<AnnotationTableObject *>(doc->findGObjectByName(annotationName));
```
yalgaer: Same question here: why do you ignore the value in the slot if the document already has an object with the same name?

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+                if (att == nullptr) {
+                    doc->addObject(att = new AnnotationTableObject(annotationName, context->getDataStorage()->getDbiRef()));
```
yalgaer: nit: do not use inlined initialization except some very common patterns like "if (auto x= cast<>)"

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+            dna = addSeqObject(doc, seq);
+        }
+    }
+
+    if (data.contains(BaseSlots::ANNOTATION_TABLE_SLOT().getId())) {
+        const QVariant &annsVar = data[BaseSlots::ANNOTATION_TABLE_SLOT().getId()];
+        const QList<SharedAnnotationData> atl = StorageUtils::getAnnotationTable(context->getDataStorage(), annsVar);
+
+        if (!atl.isEmpty()) {
+            AnnotationTableObject *att = nullptr;
```
yalgaer: Give good names to variables. What the last "t" in "att" means?

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+                }
+                algoLog.trace(QString("Adding features [%1] to UgeneDB doc %2").arg(annotationName).arg(doc->getURLString()));
+            }
+            att->addAnnotations(atl);
+        }
+    }
+}
+
+bool UgeneDBWriter::hasDataToWrite(const QVariantMap &data) const {
+    return SeqWriter::hasSequenceOrAnns(data);
```
yalgaer, line 944: This is not true since you skip empty sequences: if only empty sequence in an input you have no data to write.
Please fix this method or accept empty sequences (is better) as writable.

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+}
+
+QSet<GObject *> UgeneDBWriter::getObjectsToWrite(const QVariantMap &data) const {
+    return QSet<GObject *>() << SeqWriter::getSeqObject(data, context) << SeqWriter::getAnnObject(data, context);
```
yalgaer: nit: use modern syntax: {{}, {}}

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+
+void UgeneDBWriter::streamingStoreEntry(DocumentFormat *format, IOAdapter *io, const QVariantMap &data, WorkflowContext *context, int entryNum) {
+    U2OpStatus2Log os;
+    QScopedPointer<U2SequenceObject> seqObj(nullptr);
+    QString annotationName;
+    if (data.contains(BaseSlots::DNA_SEQUENCE_SLOT().getId())) {
+        seqObj.reset(getCopiedSequenceObject(data, context, os));
+        SAFE_POINT_OP(os, );
+
+        if (seqObj->getGObjectName().isEmpty()) {
```
yalgaer, line 961: Extract this code duplication into a method

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+            auto *att = new AnnotationTableObject(annotationName, context->getDataStorage()->getDbiRef());
+            anObjList << att;
+            att->addAnnotations(atl);
+        }
+    }
+
+    QMap<GObjectType, QList<GObject *>> objectsMap;
+    {
+        if (nullptr != seqObj.data()) {
+            QList<GObject *> seqs;
```
yalgaer: Do not use yoda style + use modern syntax with list initializers ({})

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,134 @@ bool MSAWriter::isStreamingSupport() const {
...
+    QList<GObject *> anObjList;
+    if (data.contains(BaseSlots::ANNOTATION_TABLE_SLOT().getId())) {
+        const QVariant &annsVar = data[BaseSlots::ANNOTATION_TABLE_SLOT().getId()];
+        const QList<SharedAnnotationData> atl = StorageUtils::getAnnotationTable(context->getDataStorage(), annsVar);
+
+        if (!atl.isEmpty()) {
+            if (annotationName.isEmpty()) {
+                annotationName = QString("unknown features %1").arg(entryNum);
+            }
+            auto *att = new AnnotationTableObject(annotationName, context->getDataStorage()->getDbiRef());
```
yalgaer: Here you have a different behavior from the previous method: you use all annotation tables. This looks inconsistent.

yalgaer (changes requested):
Please check the comments. 
The biggest problem is that streaming & non-streaming methods process input differently

src/plugins/workflow_designer/src/library/DocWorkers.cpp
rasputinkirill: Agreed, addSeqObject can roll name.

yalgaer: Please update the patch, so I can continue reviewing

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,124 @@ bool MSAWriter::isStreamingSupport() const {
...
+
+        if (seqObj->getGObjectName().isEmpty()) {
+            seqObj->setGObjectName(QString("unknown sequence %1").arg(entryNum));
+            annotationName = QString("unknown features %1").arg(entryNum);
+        } else {
+            annotationName = getAnnotationName(seqObj->getGObjectName());
+        }
+    }
+
+    QList<GObject *> anObjList;
```
yalgaer: You do not need a list here? There is always 1 or 0 annotation objects?

src/plugins/workflow_designer/src/library/DocWorkers.cpp
```diff
@@ -873,6 +875,124 @@ bool MSAWriter::isStreamingSupport() const {
...
+    if (data.contains(BaseSlots::ANNOTATION_TABLE_SLOT().getId())) {
+        const QVariant &annsVar = data[BaseSlots::ANNOTATION_TABLE_SLOT().getId()];
+        QList<SharedAnnotationData> atl = StorageUtils::getAnnotationTable(context->getDataStorage(), annsVar);
+
+        if (!atl.isEmpty()) {
```
yalgaer: Rename att & atl to something more readable

yalgaer (changes requested):
Please fix minor issues left, run the tests and I will approve

yalgaer (approved):

PR #723, opened 2022-01-21
Make QTimer::singleShot calls safe

yalgaer:
QTimer::singleShot may call the functor object after the 'stack' object is destroyed if no correct 'context' is provided.

I made this error once by myself and decided to check and fix all places like this in UGENE


DmitriiSukhomlinov (changes requested):
For now, our main bug-tracking system is JIRA. Every meaningful comment should be connected to the JIRA issue, overwise it's flying in the space and there's no idea where it grows from. Create issues in JIRA and connect PRs to them - you do not need make any descriptions here at GitHub, it's better to keep them in issues.

yalgaer:
Trivial code style/quality improvements can be sent as PR with no related issues. 
I do 2-3 of them per day usually - creating issues for such simple changes like "Remove sleep() from file X" is a level of bureaucracy that will only slow down our development.

Any new feature / bug fix should have an issue assigned.

PR #726, opened 2022-01-23
UGENE-7532: UGENE can set font point size to 0 in viewers and make th…

yalgaer:
…em inaccessible

### DIFF
--- a/src/corelibs/U2View/src/ov_msa/MaEditor.cpp
+++ b/src/corelibs/U2View/src/ov_msa/MaEditor.cpp
@@ -463,7 +463,7 @@ void MaEditor::updateFontMetrics() {
     const int minimumSafeFontPointSize = 8;  // This value was historically used in UGENE as minimum with no known issues.
     QFont fontToEstimate = font;
     int estimatedMinimumFontPointSize = minimumSafeFontPointSize;  // Start with a safe value and estimate smaller values.
-    while (fontToEstimate.pointSize() > 0) {
+    while (fontToEstimate.pointSize() > 1) {
         int charWidth = getUnifiedSequenceFontCharRect(fontToEstimate).width();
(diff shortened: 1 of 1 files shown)


src/corelibs/U2View/src/ov_msa/MaEditor.cpp
```diff
@@ -463,7 +463,7 @@ void MaEditor::updateFontMetrics() {
     const int minimumSafeFontPointSize = 8;  // This value was historically used in UGENE as minimum with no known issues.
     QFont fontToEstimate = font;
     int estimatedMinimumFontPointSize = minimumSafeFontPointSize;  // Start with a safe value and estimate smaller values.
-    while (fontToEstimate.pointSize() > 0) {
+    while (fontToEstimate.pointSize() > 1) {
```
DmitriiSukhomlinov, line 466: The error sign is `QFont::setPointSize: Point size <= 0 (0), must be greater than 0`.
The point size here is already greater than 0, you made them greater than 1. I don't see how it helps.

DmitriiSukhomlinov (changes requested):
Some strange fix, without any proves that it really helps, disapproved.

DmitriiSukhomlinov:
Also, don't forget, that the original problem has appeared on Windows, so it should be reproducible on this system.

src/corelibs/U2View/src/ov_msa/MaEditor.cpp
yalgaer: Could you please check the algorithm beyond the changed line? The only reason to check the value for 1 here is to decrease it by -1 3 lines later.

yalgaer:
> Also, don't forget, that the original problem has appeared on Windows, so it should be reproducible on this system.

It is reproducible on any OS, because it depends on font metrics only. Have you tried the font I mentioned?

PR #730, opened 2022-01-24
UGENE-7534. Part 1: support multiple queries in BLAST task API

yalgaer:
This is the first part to fix the problem documented in JIRA issue.
It adds multi-queries support to BLAST (we still are not using it anywhere) and improves some tests and cleans-up the code (removes extra states or memory management where it is possible).
This should be no-op for the current UGENE features state.

In the next PR I will add a real use & tests for this multi-query option.


src/corelibs/U2Core/src/util/U2SequenceUtils.h
```diff
@@ -168,15 +168,28 @@ class U2CORE_EXPORT U2SequenceDbiHints {
...
 };
 
-// untwist/expand circular sequence
-class U2CORE_EXPORT U2PseudoCircularization : public QObject {
+/**
+ * Utility class to deal with circular sequences.
+ * Supports mapping of the original sequence coordinates to a circular sequence coordinates with a specified 'overlap' range.
+ */
+class U2CORE_EXPORT U2PseudoCircularization {
+    U2PseudoCircularization() = default;
```
DmitriiSukhomlinov: If it's a utility class, the constructor should be `delete`'d

src/plugins/external_tool_support/src/blast/BlastCommonTask.cpp
```diff
@@ -238,6 +249,9 @@ void BlastCommonTask::parseTabularLine(const QByteArray &line) {
         stateInfo.setError(tr("Incorrect number of fields in line: %1").arg(elements.size()));
         return;
     }
+    int querySequenceIndex = parseQuerySequenceIndex(elements[0]);
```
DmitriiSukhomlinov, line 252: Could be crash is [0] isn't exists, add `SAFE_POINT` or use `->value(0);` if the default value is possible

src/plugins/external_tool_support/src/blast/BlastCommonTask.cpp
```diff
@@ -348,36 +366,53 @@ void BlastCommonTask::parseXMLResult() {
...
+            parseXMLHit(hits.at(i), querySequenceIndex);
+        }
     }
 }
 
-void BlastCommonTask::parseXMLHit(const QDomNode &xml) {
-    QString id, def, accession;
+int BlastCommonTask::parseQuerySequenceIndex(const QString &querySequenceName) const {
+    SAFE_POINT_EXT(querySequenceName.startsWith(QUERY_SEQUENCE_NAME_PREFIX), uiLog.trace("Unexpected query sequence name: " + querySequenceName), -1);
+    bool isOk;
```
DmitriiSukhomlinov, line 384: You have no checks of this variable

src/plugins/external_tool_support/src/blast/BlastCommonTask.h
```diff
@@ -62,16 +63,26 @@ class BlastCommonTask : public ExternalToolSupportTask {
...
     void parseTabularLine(const QByteArray &line);
 
     void parseXMLResult();
-    void parseXMLHit(const QDomNode &xml);
-    void parseXMLHsp(const QDomNode &xml, const QString &id, const QString &def, const QString &accession);
+    void parseXMLHit(const QDomNode &xml, int querySequenceIndex);
+    void parseXMLHsp(const QDomNode &xml, int querySequenceIndex, const QString &id, const QString &def, const QString &accession);
     QString getAcceptableTempDir() const;
+
+    /**
```
DmitriiSukhomlinov, line 81: 5 spaces indent

DmitriiSukhomlinov (dismissed):
Minor comments, approved

src/corelibs/U2Core/src/util/U2SequenceUtils.h
yalgaer: I do the same here: make it a private.

src/plugins/external_tool_support/src/blast/BlastCommonTask.cpp
yalgaer: Correct. Fixed.

src/plugins/external_tool_support/src/blast/BlastCommonTask.cpp
yalgaer: Could you check 3 lines above this line: if (elements.size() != 12) {....

I would discourage anyone to use value() in vectors/list because this is the way to hide real bugs: crash is better than incorrectly computed result.

src/plugins/external_tool_support/src/blast/BlastCommonTask.h
yalgaer: All is good with indentation here. Sorry, do not get it.

PR #734, opened 2022-01-25
UGENE-7531 Crash in case of the exporting of PCR result with an empty annotation table

DmitriiSukhomlinov:
The problem appears because `U2FeatureUtils::getSortedSubgroups` expects, that annotation tables always have annotations, but annotation tables could be empty.


yalgaer (commented):
The fix looks correct. Will approve after GUI tests are finished

smlnkcv (commented):
No crash

yalgaer (dismissed):

yalgaer (approved):

PR #738, opened 2022-01-26
UGENE-7491 Workflow properties not saved for some NGS workflow examples

rasputinkirill:


src/plugins/workflow_designer/src/WorkflowViewController.cpp
```diff
@@ -1975,8 +1994,14 @@ void WorkflowView::sl_saveScene() {
     }
     propertyEditor->commit();
     Task *t = new SaveWorkflowSceneTask(getSchema(), getMeta());
-    AppContext::getTaskScheduler()->registerTopLevelTask(t);
-    connect(t, SIGNAL(si_stateChanged()), SLOT(sl_onSceneSaved()));
+    if (saveImmideately) {
```
yalgaer: Why can't you call the task the same way as before?

If the schema is asynchronously deleted it can happen with a regular scenario too: I click "save" and before the task is finished (it may be blocked in the task queue) I close the editor.

I see no problem to write the schema file from the main thread: we know that it is small. If you need this: extract the code in task::run into a static method utility and do not use the task in either case.

src/plugins/workflow_designer/src/WorkflowViewController.cpp
```diff
@@ -2255,7 +2280,7 @@ bool WorkflowView::onCloseEvent() {
     return true;
 }
 
-bool WorkflowView::confirmModified() {
+bool WorkflowView::confirmModified(bool saveImmideately) {
```
yalgaer: Unused parameter?

yalgaer (changes requested):

yalgaer (changes requested):
The patch looks good now, but why don't you remove the whole task at all?
We do IO from the main thread when it is predictable: we read/write settings, create directories, etc..

Writing a small workflow file (several kb) won't take a long. A rendering itself may be more blocking than that.

I would suggest to remove this task completely and use the method you already use to store the file.
In this case you can also remove the boolean parameter.

src/plugins/workflow_designer/src/WorkflowViewController.cpp
```diff
@@ -1974,9 +1995,10 @@ void WorkflowView::sl_saveScene() {
         sl_updateTitle();
     }
     propertyEditor->commit();
-    Task *t = new SaveWorkflowSceneTask(getSchema(), getMeta());
-    AppContext::getTaskScheduler()->registerTopLevelTask(t);
-    connect(t, SIGNAL(si_stateChanged()), SLOT(sl_onSceneSaved()));
+    HRSchemaSerializer::updateWorkflowSchemaPathSettings(meta);
+    U2OpStatus2Log os;
+    HRSchemaSerializer::saveSchema(schema.get(), &meta, meta.url, os);
```
yalgaer: Please check the state of the 'os' separately from scene->setModified().
The scene->isModified should not change (to either false or true) if the saveSchema is failed.

yalgaer (changes requested):

yalgaer (approved):

PR #739, opened 2022-01-26
UGENE-7419 SAFE_POINT after deleting exported document in ugebegb format from the disk

DmitriiSukhomlinov:
In general - this error happens when UGENE works with ".ugenedb" files. UGENE thinks, that it also works with the session db, but when it's just a usual file, UGENE couldn't process some steps correctly.


smlnkcv (dismissed):

src/corelibs/U2Core/src/models/GObject.cpp
```diff
@@ -321,7 +322,7 @@ void GObject::removeRelations(const QString &removedDocUrl) {
             changed = true;
         }
     }
-    if (changed) {
+    if (changed && QFileInfo::exists(removedDocUrl)) {
```
yalgaer: You know this is a hack and you know I do not approve hacks.
Removal of relations from object relations map does not depend on existence or non-existence of the related file at all.

src/corelibs/U2Gui/src/util/project/ProjectUpdater.cpp
```diff
@@ -135,17 +135,19 @@ void ProjectUpdater::fetchObjectsInUse(const U2DbiRef &dbiRef, U2OpStatus &os) {
...
+            U2OpStatus2Log os;
             Document *doc = object->getDocument();
             if (nullptr != doc && doc->isStateLocked()) {
                 continue;
             }
             const U2EntityRef ref = object->getEntityRef();
             if (!dbiRef2Connections.contains(ref.dbiRef)) {
-                dbiRef2Connections.insert(ref.dbiRef, new DbiConnection(ref.dbiRef, os));
+                auto dbiCon = new DbiConnection(ref.dbiRef, os);
                 CHECK_CONTINUE(!os.hasError());
```
yalgaer, line 148: C++ is not a memory managed language. Please think about memory deallocation every time you create objects with "new"

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1292,6 +1292,24 @@ GUI_TEST_CLASS_DEFINITION(test_7415_3) {
...
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    // Now remove it from the disk manually
+    QFile::remove(sandBoxDir + "test_7419.ugenedb");
+
+    // Expected: the message box about lost database -> click OK -> view is closed.
+    GTUtilsDialog::waitForDialog(os, new MessageBoxDialogFiller(os, QMessageBox::Ok, "The document 'test_7419.ugenedb' was removed from its original folder. Therefore, it will be deleted from the current project"));
+
+    // The document update happens each 3 seconds, the messagebox will appear on the closest update
+    GTGlobals::sleep(3000);
```
yalgaer: Sleeps are bad for many reasons: they do not guarantee anything and make tests slow.
Wait here until all pending dialog fillers are completed: use "checkNoActiveWaiters"

yalgaer (changes requested):

src/corelibs/U2Gui/src/util/project/ProjectUpdater.cpp
DmitriiSukhomlinov: I could make this pointer shared, but the point is, that if `os` has an error, that means, that the memory hasn't been allocated and there's nothing to free. Do you want me to use `QSharedPointer`?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
DmitriiSukhomlinov: OK. By the way, if you don't want us to use `GTGlobasl::sleep`, make them private

src/corelibs/U2Core/src/models/GObject.cpp
DmitriiSukhomlinov: Yes, my bad, forget it. 

The problem here is that `setObjectRelations(rels);` needs to create a connection to the existed database, which is the current `GObject` has reference to (the path to the database file is stored in `GObject::entityRef::dbiRef::dbiId`) . As far as I can understand, it anticipates, that this database would be the session one, wich is stashed far-far on the disk and shouldn't be deteled by user. But, we have a separate database file here, which could be deleted by user and `GObject` doesn't expect it. I can check that the database file exists, but not sure that it's the solution you want to see here. 

So, what the best solution you need?

DmitriiSukhomlinov: I replaced "alignment.ugenedb" to "murine.ugenedb", cuz this problem doesn't appear on the first file (because "alignment.ugenedb" doesn't have any relations, but "murine.ugenedb" has related sequence and annotated table objects)

yalgaer: I do not know a good solution here right now. Let we 1 day to dig into the problem and tomorrow I will suggest a solution

src/corelibs/U2Gui/src/util/project/ProjectUpdater.cpp
yalgaer: CHECK is an expected behavior and should be processed correctly: you can manually delete the pointer or use a shared pointer: once there are no memory leaks both solutions are good.

As an alternative you can use SAFE_POINT: in this case we can allow memory leaks because we do not expect this behavior at all and can have non critical memory leaks like this.

Up to you, I'm OK with both options.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: This is the goal. Every week I remove dozen of sleeps but there are still a lot of them

The only case when we really need them is when we want 1) to manage some OS-level behavior. For example a sleep() between clicks to avoid double click or 2) in utility classes: like wait up to N seconds until task is finished and fail otherwise. All such utility classes are highly optimized to avoid unnecessary sleeps: checks for results every 100ms

src/corelibs/U2Core/src/models/GObject.cpp
DmitriiSukhomlinov: OK. You could just build this branch, run the test and the problem SAFE_POINT will appear

src/corelibs/U2Gui/src/util/project/ProjectUpdater.cpp
DmitriiSukhomlinov: OK, shared pointers are used

yalgaer: May be there is a better solution here (optional, up to you) : the problem is that we use a constructor to make some logic and want to check the result value. 

A classic factory method will do a better job in this case: use a static factory method instead of the constructor to instantiate the object.

The factory method should return 'nullptr' if instantiation is failed, so no memory management is needed on the caller side, it is all done in the callee (factory method)

src/corelibs/U2Core/src/models/GObject.cpp
yalgaer: Here is my 'minimal-fix' suggestion:
```
@@ -159,7 +159,7 @@ void GObject::setObjectRelations(const QList<GObjectRelation> &list) {
 void GObject::setRelationsInDb(QList<GObjectRelation> &list) const {
     U2OpStatus2Log os;
     DbiConnection con(entityRef.dbiRef, os);
-    SAFE_POINT_OP(os, );
+    CHECK_OP(os, );  // Database is not available for some reason. It is may be deleted.
```

A complete fix (out of scope in this PR) must check all DB operations for successful completion like IO ops.

src/corelibs/U2Gui/src/util/project/ProjectUpdater.cpp
DmitriiSukhomlinov: It's a solution, of course, but I prefer to use shared pointers where it's possible cuz this looks like a more advanced way of resolving the memory allocation problems - we do not need to think about it at all, the smart pointer do it for us. So, if it's not necessary, I'd rather leave `QSharedPointer` here

src/corelibs/U2Core/src/models/GObject.cpp
DmitriiSukhomlinov: OK

yalgaer (approved):

PR #748, opened 2022-02-03
UGENE-4671. Exporting Circular View as image with non-default size is

evelinabis:
incorrect.


src/plugins/circular_view/src/CircularView.cpp
```diff
@@ -408,14 +408,9 @@ void CircularViewRenderArea::paintContent(QPainter &p, bool paintSelection, bool
...
 void CircularViewRenderArea::paintContent(QPainter &p, int w, int h, bool paintSelection, bool paintMarker) {
-    qreal scaleCoeff;
-    if (w <= h) {
-        scaleCoeff = (qreal)w / width();
-        p.translate(0, (h - scaleCoeff * height()) / 2);
-    } else {
-        scaleCoeff = (qreal)h / height();
-        p.translate((w - scaleCoeff * width()) / 2, 0);
-    }
+    double dw = w, dh = h;
```
yalgaer: You do not need to redefine these variables, just cast them in the std::min function
The names look misleading: like dw = deltaWidth.
If you want to redefine them: 1) Give a non-measling name: widthAsDouble, heightAsDouble and do not use single line initializers (this is unsafe in general, like ```int* a, b``` -> a & b have different types

yalgaer (dismissed):
Looks good, will approve after the minor style issues are fixed.

yalgaer (changes requested):
The patch looks good, but XML tests are failing. Could you please merge with master first?

yalgaer (approved):

PR #758, opened 2022-02-09
Multiline view of the MSA editor

ichebyki:
Multiline view of the MSA editor


yalgaer (changes requested):
Please mark this review as "Ready for review" when GUI tests are ready (old do not fail and there is a new suite for the new mode)

This will help to find/fix most of the issues before the real review.

yalgaer:
UI issues found during the manual testing
1) COI aln has no vertical scrollbar.
2) Horizontal scrollbar should not be shown at all.
3) Create only widgets shown on the screen, but not all pages for the whole MSA.

ichebyki:
New PR was created 
https://github.com/ugeneunipro/ugene/pull/805

PR #774, opened 2022-02-19
BWA-MEM: Fix "skip mate rescue" parameter type in the workflow designer

starling13:
Fix usage of the numeric value editor for the boolean BWA MEM parameter "skip mate rescue" in the workflow schemes designer.
Before:
skip_mate_rescue_old
After:
SKIP_MATE_RESCUE


yalgaer:
Thank you, merged

PR #783, opened 2022-02-20
Missing linkage to libX11 on Linux for QSpec

### LINKED ISSUE #782: Missing linkage to libX11 on Linux for QSpec
The linux driver use functions provided by libX11:
https://github.com/ugeneunipro/ugene/blob/dd8f3f4d3e05cb220e0414957a01a3a1007f33bd/src/libs_3rdparty/QSpec/src/drivers/GTKeyboardDriverLinux.cpp#L39

When linking with "-Wl,--no-undefined", this causes built errors.

"-lX11"/"X11" should be added to the link libraries in both QSpec.pri and CMakeLists.txt.

yalgaer:
Fixes #782

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ if (CMAKE_COMPILER_IS_GNUCXX)
             " -Werror=unused-variable"
             )
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GCC_WARNING_FLAGS}")
+    LIST(APPEND COMPILER_FLAGS -Wl,-no-undefined )
 endif ()
 
 # 3rd party libraries.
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ if (CMAKE_COMPILER_IS_GNUCXX)
             " -Werror=unused-variable"
             )
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GCC_WARNING_FLAGS}")
+    LIST(APPEND COMPILER_FLAGS -Wl,-no-undefined )
```
StefanBruens:
```suggestion
    LIST(APPEND COMPILER_FLAGS -Wl,--no-undefined )
```

PR #784, opened 2022-02-21
UGENE-7191 SAFE_POINT on exporting annotation

rasputinkirill:


src/corelibs/U2Core/src/models/DocumentModel.cpp
```diff
@@ -313,6 +313,8 @@ void Document::_addObject(GObject* obj) {
 bool Document::removeObject(GObject* obj, DocumentObjectRemovalMode removalMode) {
     SAFE_POINT(df->isObjectOpSupported(this, DocumentFormat::DocObjectOp_Remove, obj->getGObjectType()), "Unsupported format operation", false);
 
+    emit si_removeRelations(obj);
```
yalgaer: This is bad design: when object is removed the Document should emit si_objectRemoved.
If this phase has some stages: before/after: rename the signal to si_beforeObjectRemoved

src/ugeneui/src/project_support/ProjectImpl.cpp
```diff
@@ -71,6 +71,23 @@ void ProjectImpl::makeClean() {
     assert(!isTreeItemModified());
 }
 
+void ProjectImpl::sl_removeAllrelationsWithObject(GObject* obj) {
+    QList<GObject*> allObjs;
+    for (Document *d : getDocuments()) {
```
yalgaer: Use qAsConst()

yalgaer: Fix method name to use correct case for words

src/ugeneui/src/project_support/ProjectImpl.cpp
```diff
@@ -119,6 +136,7 @@ void ProjectImpl::addDocument(Document* d) {
 
     connect(d, SIGNAL(si_objectAdded(GObject*)), SLOT(sl_onObjectAdded(GObject*)));
     connect(d, SIGNAL(si_objectRemoved(GObject*)), SLOT(sl_onObjectRemoved(GObject*)));
+    connect(d, SIGNAL(si_removeRelations(GObject*)), SLOT(sl_removeAllrelationsWithObject(GObject*)));
```
yalgaer: Do not use SIGNAL/SLOT, use function pointers in new code.
Function pointers are validated by the compiler. SIGNAL/SLOT only in runtime

yalgaer (changes requested):
Check comments. Code style/naming only

src/ugeneui/src/project_support/ProjectImpl.cpp
rasputinkirill: It is already constant

yalgaer: wrap getDocuments() with qAsConst() so there is a guarantee that the collection is not changed during the loop.

src/corelibs/U2Core/src/models/DocumentModel.cpp
```diff
@@ -313,11 +313,13 @@ void Document::_addObject(GObject* obj) {
...
     SAFE_POINT(df->isObjectOpSupported(this, DocumentFormat::DocObjectOp_Remove, obj->getGObjectType()), "Unsupported format operation", false);
 
+    emit si_beforeObjectRemoved(obj);
+
     switch (removalMode) {
         case DocumentObjectRemovalMode_Deallocate:
             return _removeObject(obj, true);
         case DocumentObjectRemovalMode_OnlyNotify:
-            emit si_objectRemoved(obj);
+            emit si_afterObjectRemoved(obj);
```
yalgaer: I'm OK with this change, but I only asked to name the new method correctly.
Renaming the old method we need to be very careful to avoid loosing existing SIGNAL/SLOTs.

src/corelibs/U2Core/src/selection/GObjectSelection.cpp
```diff
@@ -113,7 +113,7 @@ void GObjectSelection::removeFromSelection(const QList<GObject*>& obj) {
 void GDocumentObjectSelection::_append(GObject* o) {
     Document* d = o->getDocument();
     assert(d!=NULL);
-    connect(d, SIGNAL(si_objectRemoved(GObject*)), SLOT(void sl_onObjectRemoved(GObject*)));
+    connect(d, SIGNAL(si_afterObjectRemoved(GObject*)), SLOT(void sl_onObjectRemoved(GObject*)));
```
yalgaer: You change it -> you fix it. Use function pointers for all new/updated signal/slot connections :)

yalgaer (commented):
Please fix the qAsConst issue (other are optional) and I will approve after GUI tests are finished.

yalgaer:
Closing as inactive. Please reopen when is ready for the review (all GUI tests fixed)

yalgaer (dismissed):

src/ugeneui/src/project_support/ProjectImpl.cpp
```diff
@@ -71,6 +71,23 @@ void ProjectImpl::makeClean() {
...
 
+void ProjectImpl::sl_removeAllrelationsWithObject(GObject* obj) {
+    QList<GObject*> allObjs;
+    for (Document *d : qAsConst(getDocuments())) {
+        allObjs << d->getObjects();
+    }
+    for (GObject* object : qAsConst(allObjs)) {
+        if (object != obj) {
+            const QList<GObjectRelation> relationList = object->getObjectRelations();
+            for (const GObjectRelation& rel : relationList) {
```
yalgaer: Approved with no qAsConst this time (will fix it by myself). Will be a blocker next time.

yalgaer (approved):

PR #797, opened 2022-03-11
UGENE-5640 Strict consensus algorithm draws an empty space instead of gap if a threshold is too low

DmitriiSukhomlinov:


src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmStrict.cpp
```diff
@@ -58,8 +58,8 @@ char MSAConsensusAlgorithmStrict::getConsensusChar(const MultipleAlignment& ma,
     // use gap is top char frequency is lower than threshold
     int nSeq = (seqIdx.isEmpty() ? ma->getRowCount() : seqIdx.size());
     int currentThreshold = getThreshold();
-    int cntToUseGap = int(currentThreshold / 100.0 * nSeq);
-    int topFreq = freqsByChar[topChar];
+    double cntToUseGap = double(currentThreshold) / 100.0 * nSeq;
```
yalgaer: Cast to double is not needed.

yalgaer (changes requested):
Please fix the xml test

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmStrict.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
I checked an XML test. The result has been changed a bit. This has happened because since I removed the `int` rounding numbers became more accurate. For example, it was a comprising before the fix:
```
int cntToUseGap = int(currentThreshold / 100.0 * nSeq); // 12
int topFreq = freqsByChar[topChar]; // 12
char res = topFreq < cntToUseGap ? U2Msa::GAP_CHAR : (char)topChar; // 12 < 12 => false => (char)topChar
```

But, after the fix it was turned into a:
```
double cntToUseGap = currentThreshold / 100.0 * nSeq; //12.6
double topFreq = freqsByChar[topChar]; // 12.0
char res = topFreq < cntToUseGap ? U2Msa::GAP_CHAR : (char)topChar; // 12.0 < 12.6 => true => U2Msa::GAP_CHAR
```

yalgaer (approved):

PR #801, opened 2022-03-16

rasputinkirill:


yalgaer (commented):
Looks good please fix all builds and run GUI tests.

src/corelibs/U2Core/src/tasks/ExternalToolRunTask.cpp
```diff
@@ -120,7 +120,12 @@ void ExternalToolRunTask::run() {
     while (!externalToolProcess->waitForFinished(1000)) {
         if (isCanceled()) {
             killProcess(externalToolProcess);
-            algoLog.details(tr("Tool %1 is cancelled").arg(toolName));
+            externalToolProcess->waitForFinished(10000);
+            if (externalToolProcess->state() == QProcess::Running) {
+                algoLog.info(tr("Unable to cancel tool %1 for 10 seconds. Stop it manually by your OS task manager.").arg(toolName));
```
DmitriiSukhomlinov, line 127: Wait a bit, you can try to kill it by yourself, just by using [QProcess::kill()](https://doc.qt.io/qt-5/qprocess.html#kill). I'm sure you should try to do it and if you failed, then you text this message to user

DmitriiSukhomlinov (changes requested):
See the comment

src/corelibs/U2Core/src/tasks/ExternalToolRunTask.cpp
```diff
@@ -120,7 +120,16 @@ void ExternalToolRunTask::run() {
     while (!externalToolProcess->waitForFinished(1000)) {
         if (isCanceled()) {
             killProcess(externalToolProcess);
-            algoLog.details(tr("Tool %1 is cancelled").arg(toolName));
+            externalToolProcess->waitForFinished(2000); //wait 2 seconds to let OS kill process in previous method
+            if (externalToolProcess->state() == QProcess::Running) {
```
DmitriiSukhomlinov: Check here the return value of `waitForFinished()`, not the state

DmitriiSukhomlinov: Also, you should check `externalToolProcess->exitStatus()` - it should be `QProcess::NormalExit`

src/corelibs/U2Core/src/tasks/ExternalToolRunTask.cpp
```diff
@@ -120,7 +120,16 @@ void ExternalToolRunTask::run() {
...
+                externalToolProcess->kill();
+            }
+            externalToolProcess->waitForFinished(10000);
```
DmitriiSukhomlinov: The same checks

DmitriiSukhomlinov (changes requested):
See comments

DmitriiSukhomlinov (approved):

yalgaer (approved):

PR #806, opened 2022-03-18
UGENE-5421 No "Save dotplot" dialog on sequence removal

rasputinkirill:


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_5001_6000.cpp
```diff
@@ -1509,6 +1509,21 @@ GUI_TEST_CLASS_DEFINITION(test_5417) {
     CHECK_SET_ERR(!lt.hasErrors(), "Errors in log: " + lt.getJoinedErrorString());
 }
 
+GUI_TEST_CLASS_DEFINITION(test_5421) {
+    // 1. Build doplot for "data/samples/Genbank/murine.gb" and "data/samples/Genbank/sars.gb".
+    // 2. Remove sars.gb from project
+    // Expected state: Save dialog appeared.
+    GTUtilsDialog::waitForDialog(os, new DotPlotFiller(os));
+    Runnable *buildDotPlot = new BuildDotPlotFiller(os, dataDir + "samples/Genbank/sars.gb", dataDir + "samples/Genbank/murine.gb");
```
yalgaer: Do not use local variables for fillers. Filles do not survive the waitForDialogCall().

src/plugins/dotplot/src/DotPlotWidget.cpp
```diff
@@ -778,7 +779,16 @@ bool DotPlotWidget::sl_showSettingsDialog(bool disableLoad) {
 
 // ask user if he wants to save dotplot first
 void DotPlotWidget::sl_showDeleteDialog() {
```
yalgaer: Add an explicit parameter about the mode: show or not to show 'cancel' button: isCancelable = true

src/plugins/dotplot/src/DotPlotWidget.cpp
```diff
@@ -778,7 +779,16 @@ bool DotPlotWidget::sl_showSettingsDialog(bool disableLoad) {
...
-    int answer = QMessageBox::information(this, tr("Save dot-plot"), tr("Save dot-plot data before closing?"), QMessageBox::Yes, QMessageBox::No, QMessageBox::Cancel);
+    if (deleteDialogWasShownAndNotCancelled) {
```
yalgaer: Can you check this on the caller side. 
The current logic has a logical flow: a caller calls 'showDeleteDialog' and nothing happens because of some internal magic.
A better solution would be:
1. Internal logic asserts (SAFE_POINT) when dialog can't be shown
2. The caller checks if the dialog can be shown or not and shows it only if it can.

yalgaer (changes requested):
Style issues only.

yalgaer (commented):
Looks good please run GUI tests

src/plugins/dotplot/src/DotPlotWidget.cpp
```diff
@@ -777,26 +782,32 @@ bool DotPlotWidget::sl_showSettingsDialog(bool disableLoad) {
 }
 
 // ask user if he wants to save dotplot first
-void DotPlotWidget::sl_showDeleteDialog() {
-    int answer = QMessageBox::information(this, tr("Save dot-plot"), tr("Save dot-plot data before closing?"), QMessageBox::Yes, QMessageBox::No, QMessageBox::Cancel);
-    bool saveDotPlot;
-
+void DotPlotWidget::sl_showDeleteDialog(bool isCancelable) {
+    int answer = 0;
```
DmitriiSukhomlinov, line 786: Set  `QMessageBox::StandardButton` as type

DmitriiSukhomlinov (dismissed):

src/plugins/dotplot/src/DotPlotWidget.cpp
rasputinkirill: Only integer is acceptable option.
`static int information(QWidget *parent, const QString &title,
                           const QString& text,
                           int button0, int button1 = 0, int button2 = 0);`

yalgaer (approved):

PR #811, opened 2022-03-21
UGENE-7577. [CrashHandler] Update breakpad

### LINKED ISSUE #780: Current breakpad requires newer lss (linux-syscalls-support) on some archs.
The breakpad update broke some architectures (e.g. i586), as it removed some compat code. (e.g. it always uses the mmap2 syscall instead of the older mmap).

Please pull a newer lss into the repo.

### LINKED ISSUE #781: Linux with glibc >= 2.34 requires updated (or patched) breakpad
From https://chromium.googlesource.com/breakpad/breakpad/+/605c51ed96ad44b34c457bbca320e74e194c317e

> On glibc > 2.33, `SIGSTKSZ` might not be constant (in which case
> it expands to a call to `sysconf` which returns a `long int`); see
> https://sourceware.org/pipermail/libc-alpha/2020-October/118513.html
> 
> Pass unsigned explicitly to std::max, to avoid relying on template
> argument deduction. This works both with the old-style constant
> `SIGSTKSZ` and the new configurable one.

Please cherry-pick the commit above, or do a bulk update of breakpad (+lss).

### LINKED ISSUE #796: error: no matching function for call to ‘max(int, long int)’
when building ugene 42.0, I got this error:

```text
g++ -c -pipe -Wall -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -std=gnu++1y -w -D_REENTRANT -flto -fno-fat-lto-objects -fPIC -DUGENE_VERSION=42.0 -DQT_DISABLE_DEPRECATED_BEFORE=0x050700 -DNDEBUG -DQT_NO_DEBUG -DQT_CORE_LIB -I. -Isrc -I/usr/include/qt -I/usr/include/qt/QtCore -Irelease -I/usr/lib/qt/mkspecs/linux-g++ -o _tmp/obj/release/safe_readlink.o src/common/linux/safe_readlink.cc
src/client/linux/handler/exception_handler.cc: In function ‘void google_breakpad::{anonymous}::InstallAlternateStackLocked()’:
src/client/linux/handler/exception_handler.cc:144:49: error: no matching function for call to ‘max(int, long int)’
  144 |   static const unsigned kSigStackSize = std::max(16384, SIGSTKSZ);
      |                                         ~~~~~~~~^~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/11.2.0/bits/char_traits.h:39,
                 from /usr/include/c++/11.2.0/string:40,
                 from src/client/linux/handler/exception_handler.h:38,
                 from src/client/linux/handler/exception_handler.cc:70:
```
this is caused by:
https://github.com/ugeneunipro/ugene/blob/2ec47dd8f9b7ebf016205cdfc9453f6e3c866b6c/src/libs_3rdparty/breakpad/src/client/linux/handler/exception_handler.cc#L144
where `16384` is `int`, but `SIGSTKSZ` is `long int`, change `16384` to `16384l` should fix it.

my step to build:
download the source tarball and extract and change dir to source dir, then
```bash
qmake CONFIG+=x64 \
    UGENE_CUDA_DETECTED=0 \
    UGENE_OPENCL_DETECTED=1 \
    UGENE_USE_BUNDLED_ZLIB=0 \
    UGENE_USE_BUNDLED_SQLITE=0
make
```

evelinabis:


yalgaer (approved):
Looks good, please run GUI tests and confirm than in manual testing crash handler works as expected.

PR #813, opened 2022-03-23
UGENE-4783 Levitsky consensus is not updates on alignment modification

rasputinkirill:


src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
```diff
@@ -133,7 +133,10 @@ char MSAConsensusAlgorithmLevitsky::getConsensusChar(const MultipleAlignment& ma
 
     // count local freqs first
     QVarLengthArray<int> localFreqs(256);
-    memset(localFreqs.data(), 0, localFreqs.size() * 4);
+    for (int i = 0; i < 256; i++) {
```
yalgaer: Why did you replace the memset with this boilerplate code? If you do not link old-style memset you can use std::fill 1-liner

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
```diff
@@ -147,6 +150,9 @@ char MSAConsensusAlgorithmLevitsky::getConsensusChar(const MultipleAlignment& ma
     int selectedGlobalFreq = nSeq * ma->getLength();
     int thresholdScore = getThreshold();
     int minFreq = int(float(nSeq) * thresholdScore / 100);
+    if (minFreq == 0) {
+        minFreq = 1;
```
yalgaer: This is an incorrect value. Keep the correct one as 'double' and make ```localFreq``` below a double too

yalgaer (changes requested):
Minor issues only.
Please run GUI tests

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
rasputinkirill: I can't get what you mean. `QVarLengthArray` don't have method like 'fill', and we cant call std::fill for qt container.
About previous code, i don't like way to operate with container content throu the memory modification even it possible with container methods. At least if you want keep previous code it should be `memset(localFreqs.data(), 0, localFreqs.size() * sizeof(int));`.

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
rasputinkirill: No I don't agree with that.
The result will be the same, but algorithm logic became hard to read, and harder to calculate. Because for now we just count consensus symbol hits (integer), convert threshold value percent (double) to hits threshold (integer). But you suggest to calculate frequency (double) for every symbol and compare this with threshold value percent (double).
I can suggest another solution if you don't like current - it will work properly if we fill `localFreqs` with "-1" values when it initialized. So in that way '0' minimum frequency became valuable.

yalgaer: Frequency of  a symbol is always double: in 'ABC' string the frequency of A is 33.333333...%. 
Keep the original (source) data as long as possible and convert it in the last place.

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
yalgaer: https://en.cppreference.com/w/cpp/algorithm/fill
std::fill(localFreqs.begin(), localFreqs.end(), 0);

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.h
```diff
@@ -74,7 +74,7 @@ class U2ALGORITHM_EXPORT MSAConsensusAlgorithmLevitsky : public MSAConsensusAlgo
     virtual MSAConsensusAlgorithmLevitsky* clone() const;
 
 private:
-    QVarLengthArray<int> globalFreqs;
+    QVarLengthArray<int> globalFreqs = QVarLengthArray<int>(256);
```
yalgaer: Is there any reason for this change? Why not to assing QVarLengthArray<int>(256) directly in the constructor?

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
```diff
@@ -133,28 +133,27 @@ char MSAConsensusAlgorithmLevitsky::getConsensusChar(const MultipleAlignment& ma
...
     // find all symbols with freq > threshold, select one with the lowest global freq
     char selectedChar = U2Msa::GAP_CHAR;
-    int selectedGlobalFreq = nSeq * ma->getLength();
+    double selectedGlobalPercentage = 100;
     int thresholdScore = getThreshold();
-    int minFreq = int(float(nSeq) * thresholdScore / 100);
     for (int c = 'A'; c <= 'Y'; c++) {
-        int localFreq = freqsData[uchar(c)];
-        if (localFreq < minFreq) {
+        double localPercentage = ((double)freqsData[uchar(c)] / (double)nSeq) * 100;
```
yalgaer: nit:
1. one (double) is enough.
2. when you work with double you can remove *100 completely and work on 0..1 (share) range.

yalgaer (commented):
Both XML and GUI tests are failing. Please fix and I will approve

rasputinkirill:
I will update xml test expected file after merging with master. Because for now consensus calculated differently.
To use `std::fill` we need and array (`int[256]`) , but not `int*`. There is no method in QVarLengthArray to get an array.

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
```diff
@@ -114,11 +114,10 @@ static void registerHit(int* data, char c) {
...
 MSAConsensusAlgorithmLevitsky::MSAConsensusAlgorithmLevitsky(MSAConsensusAlgorithmFactoryLevitsky* f, const MultipleAlignment& ma, bool ignoreTrailingLeadingGaps, QObject* p)
-    : MSAConsensusAlgorithm(f, ignoreTrailingLeadingGaps, p) {
-    globalFreqs.resize(256);
-    memset(globalFreqs.data(), 0, globalFreqs.size() * 4);
-
-    int* freqsData = globalFreqs.data();
+    : MSAConsensusAlgorithm(f, ignoreTrailingLeadingGaps, p)
+    , globalFreqs(QVarLengthArray<int>(256)) {
+    int *freqsData = globalFreqs.data();
+    std::fill(freqsData, freqsData + 256, 0);
```
yalgaer: You do not need a raw pointer and magic constants.
Use the classic code I pointed in the previous message: std:fill(globalFreqs.begin(), globalFreqs.end(), 0).
Please, update all places like this in your patch.

yalgaer (changes requested):

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
```diff
@@ -114,11 +114,10 @@ static void registerHit(int* data, char c) {
...
 
 MSAConsensusAlgorithmLevitsky::MSAConsensusAlgorithmLevitsky(MSAConsensusAlgorithmFactoryLevitsky* f, const MultipleAlignment& ma, bool ignoreTrailingLeadingGaps, QObject* p)
-    : MSAConsensusAlgorithm(f, ignoreTrailingLeadingGaps, p) {
-    globalFreqs.resize(256);
-    memset(globalFreqs.data(), 0, globalFreqs.size() * 4);
-
-    int* freqsData = globalFreqs.data();
+    : MSAConsensusAlgorithm(f, ignoreTrailingLeadingGaps, p)
+    , globalFreqs(QVarLengthArray<int>(256)) {
+    int *freqsData = globalFreqs.data();
```
yalgaer, line 119: nit: freqsData is useless. You can replace it with globalFreqs 1:1

yalgaer (approved):

PR #814, opened 2022-03-25
UGENE-7117 Can not remove folder for temporary files error if cancel building tree

smlnkcv:
The problem has appeared because files was removing a bit longer that anticipated and then `rmdir` function started to remove the directory, this directory wasn't empty (for [rmdir](https://doc.qt.io/qt-5/qdir.html#rmdir) success he removing directory should be empty, see docs)


yalgaer (commented):
Looks good. Please run UI tests and I will approve

yalgaer (approved):

PR #815, opened 2022-03-29
UGENE-7573. [3d] Sas+spin+save=crash.

evelinabis:
Bug description

1. The MolecularSurfaceCalcTask task is created.
2. The MolecularSurfaceCalcTask task is deleted.
3. The SaveProjectTask task is created at the address of the pointer of the deleted MolecularSurfaceCalcTask (so the memory manager decided).
4. The sl_onTaskFinished slot receives events from all tasks and does not dynamically check for compliance with the MolecularSurfaceCalcTask class, it only checks addresses.
5. Next, methods are called on the SaveProjectTask class as on the MolecularSurfaceCalcTask class, which led to a crash.


src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.cpp
```diff
@@ -967,6 +965,7 @@ void BioStruct3DGLWidget::sl_showSurface() {
 
     QString surfaceType = qobject_cast<QAction*>(sender())->text();
     surfaceCalcTask = new MolecularSurfaceCalcTask(surfaceType, atoms);
+    connect(surfaceCalcTask, &Task::si_stateChanged, this, &BioStruct3DGLWidget::sl_onTaskFinished);
```
yalgaer: This is very unsafe to keep 'surfaceCalcTask' globally because it is automatically deleted by the Task Scheduler when it is finished. This way we have a broken pointer cached in the 'surfaceCalcTask'.

I would recommend changing 'surfaceCalcTask' to a local variable and using sl_onTaskFinished(Task* task) 'task' parameter inside of the sl_onTaskFinished to restore it value.

The sl_onTaskFinished should also get a better name like sl_onSurfaceCalcTaskFinished

yalgaer (changes requested):

yalgaer:
Closing as inactive. Please re-open when ready

src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.cpp
evelinabis: > This is very unsafe

It's wrapped in a QPointer.

> using sl_onTaskFinished(Task* task) 'task' parameter inside of the sl_onTaskFinished to restore it value.

In the previous implementation, the `sl_onTaskFinished` signal triggered for all `TaskScheduler::si_stateChanged` signals from all tasks (not just `surfaceCalcTask`) during the lifetime of the BioStruct3DGLWidget (not just `MolecularSurfaceCalcTask`). The `sl_onTaskFinished(Task*)` slot signature will not allow it to be connected to the **Task**::si_stateChanged() signal (only to **TaskScheduler**::si_stateChanged(Task *)), which means that it will not be possible to limit the "scope" of the slot and the "lifetime" of the slot.

Getting rid of the pointer is a good idea, but then I think it's better to use `sender()` and cast to `MolecularSurfaceCalcTask` in the current implementation instead of using a slot with signature `sl_onTaskFinished(Task* task)`.

src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.cpp
```diff
@@ -985,8 +988,12 @@ void BioStruct3DGLWidget::sl_selectSurfaceRenderer(QAction* action) {
     update();
 }
 
-void BioStruct3DGLWidget::sl_onTaskFinished(Task* task) {
-    if (surfaceCalcTask != task || surfaceCalcTask->getState() != Task::State_Finished) {
+void BioStruct3DGLWidget::sl_onSurfaceCalcTaskFinished(Task* task) {
+    if (task->getState() != Task::State_Finished) {
+        return;
+    }
+    auto surfaceCalcTask = qobject_cast<MolecularSurfaceCalcTask*>(task);
```
yalgaer: What if you have two 3D views and each of them will run 'MolecularSurfaceCalcTask' ?
Both views will receive this callback. You need to subscribe to the task itself, not to the global task scheduler instance.

yalgaer (changes requested):
See the comment about multiple views above

src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.cpp
evelinabis: How can I do this?

evelinabis: Do I understand correctly that you are suggesting that I connect the `Task::si_stateChanged()` signal to the `BioStruct3DGLWidget::sl_onTaskFinished(Task*)` slot? Based on your previous comment.

yalgaer: Short answer is yes, connect si_stateChanged to sl_onTaskFinished
Long and a better answer is:
1. Do not use dedicated methods for short single-use callbacks. Use lambdas
2. Use TaskSignalMapper instead of Task::si_stateChanged(). It has dedicated methods like si_taskSucceeded

src/plugins/biostruct3d_view/src/BioStruct3DGLWidget.cpp
```diff
@@ -966,7 +965,14 @@ void BioStruct3DGLWidget::sl_showSurface() {
     atoms = ctx.biostruct->getAllAtoms();
 
     QString surfaceType = qobject_cast<QAction*>(sender())->text();
-    surfaceCalcTask = new MolecularSurfaceCalcTask(surfaceType, atoms);
+    auto surfaceCalcTask = new MolecularSurfaceCalcTask(surfaceType, atoms);
+    connect(new TaskSignalMapper(surfaceCalcTask), &TaskSignalMapper::si_taskFinished, this, [this](Task* task) {
```
yalgaer: nit: use ```si_taskSucceeded``` so if the task is failed you so not have a problem.

yalgaer (changes requested):
Will approve after tests passed

yalgaer (approved):

PR #816, opened 2022-03-29
UGENE-4882 Can't select line on dotplot

rasputinkirill:


src/plugins/dotplot/src/DotPlotWidget.cpp
```diff
@@ -1497,66 +1497,52 @@ void DotPlotWidget::selectNearestRepeat(const QPointF& p) {
     nearestSelecting = false;
 }
 
+float DotPlotWidget::calculateDistance(float x, float y, const DotPlotResults& r) const {
+    // find nearest point on line segment to desired point
+    // then calculate distance
+    double t = (double)(((x - r.x) + (y - r.y)) * r.len) / (2 * pow(r.len, 2));
```
yalgaer: Could you please add a reference to the algorithm or your own detailed description?
Like this: https://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment
or this
http://csharphelper.com/blog/2016/09/find-the-shortest-distance-between-a-point-and-a-line-segment-in-c/

++ If you develop a new algorithm by yourself please add API tests because this is a very error-prone algorithm
To make API tests simpler you can make this method static and pass all required parameters.

src/plugins/dotplot/src/DotPlotWidget.cpp
```diff
@@ -1497,66 +1497,52 @@ void DotPlotWidget::selectNearestRepeat(const QPointF& p) {
...
+    t = std::min(1.0, std::max(0.0, t));
```
yalgaer: use qBound

src/plugins/dotplot/src/DotPlotWidget.cpp
```diff
@@ -1497,66 +1497,52 @@ void DotPlotWidget::selectNearestRepeat(const QPointF& p) {
...
+    float nearestPointX = r.x + t * r.len;
```
yalgaer: Min of double and float types. Can the result be double too?

yalgaer (changes requested):

src/plugins/dotplot/src/DotPlotWidget.cpp
rasputinkirill: Nearest point on line segment to desired point - https://www.cyberforum.ru/post16063107.html
Then we calculate distance between two points.

It using class data members for now, but it can be rewritten as static if you want.

src/plugins/dotplot/src/DotPlotWidget.cpp
rasputinkirill: I made all float, because whole class operates with floats.

src/plugins/dotplot/src/DotPlotWidget.cpp
rasputinkirill: done

smlnkcv (commented):
Issue is reproducible on the dotplot_for_seq.dpt attached file

src/plugins/dotplot/src/DotPlotWidget.cpp
yalgaer: Is there any reason for the t to be 'double'?
My comment is to avoid implicit (hidden) type conversion.

yalgaer: Yes lets make this function an independent (static) at least. This is a very good candidate for a GeometryUtils-like class.

src/plugins/dotplot/src/DotPlotWidget.cpp
yalgaer: Yes lets make this function an independent (static) at least. This is a very good candidate for a GeometryUtils-like class.

smlnkcv (approved):

DmitriiSukhomlinov (approved):
I don't even wanna think about sorting out this math, approved)

yalgaer (approved):

PR #817, opened 2022-03-30
UGENE-7575 Assert in MSA editor after Reset Zoom

yalgaer:


src/corelibs/U2View/src/ov_msa/ScrollController.cpp
```diff
@@ -346,9 +347,29 @@ GScrollBar* ScrollController::getVerticalScrollBar() const {
     return vScrollBar;
 }
 
-void ScrollController::sl_zoomScrollBars() {
-    zoomHorizontalScrollBarPrivate();
-    zoomVerticalScrollBarPrivate();
+void ScrollController::updateScrollBarsOnFontOrZoomChange() {
+    CHECK(!maEditor->isAlignmentEmpty(), );
+    SignalBlocker signalBlocker(hScrollBar);
```
DmitriiSukhomlinov: We use `QSignalBlocker` now, usually, `SignalBlocker` is a bit deprecated, it was used when `QSignalBlocker` wasn't implemented yet.

DmitriiSukhomlinov (dismissed):

src/corelibs/U2View/src/ov_msa/ScrollController.cpp
yalgaer: Fixed

PR #821, opened 2022-04-04
UGENE-4783_2 small fix

rasputinkirill:
Small fix for UGENE-4783, and GUI test update

### DIFF
--- a/src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
+++ b/src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
@@ -141,7 +141,8 @@ char MSAConsensusAlgorithmLevitsky::getConsensusChar(const MultipleAlignment& ma
     }
     // find all symbols with freq > threshold, select one with the lowest global freq
     char selectedChar = U2Msa::GAP_CHAR;
-    double selectedGlobalPercentage = 1.1;
+    double selectedGlobalPercentage = 1;
+    bool firstConsensusCharFound = false;
     double thresholdScore = getThreshold() / 100.0;
     for (int c = 'A'; c <= 'Y'; c++) {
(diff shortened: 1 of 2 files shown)


src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
```diff
@@ -141,7 +141,7 @@ char MSAConsensusAlgorithmLevitsky::getConsensusChar(const MultipleAlignment& ma
     }
     // find all symbols with freq > threshold, select one with the lowest global freq
     char selectedChar = U2Msa::GAP_CHAR;
-    double selectedGlobalPercentage = 1;
+    double selectedGlobalPercentage = 1.1;
```
yalgaer: Could you please keep 1 (100%) and use the following comparison below:
```
if (globalPercentage <= selectedGlobalPercentage) {
```

yalgaer (changes requested):

src/corelibs/U2Algorithm/src/util_msa_consensus/MSAConsensusAlgorithmLevitsky.cpp
rasputinkirill: No its not correct, we need to set first found (by alphabet) among the symbols with the same value of frequency to consensus. If you want to keep `double selectedGlobalPercentage = 1;` i can suggest to introduce `bool firstHit` variable.

yalgaer: Ok, lets add 'isFirstHit/Match' variable. This way it will be very explicit what is going on and there will be no magic constants.

yalgaer (changes requested):
Check the comments, will approve after you implement the way you described.

yalgaer (approved):

PR #825, opened 2022-04-06
UGENE-7584 Spidey alignment fails if delete created annotation until …

DmitriiSukhomlinov:
…process is finished


yalgaer (changes requested):
We should avoid showing dialogs/message boxes from within TaskScheduler callbacks: such dialogs stop the event loop - it means that no other tasks are processed while the dialog is shown at all - all tasks are blocked.

Even worse - this may lead to a dead-locks: when a task is waited to be finished inside of the event loop created by a the dialog started in prepare/report/onSubtaskFinished etc..

An acceptable simple solution might be either canceling the task completely or creating a new annotation table with an auto-generated name.

More complex option is to handle the task results outside of the TaskScheduler callbacks: wait for the task to finish and run results processing code (with dialogs) later.

DmitriiSukhomlinov:
> We should avoid showing dialogs/message boxes from within TaskScheduler callbacks: such dialogs stop the event loop - it means that no other tasks are processed while the dialog is shown at all - all tasks are blocked.
> 
> Even worse - this may lead to a dead-locks: when a task is waited to be finished inside of the event loop created by a the dialog started in prepare/report/onSubtaskFinished etc..
> 
> An acceptable simple solution might be either canceling the task completely or creating a new annotation table with an auto-generated name.
> 
> More complex option is to handle the task results outside of the TaskScheduler callbacks: wait for the task to finish and run results processing code (with dialogs) later.

I agree, fixed

src/plugins_3rdparty/primer3/src/primer3_core/primer3.h
```diff
@@ -176,7 +176,7 @@ typedef struct primrec {
  * returns a pair of primers.)
  */
 typedef struct primpair {
-  double pair_quality;
```
yalgaer: Please remove unrelated changes.

src/plugins/external_tool_support/src/spidey/SpideySupportTask.cpp
```diff
@@ -221,24 +225,46 @@ void PrepareInputForSpideyTask::run() {
...
-        return res;
+        return {};
     }
 
     if (subTask == spideyAlignmentTask) {
+        if (aObj == nullptr) {
+            taskLog.details("The result of the \"SpideySupportTask\" should be saved to the annotation table, which has been removed. The new annotation table is about to be created.");
+
+            auto dnaSeq = settings.getCDnaSequence();
+            auto documentFilePath = GUrlUtils::getDefaultDataPath() + "/MyDocument.gb";
```
yalgaer, line 249: Please roll the result file name or we can overwrite user data.

yalgaer (changes requested):
Minor issues only. Overall the solution is correct.

src/plugins/external_tool_support/src/spidey/SpideySupportTask.cpp
DmitriiSukhomlinov: You won't override the user data in any case - even if you write to the existed "MyDocument.gb" and it has an annotation with the same name. There is a point, that you create the same document as a default one.

src/plugins_3rdparty/primer3/src/primer3_core/primer3.h
DmitriiSukhomlinov: Fixed

src/plugins/external_tool_support/src/spidey/SpideySupportTask.cpp
yalgaer: This is how I see the problem: 

1. You have an existing document that is not added to the project: /var/MyDocument.gb
2. You create the same in-memory document with a new data and add it to the project as 'loaded': /var/MyDocument.gb

What will happen next? 
I bet the newly created document will override the old one and this means a data loss for a user?

yalgaer (changes requested):
Please check the comment about the document overwrite

src/plugins/external_tool_support/src/spidey/SpideySupportTask.cpp
DmitriiSukhomlinov: Fixed

yalgaer (commented):
Good, will approve after GUI tests.

DmitriiSukhomlinov:
@yalgaer all tests have passed

yalgaer (approved):

PR #827, opened 2022-04-07
UGENE-5730 Exporting to existing and opened document

rasputinkirill:


yalgaer (changes requested):
The fix is very different from what is described in the bug.
If a user wants to overwrite the file UGENE should overwrite.

rasputinkirill:
> The fix is very different from what is described in the bug. If a user wants to overwrite the file UGENE should overwrite.

File will be overwritten in current version. I just changed 'error' type of message and removed test which tests corresponding error message. Because that old test case check test behavior that not actual in current version.

yalgaer (approved):
Thank you for the details. Approved

PR #829, opened 2022-04-13
UGENE-6703 MSA: Using "Remove all gaps" action creates undo step for alignment without gaps

rasputinkirill:


src/corelibs/U2Core/src/gobjects/MultipleAlignmentObject.cpp
```diff
@@ -737,6 +737,26 @@ void MultipleAlignmentObject::releaseState() {
     }
 }
 
+bool MultipleAlignmentObject::isGapRemovePossible() const {
```
yalgaer: nit: rename it to hasGaps()

src/corelibs/U2Core/src/gobjects/MultipleAlignmentObject.cpp
```diff
@@ -737,6 +737,26 @@ void MultipleAlignmentObject::releaseState() {
...
+        if (vector.size() > 1) {
+            //if we have more than one set of gaps in a row - they are 100% removable
+            return true;
+        }
+    }    
+    for (int i = 0; i < getRowCount(); i++) {
+        //if remaining gaps isn't trailing we can remove them too
+        int gapLength = listGapModel[i].size() == 0 ? 0 : listGapModel[i][0].length;
+        const MultipleAlignmentRow &row = getRow(i);
+        qint64 rowLengthWithoutTrailing = row->getRowLengthWithoutTrailing();
```
yalgaer: Gap model never has trailing gaps or U2MsaGaps with zero length. Why the gapModel.isEmpty() check is not enough?

rasputinkirill: In this case we will skip situation when there is no trailing gaps in a row but it have one set of gaps in the middle of it.

yalgaer: This method looks very ambiguous: which gaps do you mean

1. Leading: the first gap in the gap model starts with 0
2. Trailing: sum of all gap lengths + core-length < alignment-length
3. Middle gaps: gap model size > 1 || gaps[0].startPos > 0
4. Any visible gap in the view: coreLength < alignmentLength

What kind if gaps do you want to check here? 
I think all removable: leading + middle.
 In this case please rename the method to hasNonTrailingGap and implement it like !gapModel.isEmpty()

?

rasputinkirill: You not right about point 3, because it is possible have row with only middle gaps.
Anyway i updated this method and test, with leading gaps test case.

yalgaer: I think 3 is right: you should have either at least 2 gap of the first gap starting with > 0 offset.

src/corelibs/U2Core/src/gobjects/MultipleAlignmentObject.cpp
```diff
@@ -737,6 +737,21 @@ void MultipleAlignmentObject::releaseState() {
     }
 }
 
+bool MultipleAlignmentObject::hasNonTrailingGap() const {
+    const QList<QVector<U2MsaGap>> &listGapModel = getGapModel();
+    for (int i = 0; i < getRowCount(); i++) {
+        if (listGapModel[i].isEmpty()) {
```
yalgaer: ```!istGapModel[i].isEmpty()``` alone is enough to check if the row has non-trailing gaps.

src/corelibs/U2Core/src/gobjects/MultipleAlignmentObject.h
```diff
@@ -173,6 +173,10 @@ class U2CORE_EXPORT MultipleAlignmentObject : public GObject {
 
     void saveState();
     void releaseState();
+    /*
+    * Returns true if there is removable gaps in alignment.
```
yalgaer: nit: The comment does not correlate with the method name and is focused on a specific scenario.

yalgaer (changes requested):

src/corelibs/U2Core/src/gobjects/MultipleAlignmentObject.cpp
yalgaer: Here you check that the row has no 'inner' gaps.
There are 3 types of gaps: leading, inner, trailing.
What you need to check here is that there are either leading or inner gaps. In this case you can remove gaps from the row.

To check if you can remove any gaps you do not need any method at all: use ```!getGaps().isEmpty()```

src/corelibs/U2Core/src/datatype/msa/MultipleAlignment.cpp
```diff
@@ -136,13 +136,8 @@ int MultipleAlignmentData::getRowCount() const {
 
 QList<QVector<U2MsaGap>> MultipleAlignmentData::getGapModel() const {
     QList<QVector<U2MsaGap>> gapModel;
-    const int alignmentLength = getLength();
```
yalgaer, line 139: Good change, so we have no discrepancy between MSA & Row gap models

src/corelibs/U2Core/src/gobjects/MultipleAlignmentObject.cpp
```diff
@@ -737,6 +737,16 @@ void MultipleAlignmentObject::releaseState() {
     }
 }
 
+bool MultipleAlignmentObject::hasNonTrailingGap() const {
+    const QList<QVector<U2MsaGap>> &listGapModel = getGapModel();
```
yalgaer, line 741: nit: use std::any_of 1-liner

yalgaer (commented):
Looks good, will approve after the testing

yalgaer (approved):

PR #830, opened 2022-04-13
UGENE-7555. Assembly has no mapped reads

yalgaer:
See JIRA issue for more details.


src/corelibs/U2View/src/ov_assembly/AssemblyBrowserFactory.cpp
```diff
@@ -79,7 +80,31 @@ Task* AssemblyBrowserFactory::createViewTask(const MultiGSelection& multiSelecti
...
-        foreach (GObject* o, asmObjects) {
+        // Move up to MAX_VIEWS assembly objects with reads to the start of the list.
+        // So they will be selected to be opened with a higher priority than empty assemblies.
+
+        // Select assembly objects with reads first.
+        QList<GObject*> assemblyObjectsWithReads;
+        for (GObject* o : qAsConst(asmObjects)) {
+            if (auto assemblyObject = qobject_cast<AssemblyObject*>(o)) {
+                U2OpStatusImpl os;
+                if (assemblyObject->getReadCount(os) > 0) {
```
DmitriiSukhomlinov, line 91: You need to check the `os` status

src/corelibs/U2Core/src/gobjects/AssemblyObject.h
```diff
@@ -34,6 +34,9 @@ class U2CORE_EXPORT AssemblyObject : public GObject {
 public:
     AssemblyObject(const QString& objectName, const U2EntityRef& ref, const QVariantMap& hints = QVariantMap());
 
+    /** Returns number of reads in the object. */
+    qint64 getReadCount(U2OpStatus& os) const;
```
DmitriiSukhomlinov, line 38: `getReadsCount`

DmitriiSukhomlinov (approved):

src/corelibs/U2Core/src/gobjects/AssemblyObject.h
yalgaer: Nope, getReadCount is the correct variant.

src/corelibs/U2View/src/ov_assembly/AssemblyBrowserFactory.cpp
yalgaer: This check is not needed because we do not want to interrupt the whole process because of the local optimization: if read count is not available for any reason here we continue as no-op.

PR #834, opened 2022-04-21
UGENE-4500 Edit sequence and save to new file with annotations works incorrectly

rasputinkirill:


src/corelibs/U2Core/src/tasks/ModifySequenceObjectTask.cpp
```diff
@@ -200,6 +200,9 @@ void ModifySequenceContentTask::cloneSequenceAndAnnotations() {
         }
     }
     docs.append(newDoc);
+    if (AppContext::getProject() != nullptr) {
```
yalgaer: Could you please explain how this line changes incorrect annotation coordinates?

rasputinkirill: Check `FixAnnotationsUtils::fixAnnotations()` method, it can't find target annotation table if document not in the project. So as document will be added to project after editing anyway, i decided to add it right after creation.
We can do it in other way - remove condition `if (AppContext::getProject() != nullptr)` from `FixAnnotationsUtils::fixAnnotations()` and always search 'docs' for target annotation table.
As for me there is no much difference.

yalgaer: I see, 
the good solution would be remove any dependency on project state from FixAnnotationUtils: pass all loaded annotation table objects in project as a parameter to it. 

This way it will not depend if the document is added to the project or not.

yalgaer (commented):
Please run all tests.

My comment about "a better fix" is optional, but it will be nice if you can improve UGENE code this way.

yalgaer (changes requested):
I think the current solution is not correct: you do really need to update all known annotations table associated with the modified sequence object.
All known here: all in the project + pending (that will be added to the project later, after saving)

The correct solution is to collect all annotation tables into the list and pass them to the FixAnnotationsUtils method. 

This way FixAnnotationsUtils will be independent from the AppContext::getProject() state.

rasputinkirill:
I did it in other way.
I removed default value for `QList<Document*> docs` parameter, so caller should provide  to annotation fixer documents which can contain related tables.

yalgaer (commented):
Looks good. Will approve once GUI tests are finished

yalgaer (approved):

PR #842, opened 2022-05-08
UGENE-7608. CreateAnnotations task may be unsafe when creates annotations in a separate thread

yalgaer:
Out of all places the CreateAnnotationsTask is used there is only 1 single  place (variations) where we can run it in a separate thread.
A potential alternative solution is to completely replace the CreateAnnotationTask with an utility method


src/corelibs/U2Core/src/models/GObjectReference.h
```diff
@@ -40,7 +40,7 @@ class U2CORE_EXPORT GObjectReference {
         : docUrl(_docUrl), objName(_objName), entityRef(_entityRef), objType(_type) {
     }
 
-    GObjectReference(const GObject* obj, bool deriveLoadedType = true);
+    explicit GObjectReference(const GObject* obj, bool deriveLoadedType = true);
```
yalgaer, line 43: I made this explicit because for the constructors in the CreateAnnotationTask it was ambiguous which one is called: with GObject* or GObjectReference as the first parameter.

PR #844, opened 2022-05-14
UGENE-7454. [Sv] Crash when removing current sequence.

evelinabis:


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1662,6 +1663,39 @@ GUI_TEST_CLASS_DEFINITION(test_7451) {
...
+    //     Expected: no crash.
+    GTUtilsProject::openFile(os, dataDir + "samples/PDB/1CF7.PDB");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    QPoint splitterCenter = [&os] {
+        auto projectView = GTWidget::findWidget(os, "project_view");
+        QRect rect = projectView->geometry();
+        return projectView->mapToGlobal({rect.right() + 4, rect.center().y()});
+    }();
+    auto delta = [&os]() -> QPoint {
```
yalgaer: replace the following 5 lines with a new method ```GTMainWindow::getMainWindowWidgetByName(os, name)``` 1-liner.
This method may be re-usable in other places.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1662,6 +1663,39 @@ GUI_TEST_CLASS_DEFINITION(test_7451) {
...
+    // Increase the width of the Project View.
+    // Open any tab in the Options Panel and increase its width.
+    //     Expected: the current Sequence View is narrow.
+    // Find the action toolbar extension for the first sequence in the Sequence View (">>" button). Press ">>"->
+    //         "X Remove sequence".
+    //     Expected: no crash.
+    GTUtilsProject::openFile(os, dataDir + "samples/PDB/1CF7.PDB");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    QPoint splitterCenter = [&os] {
```
yalgaer: Replace the following 5 lines with 2 lines line
```
QRect rect = GTWidget::findWidget(os, "project_view")->geometry();
QPoint splitterCenter = projectView->mapToGlobal({rect.right() + 4, rect.center().y()});
```

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -736,7 +737,8 @@ void AnnotatedDNAView::removeSequenceWidget(ADVSequenceWidget* sequenceWidget) {
     updateMultiViewActions();
     emit si_sequenceWidgetRemoved(sequenceWidget);
     scrolledWidgetLayout->removeWidget(sequenceWidget);
-    delete sequenceWidget;  // v->deleteLater(); //problem: updates for 'v' after seqCtx is destroyed
+    sequenceWidget->disconnect();
```
yalgaer: This may be not enough, because sequenceWidget may contain other widgets that are still connected.
A simpler solution could be the following (delete widget out of the event processing loop)
```
void ADVSingleSequenceWidget::sl_closeView() {
    QTimer::singleShot(0, this, [this] { closeView(); });
}
```

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -1662,6 +1663,39 @@ GUI_TEST_CLASS_DEFINITION(test_7451) {
     GTUtilsStartPage::checkRecentListUrl(os, "test_7451.fa", false);
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7454) {
+    // Open data/samples/PDB/1CF7.PDB.
+    // Increase the width of the Project View.
+    // Open any tab in the Options Panel and increase its width.
```
yalgaer: Isn't increasing of project view width alone enough? Why do you need to open the options panel at all?

yalgaer (changes requested):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: ❓

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
evelinabis: I will do the suggested solution. But I don't understand: does your solution change something in terms of "This may be not enough, because sequenceWidget may contain other widgets that are still connected"? In my head, singleShot+closeView doesn't cancel other connections either.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: OP opening removed.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: You are using 'heavy artillery' for the trivial code: 2 lines or the trivial code looks more reasonable (and readable) than 5 lines of the modern one. 

The modern code is very good when it solves some problem: reduces boilerplate, improves readability.

Here the modern code only improves the 'visibility scope' which is not problem at all if you have only 2-liner like in my example.

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
yalgaer: > "This may be not enough, because sequenceWidget may contain other widgets that are still connected" 

This is a problem related to your solution: you disconnect some data structures to remove them later, but not all of them (nested structures may still be connected and react to signals when the parent structure is already disconnected).

In the solution I suggested this problem does not exist: you do not disconnect at all: you atomically delete the whole structure.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: My fault: I didn't clarify the question.
Written strings are not compiled. Choose your preferred option:
```
QRect rect = GTWidget::findWidget(os, "project_view")->geometry();
QPoint splitterCenter = GTWidget::findWidget(os, "project_view")->mapToGlobal({rect.right() + 4, rect.center().y()});
```
or
```
QWidget* projectView = GTWidget::findWidget(os, "project_view");
QRect rect = projectView->geometry();
QPoint splitterCenter = projectView->mapToGlobal({rect.right() + 4, rect.center().y()});
```

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
evelinabis: Disconnect is not needed in my solution either. This is an attempt to understand the comment `//problem: updates for 'v' after seqCtx is destroyed`.

evelinabis: ✓

DmitriiSukhomlinov (dismissed):

smlnkcv (dismissed):

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
```diff
@@ -469,7 +469,7 @@ bool AnnotatedDNAView::onObjectRemoved(GObject* o) {
             emit si_sequenceRemoved(seqCtx);
             seqContexts.removeOne(seqCtx);
             removeAutoAnnotations(seqCtx);
-            delete seqCtx;
+            seqCtx->deleteLater();
```
yalgaer: This should not be needed anymore, right? 
The only fix needed is that QTimer line that moves the deletion of the the nested event processing.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: The first one is better because it is shorter and 
1) findWidget will never return nullptr
2) you will rarely debug this code because debugging of UI tests may 1) break your OS (like some button is pressed and have not been released yet) or 2) will switch the active window so the test can't continue.

src/corelibs/U2View/src/ov_sequence/AnnotatedDNAView.cpp
evelinabis: Both options work, I like `deleteLater`, but I removed it.

✓

yalgaer (approved):

PR #851, opened 2022-05-25
UGENE-7550. Crash 101 notifications

evelinabis:


smlnkcv (dismissed):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2391,6 +2394,30 @@ GUI_TEST_CLASS_DEFINITION(test_7548) {
     }
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7550) {
+    auto clickFindProducts = [&os] {
+        if (GTUtilsMdi::activeWindow(os)->objectName() != "123 [cant_translate.fa]") {
```
yalgaer: 1. Do not close windows. If you need your window to be active: activate it like this
```GTUtilsMdi::activateWindow ```

+ I think you need to activate it only once, not for every notification stack close event? This will make this whole lambda a 1 liner: ```GTUtilsOptionPanelSequenceView::pressFindProducts(os);```

2. Remove lambda: this is a single-use 1-2-lines function. 
It breaks 'reading order' (I need to read here, next there, etc...) and does not solve either 'single-code-multi-use' or 'big-code-chunk-with-a-name' problems.

File: src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp [hunk 2]
```diff
@@ -2391,6 +2394,30 @@ GUI_TEST_CLASS_DEFINITION(test_7548) {
...
+    // Set "AAAAAAAAAAAAAAA" as forward and reverse primers.
+    // Click "Find product(s) anyway" 100 times.
+    // Click "Find product(s) anyway" again.
+    //     Expected state: the number of notifications is "99+", there is no crash.
+    GTUtilsMdi::closeActiveWindow(os);
+    GTFileDialog::openFile(os, testDir + "_common_data/fasta/cant_translate.fa");
+    GTUtilsSequenceView::checkSequenceViewWindowIsActive(os);
+    GTUtilsOptionPanelSequenceView::setForwardPrimer(os, "AAAAAAAAAAAAAAA");
+    GTUtilsOptionPanelSequenceView::setReversePrimer(os, "AAAAAAAAAAAAAAA");
+    for (auto notificationStack = AppContext::getMainWindow()->getNotificationStack(); notificationStack->count() < 100;
```
yalgaer: Do not access UI state from the test thread: this is unsafe because UI model changes concurrently from another (UI) thread. This is one of the main reason why our tests are flaky.

You need to add helper methods need with 'runInMainThread()' around the UI accessors call them from the test method.

src/ugeneui/src/task_view/TaskStatusBar.cpp
```diff
@@ -34,6 +34,13 @@
 
 namespace U2 {
 
+// Returns a string with the number of notifications, taking into account stack limits.
+static QString notificationCountToString(int notificationCount) {
+    return notificationCount < NotificationStack::MAX_STACK_SIZE
```
yalgaer, line 39: So you want to show something like ```99+``` instead of ```100+``` here?
I would show '100+' instead.

yalgaer (changes requested):
The fix is good. 
Please update the test to avoid access to UI state from the test thread.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: > 1. Do not close windows. If you need your window to be active: activate it like this
>    `GTUtilsMdi::activateWindow `

Why?
> activate it only once, not for every notification stack close event

What does it mean? What is the notification stack close event? I close a window when a notification is clicked instead of the "Find product(s)" button.

>    It breaks 'reading order' (I need to read here, next there, etc...)

If the lambda were written just before the loop would it improve readability?

> does not solve either 'single-code-multi-use' or 'big-code-chunk-with-a-name' problems.

Why?

src/ugeneui/src/task_view/TaskStatusBar.cpp
evelinabis: image

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: Using `GTUtilsMdi::activateWindow` will slow down the test.

evelinabis: Can't click 101 times when using `GTUtilsMdi::activateWindow`.

evelinabis: Okay, I understand the problem. I don't know how it is on Linux and Mac, but on Windows with my screen resolution it goes like this:
1. The "Find product(s)" button is clicked 10 times.
2. On the 11th, not the button is pressed, but a notification.
3. The "In Silico PCR" task report window opens.
4. Notifications disappeared because a new window opened.
5. The `GTUtilsMdi::activateWindow` method is called. It tries to open a window via MainMenuBar. The mouse clicked on "&Window".
6. Notifications appear because 1 notification is outdated and the number of notifications displayed on the screen has changed: they need to be redrawn.
7. Appeared notifications interrupt the called submenu.
8. The mouse keeps moving as if the submenu is open, but nothing really happens: the windows don't switch.

I can try switching between tabbed windows using Ctrl+Tab. Or close them.

evelinabis: Lambda removed.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: Main thread scenario added, please check the correctness.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: > Why?

The test should have very precise logic: not "close whatever is opened", but "close window X", "activate window Y". This way the test will be more sensitive to changes.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2391,6 +2394,37 @@ GUI_TEST_CLASS_DEFINITION(test_7548) {
     }
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7550) {
+    class Click101TimesScenario : public CustomScenario {
+    public:
+        void run(GUITestOpStatus& os) override {
+            for (auto notificationStack = AppContext::getMainWindow()->getNotificationStack();
+                 notificationStack->count() < 100;) {
+                if (GTUtilsMdi::activeWindow(os)->objectName() != "123 [cant_translate.fa]") {
```
yalgaer: Why do you need ```if``` here? 
The window is either present or not and only one way is the correct, right?

yalgaer (commented):
Looks, good, lets finish the discussion about conditional window closing and I will approve.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: > Why do you need `if` here?

If the current window is not our sequence, then this window should be closed so that the sequence window is activated and it is possible to click the "Find product(s)" button.
> The window is either present or not and only one way is the correct, right?

Yes.

yalgaer: So the window in the test must always be closed right?

evelinabis:
Yesterday's [problem](https://github.com/ugeneunipro/ugene/pull/851#discussion_r882487442) is not reproduced for me today. Let me implement the test via `GTUtilsMdi::activateWindow`, but that doesn't change the problem from the comment: you can manually check that notifications interrupt an open submenu on Windows -- the test runs the risk of being flaky.
I can try to implement workarounds. If "switching between windows" is important, I can try to implement another way to switch between tabs with the mouse without using the menu bar.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: What is "the window"? It is important that one window -- the "123 [cant_translate.fa]" sequence window -- be active during the entire test. How this condition is achieved does not matter (as long as it works decently): i.e. I don't care if extra windows close, switch, or don't appear at all.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2391,6 +2394,35 @@ GUI_TEST_CLASS_DEFINITION(test_7548) {
...
+        void run(GUITestOpStatus& os) override {
+            for (auto notificationStack = AppContext::getMainWindow()->getNotificationStack();
+                 notificationStack->count() < 100;) {
+                GTUtilsMdi::activateWindow(os, "123 [cant_translate.fa]");
+                GTUtilsOptionPanelSequenceView::pressFindProducts(os);
+            }
+            for (int i = 0; i < 3; i++) {
+                GTUtilsMdi::activateWindow(os, "123 [cant_translate.fa]");
+                GTUtilsOptionPanelSequenceView::pressFindProducts(os);
+            }
```
evelinabis, line 2409: Sometimes need to click 102 times to reproduce.

yalgaer (approved):

PR #854, opened 2022-05-31
UGENE-7580 Portable version should keep the ".ini" file in the UGENE directory

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7580


yalgaer:
This solution won't be approved. Please check the JIRA issue for the discussion.

PR #860, opened 2022-06-06
Upgrade GL2PS version from v1.243 (2006) to the latest v1.4.2 (2020)

yalgaer:
We already have a GUI test that will check that export is not broken


DmitriiSukhomlinov:
What test?

yalgaer:
> What test?

The test that checks that UGENE correctly exports PDF documents.

PR #866, opened 2022-06-14
UGENE-7619. Minor issue in warning if "Use pattern name" option during annotation creation is selected

yalgaer:


src/corelibs/U2View/src/ov_sequence/find_pattern/FindPatternWidget.cpp
```diff
@@ -678,15 +678,19 @@ void FindPatternWidget::updateErrorLabelState() {
                               messageFlagMap.contains(SequenceIsTooBig) ||
                               (messageFlagMap.size() == 1 && messageFlagMap.contains(PleaseInputAtLeastOneSearchPatternTip));
     if (hasNoPatternErrors) {
-        GUIUtils::setWidgetWarning(textPattern, false);
+        GUIUtils::setWidgetWarningStyle(textPattern, false);
     }
 }
 
 QString FindPatternWidget::buildErrorLabelHtml() const {
+    QString errorColor = Theme::errorColorLabelHtmlStr();
```
DmitriiSukhomlinov, line 686: `static const QString ERROR_COLOR` and so on below

DmitriiSukhomlinov:
GUI tests?

src/corelibs/U2View/src/ov_sequence/find_pattern/FindPatternWidget.cpp
yalgaer: This is wrong. Theme is a dynamic object and should be re-read on use.

yalgaer:
> GUI tests?

This case is already covered by test_3270. See the whole patch.

PR #868, opened 2022-06-21
UGENE-7530 Improve the result table

rasputinkirill:


src/plugins/pcr_primer_design_for_dna_assembly/src/PCRPrimerDesignForDNAAssemblyOPWidget.cpp
```diff
@@ -150,15 +152,21 @@ PCRPrimerDesignForDNAAssemblyOPWidget::PCRPrimerDesignForDNAAssemblyOPWidget(Ann
     connect(annDnaView->getActiveSequenceContext()->getSequenceObject(), &U2SequenceObject::si_sequenceChanged, this,
             &PCRPrimerDesignForDNAAssemblyOPWidget::sl_sequenceModified);
     connect(productsTable, SIGNAL(doubleClicked(const QModelIndex &)), SLOT(sl_extractProduct()));
+    connect(pbExportProduct, SIGNAL(clicked()), SLOT(sl_extractProduct()));
```
yalgaer: nit: use compile time safe function pointers for signals

src/plugins/pcr_primer_design_for_dna_assembly/src/options_panel/ResultTable.h
```diff
@@ -36,26 +36,32 @@ struct ResultTableData {
     AnnotatedDNAView *associatedView = nullptr;
     AnnotationGroup *associatedGroup = nullptr;
     QList<U2Region> currentProducts;
+    QPair<int, int> primerLengths;
```
yalgaer, line 39: Please document what is the first/second in the pair here and below (getSelectedFragment).
I would recommend not using Pair in the public API, because it is very unreadable (requires a lot of comments)

src/plugins/pcr_primer_design_for_dna_assembly/src/tasks/ExtractPrimerTask.cpp
```diff
@@ -89,7 +135,7 @@ void ExtractPrimerTask::run() {
                 backboneStrand = U2Strand(U2Strand::Complementary);
             }
             productSequence.seq.prepend(settings.backboneSequence);
-            fragmentRegion.startPos =+ settings.backboneSequence.length();
+            fragmentRegion.startPos = +settings.backboneSequence.length();
```
yalgaer, line 139: question: why do you have '+' here?

rasputinkirill: We need to set new start position for product because we have backbone sequence before product.

yalgaer:
Closing as inactive. Please re-open when work is resumed.

PR #869, opened 2022-06-21
UGENE-7540 SAFE_POINT if you have 2 annotations with similar names

rasputinkirill:
I removed this `QHash<QString, QList<PVRowData*>> rowByName;` because stored information was redundant, all necessary information was already stored in data-members. 
Also in case of rename of annotation that data wasnt update it was case of crash.


src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -74,6 +74,15 @@ bool PVRowData::fitToRow(const QVector<U2Region>& location) {
     return true;
 }
 
+bool PVRowData::containsAnnotationWithName(const QString& name) const {
+    for (Annotation* a : annotations) {
```
yalgaer: Every row may contain only annotations with the same name. It means that you do not need to iterate all of them (there may be 10k+ annotations in a row like enzymes) and only need to check the first in the list.

yalgaer (changes requested):
Please add UI test that checks that annotation was moved to another row after it was renamed

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
rasputinkirill: done

rasputinkirill:
> Please add UI test that checks that annotation was moved to another row after it was renamed

Annotation not moved because it already in the different row after manual creation of loading from file.
This is same behavior as in previous version.

yalgaer:
> > Please add UI test that checks that annotation was moved to another row after it was renamed
> 
> Annotation not moved because it already in the different row after manual creation of loading from file. This is same behavior as in previous version.

I see, OK, thank you.

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -74,6 +76,12 @@ bool PVRowData::fitToRow(const QVector<U2Region>& location) {
     return true;
 }
 
+bool PVRowData::containsAnnotationWithName(const QString& name) const {
```
yalgaer: You do not need this method at all, because row->key is the name of the annotations inside of the row

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -86,22 +94,12 @@ void PVRowsManager::addAnnotation(Annotation* a) {
...
-    bool isRestrictionSite = data->type == U2FeatureTypes::RestrictionSite;
-    if (isRestrictionSite) {
-        name = QObject::tr("Restriction Site");
-    } else {
-        name = data->name;
-    }
-
-    if (rowByName.contains(name)) {
-        foreach (PVRowData* row, rowByName[name]) {
+    QString name = data->type == U2FeatureTypes::RestrictionSite ? PVRowData::RESTRICTION_SITE_NAME : data->name;
```
yalgaer, line 91: This code is duplicated: please extract into getAnnotationRowName(Annotation*)

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -74,6 +76,12 @@ bool PVRowData::fitToRow(const QVector<U2Region>& location) {
     return true;
 }
 
+bool PVRowData::containsAnnotationWithName(const QString& name) const {
+    SAFE_POINT(!annotations.isEmpty(), "No annotations in the row", false);
+    QString annotationName = annotations.first()->getData()->type == U2FeatureTypes::RestrictionSite ? RESTRICTION_SITE_NAME : annotations.first()->getData()->name;
+    return annotationName == name;
+}
+
 inline bool compare_rows(PVRowData* x, PVRowData* y) {
```
yalgaer: nit make it static

src/corelibs/U2View/src/ov_sequence/PanView.cpp
```diff
@@ -308,7 +308,7 @@ void PanView::sl_onAnnotationSettingsChanged(const QStringList& changedSettings)
     AnnotationSettingsRegistry* asr = AppContext::getAnnotationsSettingsRegistry();
     foreach (const QString& name, changedSettings) {
         AnnotationSettings* as = asr->getAnnotationSettings(name);
-        bool hasRow = rowsManager->contains(name);
+        bool hasRow = rowsManager->containsAnnotationWithName(name);
```
yalgaer: nit: it is better to call this method: hasRowWithName()  because rows are named differently from annotations (enzymes)

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -172,8 +160,13 @@ int PVRowsManager::getRowCount() const {
     return rows.size();
 }
 
-bool PVRowsManager::contains(const QString& key) const {
-    return rowByName.contains(key);
+bool PVRowsManager::containsAnnotationWithName(const QString& name) const {
+    for (PVRowData* row : rows) {
```
yalgaer: use qAsConst or even a 1-liner from <algorithm>

yalgaer (changes requested):
Minor issues only. Will approve after they are fixed.

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -28,6 +28,8 @@
 
 namespace U2 {
 
+const QString PVRowData::RESTRICTION_SITE_NAME = QObject::tr("Restriction Site");
```
DmitriiSukhomlinov, line 31: You haven't updated the translation file

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
```diff
@@ -86,22 +94,12 @@ void PVRowsManager::addAnnotation(Annotation* a) {
...
-        name = QObject::tr("Restriction Site");
-    } else {
-        name = data->name;
-    }
-
-    if (rowByName.contains(name)) {
-        foreach (PVRowData* row, rowByName[name]) {
+    QString name = data->type == U2FeatureTypes::RestrictionSite ? PVRowData::RESTRICTION_SITE_NAME : data->name;
+    if (containsAnnotationWithName(name)) {
+        foreach (PVRowData* row, rows) {
```
DmitriiSukhomlinov: Never use `foreach`, range based for only

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
rasputinkirill: done

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
rasputinkirill: Now its not duplicated

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
rasputinkirill: this code was removed

src/corelibs/U2View/src/ov_sequence/PanViewRows.cpp
rasputinkirill: Translations will be updated massively before release

yalgaer (approved):

DmitriiSukhomlinov (approved):

PR #870, opened 2022-06-23
UGENE-7618 Crash reading gff file with long line

rasputinkirill:


src/corelibs/U2Formats/src/GFFFormat.cpp
```diff
@@ -274,11 +271,12 @@ void GFFFormat::load(IOAdapter* io, const U2DbiRef& dbiRef, QList<GObject*>& obj
         CHECK_EXT(!io->hasError(), os.setError(io->errorString()), );
 
         len = readLongLine(qstrbuf, io, buff, GFFFormat::LOCAL_READ_BUFFER_SIZE, os);
-
         CHECK_OP(os, );
 
         // skip empty lines
-        if (TextUtils::remove(buff.data(), len, TextUtils::WHITES) == 0) {
```
yalgaer, line 281: What is wrong with this method? If this change is not needed, please keep the old code. If the change is needed here: please fix the TextUtils because it is used all around the code - may be we have the same bug in other places.

The reason why we have TextUtils and do not use QString::simplified-like method is performance: TextUtils makes all changes in place and is used in almost all format readers.

yalgaer (changes requested):
See the comment + please run UI tests

src/corelibs/U2Formats/src/GFFFormat.cpp
rasputinkirill: In this case if we read long line which length bigger than `buff` we got only last part of long line here, bu we need to check whole line.

yalgaer: I see, the solution is good enough. Approved.

yalgaer (approved):

DmitriiSukhomlinov (approved):

PR #871, opened 2022-06-25
UGENE-4721. Tree OP is not updated on switching tree in Tree view

yalgaer:


src/corelibs/U2View/src/ov_msa/phy_tree/MSAEditorMultiTreeViewer.cpp
```diff
@@ -50,7 +50,8 @@ MSAEditorMultiTreeViewer::MSAEditorMultiTreeViewer(const QString& title, MSAEdit
 
     this->setLayout(treeAreaLayout);
 
-    connect(treeTabArea, SIGNAL(si_tabsCountChanged(int)), SIGNAL(si_tabsCountChanged(int)));
+    connect(treeTabArea, &MsaEditorTreeTabArea::si_tabsCountChanged, this, &MSAEditorMultiTreeViewer::si_tabsCountChanged);
+    connect(treeTabArea, &MsaEditorTreeTabArea::si_activeTabChanged, this, [this] { emit si_activeTreeViewChanged(); });
```
DmitriiSukhomlinov, line 54: Are you sure, that you need an extra lambda here? As far as I remember, it's possible just to set a signal as the second argument and it will be proxied. But, I used it when we used `SIGNAL` or `SLOT` macros, I'm not 100% sure, that it work in this new format

src/corelibs/U2View/src/ov_phyltree/TreeSettings.h
```diff
@@ -28,9 +28,12 @@
 #include <QVariant>
 
 namespace U2 {
-enum TREE_TYPE { DEFAULT,
-                 PHYLOGRAM,
-                 CLADOGRAM };
+
+enum TreeType {
```
DmitriiSukhomlinov, line 32: `enum class` to avoid problems with implicit conversion

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2View/src/ov_msa/phy_tree/MSAEditorMultiTreeViewer.cpp
yalgaer: Yes I know it works with SIGNAL/SLOT but not sure if it works with raw function pointers.

src/corelibs/U2View/src/ov_phyltree/TreeSettings.h
yalgaer: How to convert enum class from/to QVariant?

yalgaer: Will do it at as separate PR if you provide details

PR #874, opened 2022-07-03
Optimize FindEnzymes task cancellation

yalgaer:


DmitriiSukhomlinov (approved):

DmitriiSukhomlinov (changes requested):
Wait a bit. You added some `CHECK_OP`s to the `getResultsAsAnnotations` function, which is called from `onSubTaskFinished`. But, `onSubTaskFinished` is called from the main thread, which means, that long calculations in this could freeze GUI. Doesn't that mean. that these calculations shouldn't be here at all, because they are too much time-consuming? The should be moved to another task, to the `run` function

yalgaer:
> Wait a bit. You added some `CHECK_OP`s to the `getResultsAsAnnotations` function, which is called from `onSubTaskFinished`. But, `onSubTaskFinished` is called from the main thread, which means, that long calculations in this could freeze GUI. Doesn't that mean. that these calculations shouldn't be here at all, because they are too much time-consuming? The should be moved to another task, to the `run` function

This method is not time-consuming. I removed my checks from it completely.

PR #875, opened 2022-07-04
UGENE-7635. Notification counter shows incorrect information

yalgaer:


src/corelibs/U2Gui/src/Notification.cpp
```diff
@@ -281,32 +211,22 @@ QAction* Notification::getAction() const {
...
+    onScreenTimer.setInterval(100);
+    connect(&onScreenTimer, &QTimer::timeout, this, &NotificationStack::sl_updateNotificationState);
+    onScreenTimer.start();
 }
 
 NotificationStack::~NotificationStack() {
-    foreach (Notification* notification, notifications) {
-        delete notification;
-    }
+    qDeleteAll(floatingNotifications);
```
DmitriiSukhomlinov, line 229: You stopped deleting usual `notifications`

src/corelibs/U2Gui/src/Notification.cpp
```diff
@@ -453,4 +356,68 @@ bool NotificationStack::eventFilter(QObject* target, QEvent* event) {
...
+    dlg->setObjectName("NotificationDialog");
+    QVBoxLayout vLayout;
+    QHBoxLayout hLayout;
+    QPushButton ok;
+    QCheckBox isDelete;
+
+    ok.setText(tr("OK"));
+    isDelete.setText(tr("Remove notification after closing"));
+    isDelete.setChecked(true);
+    connect(&ok, SIGNAL(clicked()), dlg.data(), SLOT(accept()));
```
DmitriiSukhomlinov, line 381: Old `SIGNAL` and `SLOT`

src/corelibs/U2Gui/src/Notification.cpp
```diff
@@ -453,4 +356,68 @@ bool NotificationStack::eventFilter(QObject* target, QEvent* event) {
...
+    QList<Notification*> notificationsToRemove;
+    for (auto notification : qAsConst(floatingNotifications)) {
+        qint64 onScreenTimeMillis = QDateTime::currentMSecsSinceEpoch() - notification->timestamp;
+        if (onScreenTimeMillis >= Notification::ON_SCREEN_TIMEOUT_MILLIS) {
+            notificationsToRemove.append(notification);
+        }
+    }
+    CHECK(!notificationsToRemove.isEmpty(), );
+    for (auto notification : qAsConst(notificationsToRemove)) {
+        floatingNotifications.removeOne(notification);
```
DmitriiSukhomlinov, line 417: Shouldn't you use `remove(Notification*)` here?

DmitriiSukhomlinov: I got, that you call `updateOnScreenNotificationPositions()` several times in this case, but, I suppose, it's better to use the same deletion way [at this line](https://github.com/ugeneunipro/ugene/pull/875/files#diff-0ac7ab4346eced218a224dbce7a4950cc03261784fd3d3b32d2732345d42b82dR274) as here too.

DmitriiSukhomlinov (changes requested):
See comments

src/corelibs/U2Gui/src/Notification.cpp
yalgaer: Yes, this is correct behaviour

src/corelibs/U2Gui/src/Notification.cpp
yalgaer: This method was moved from Notification to NotificationStack as is. There is no my changes in this code.

src/corelibs/U2Gui/src/Notification.cpp
yalgaer: I prefer atomic delete here with no in-the-middle re-layout of the visible UI stack on every delete.

The other line you mention is used to delete embedded notifications that have a trickier code and in that case almost never trigger re-layout or the floating stack - so there is a reason for a code-reuse.

PR #887, opened 2022-07-18
UGENE-7621. [Gui, Save dialogs, Windows] Qt error in log when opening file save dialogs

evelinabis:
[UGENE-7621](https://ugene.dev/tracker/browse/UGENE-7621)


src/corelibs/U2Gui/src/util/SaveDocumentController.cpp
```diff
@@ -309,7 +309,9 @@ void SaveDocumentController::addFormatExtension(QString& path) const {
 
 QString SaveDocumentController::prepareDefaultFileFilter() const {
     bool useGzipVariant = conf.compressCheckbox && conf.compressCheckbox->isEnabled();
-    return FileFilters::createFileFilter(currentFormatName, formatsInfo.getExtensionsByName(currentFormatName), useGzipVariant);
+    return FileFilters::createSingleFileFilter(currentFormatName,
```
yalgaer, line 312: createSingleFileFilter will not create `All files(*.*)` filter right? 

In this case this solution is wrong: we should have `All files (*.*)` filter in every file open/save dialog, because formats in UGENE are not extension based but content (header) based.

Limited per extension filters should be considered as helpers, but not restrictions.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2787,6 +2788,32 @@ GUI_TEST_CLASS_DEFINITION(test_7611) {
...
+            GTLogTracer logTracer("QWindowsNativeFileDialogBase::selectNameFilter: Invalid parameter");
+
+            GTWidget::click(os, GTWidget::findToolButton(os, "browseButton"));
+
+            GTUtilsLog::checkContainsMessage(os, logTracer, false);
+            GTWidget::close(os, GTWidget::getActiveModalWidget(os));
+            GTUtilsDialog::clickButtonBox(os, QDialogButtonBox::Cancel);
+        }
+    };
+    qputenv(ENV_USE_NATIVE_DIALOGS, "1");
```
yalgaer: You can't use it in UI tests: different WM/DE may use file dialogs incompatible with our testing system. Even if the test works on Linux the check expression with "QWindowsNativeFileDialogBase" is not meaningful to check at all.

I think we can skip UI test for this bug because tests do not use native file dialogs.

yalgaer (changes requested):
Current solution is not correct.

I think there is a high chance that the  problem is with filters separator characters on Windows.
The final solution should include "All files" filter (like all other file open/save dialogs in UGENE).

evelinabis:
> I think there is a high chance that the problem is with filters separator characters on Windows. The final solution should include "All files" filter (like all other file open/save dialogs in UGENE).

I didn't remove the "All files" filter.

src/corelibs/U2Gui/src/util/SaveDocumentController.cpp
evelinabis: > createSingleFileFilter will not create `All files(*.*)` filter right?

Yes.

> In this case this solution is wrong: we should have All files (*.*) filter in every file open/save dialog, because formats in UGENE are not extension based but content (header) based.

I didn't touch it. I corrected the default filter setting. It should be one.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: I can skip, but note that I did the test only for Windows. Please check that I did it right.

src/corelibs/U2Gui/src/util/SaveDocumentController.cpp
yalgaer: Right. Resolving

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: I would suggest removing this test completely.
If you want to keep it you can register it for Windows only.

evelinabis: I agree with the suggestion, the test has been removed.

yalgaer (approved):

PR #889, opened 2022-07-19
UGENE-5371. Problem with opening assembly from folder with non-ASCII symbols in path

yalgaer:
The solution is to always pass a file handle (int) instead of file name (const char*) to samtools while working with BAM/SAM files

### DIFF
--- a/src/corelibs/U2Formats/src/BAMUtils.cpp
+++ b/src/corelibs/U2Formats/src/BAMUtils.cpp
@@ -74,7 +74,11 @@ static wchar_t* toWideCharsArray(const QString& text) {
 FILE* BAMUtils::openFile(const QString& fileUrl, const QString& mode) {
 #ifdef Q_OS_WIN
     QScopedPointer<wchar_t> unicodeFileName(toWideCharsArray(fileUrl));
-    QScopedPointer<wchar_t> unicodeMode(toWideCharsArray(mode));
+    QString modeWithBinaryFlag = mode;
+    if (!modeWithBinaryFlag.contains("b")) {
+        modeWithBinaryFlag += "b";  // Always open file in binary mode, so any kind of sam, sam.gz, bam, bai files are processed the same way.
+    }
+    QScopedPointer<wchar_t> unicodeMode(toWideCharsArray(modeWithBinaryFlag));
     return _wfopen(unicodeFileName.data(), unicodeMode.data());
 #else
     return fopen(fileUrl.toLocal8Bit(), mode.toLatin1());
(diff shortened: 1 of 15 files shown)


src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -463,28 +522,54 @@ bool BAMUtils::hasValidFastaIndex(const GUrl& fastaUrl) {
...
+ * Builds and saves index for BAM file. Returns 0 if the index was created correctly.
+ * Exact copy of 'bam_index_build2' with a correct unicode file names support.
+ */
+static int bam_index_build_unicode(const QString& bamFileName) {
+    FILE* bFile = BAMUtils::openFile(bamFileName, "rb");
+    CHECK(bFile != nullptr, -1);
+    bamFile fp = bam_dopen(fileno(bFile), "rb");
+    if (fp == nullptr) {
+        closeFileIfOpen(bFile);
+        fprintf(stderr, "[bam_index_build2] fail to open the BAM file.\n");
```
DmitriiSukhomlinov, line 535: Why don't you use the UGENE standard logging?

src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -463,28 +522,54 @@ bool BAMUtils::hasValidFastaIndex(const GUrl& fastaUrl) {
...
+        return -1;
+    }
+    FILE* fpidx = BAMUtils::openFile(bamFileName + ".bai", "wb");
+    if (fpidx == nullptr) {
+        fprintf(stderr, "[bam_index_build2] fail to create the index file.\n");
+        return -1;
+    }
+    bam_index_save(idx, fpidx);
+    bam_index_destroy(idx);
+    fclose(fpidx);
```
DmitriiSukhomlinov, line 552: Has a return value, which could be an error, need to check it

DmitriiSukhomlinov (changes requested):
Minor comments

src/corelibs/U2Formats/src/BAMUtils.cpp
yalgaer: This method is a copy-paste of the original samtools method, so I preserved as many original lines as possible (our samtools logs to stderr)

src/corelibs/U2Formats/src/BAMUtils.cpp
yalgaer: Same here: this is the original behavior of samtools.

PR #899, opened 2022-07-25
UGENE-7578 Print button does not work in the tree

rasputinkirill:


src/corelibs/U2View/src/ov_phyltree/TreeViewer.cpp
```diff
@@ -209,6 +210,8 @@ void TreeViewer::createActions() {
 
     // Print Tree
     printAction = new QAction(QIcon(":/core/images/printer.png"), tr("Print Tree..."), ui);
+    printAction->setDisabled(QPrinterInfo::availablePrinters().isEmpty());
```
yalgaer: Usually when there are no active printers I still can print to PDF file. How this case is handled?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2563,6 +2563,15 @@ GUI_TEST_CLASS_DEFINITION(test_7573) {
     GTUtilsTaskTreeView::waitTaskFinished(os);
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7578) {
```
yalgaer: This test is not needed because it depends a lot on the environment

src/corelibs/U2View/src/ov_phyltree/TreeViewer.cpp
rasputinkirill: If there are no printers installed (even virtual pdf printer) qt will not show "Print" dialog.
I've checked it on my PC without printers installed.

yalgaer: Here how it works on Linux when no printers installed: there is still "Save to PDF" option.

image

yalgaer (commented):
Here is text from QT docs:
`It is NOT recommended to use this as creating each printer instance may take a long time, especially if there are remote networked printers, and retained instances may become outdated if changes are made on the local system or remote print server. Use [availablePrinterNames](https://doc.qt.io/qt-6/qprinterinfo.html#availablePrinterNames)() instead and only instantiate printer instances as you need them.`

What is the result of avalablePrinterNames when you have not printers installed? Does it include a virtual "Save to PDF" as a name?

rasputinkirill:
QPrinterInfo::availablePrinterNames() in my case list still empty.
I don't have an option "Save to PDF" on my windows and linux mint virtual machine.

rasputinkirill:
I've checked that again, both functions not recognize "Save to PDF" virtual printer on linux (i suppose the same with Mac).
So i added OS check for my fix.

src/corelibs/U2View/src/ov_phyltree/TreeViewer.cpp
```diff
@@ -209,6 +210,10 @@ void TreeViewer::createActions() {
 
     // Print Tree
     printAction = new QAction(QIcon(":/core/images/printer.png"), tr("Print Tree..."), ui);
+#ifdef Q_OS_WIN
+    printAction->setDisabled(QPrinterInfo::availablePrinterNames().isEmpty());
```
yalgaer: Printers may appear after the MSA editor is open.
You should make the check after the button is clicked and show "No printers"-like message box if there are no printers.

yalgaer (changes requested):

src/corelibs/U2View/src/ov_phyltree/TreeViewer.cpp
```diff
@@ -1407,6 +1409,12 @@ void TreeViewerUI::sl_showDistanceLabelsTriggered(bool on) {
 }
 
 void TreeViewerUI::sl_printTriggered() {
+#ifdef Q_OS_WIN
```
yalgaer: Use isOsWindows() instead of the macro. It has 2 benefits:
1. Checks on compile time all branches 
2. Checks that isOsWindows function is correctly included: macro may be not included at all.

yalgaer (changes requested):
What will happen if you copy qt windows print support plugin dll into the UGENE distribution?

rasputinkirill:
Qt5PrintSupport.dll is already in distribution and described bug still present.

yalgaer:
> Qt5PrintSupport.dll is already in distribution and described bug still present.

I mean not the core-lib, but `plugins/windowsprintersupport.dll`

rasputinkirill:
Same effect, printer selection dialog does not appear.

yalgaer (changes requested):
Here how I tested
1. Download UGENE v43
2. Build a tree and click "Print" -> no dialog appears
3. Copy `printsupport/windowsprintersupport.dll` into the UGENE dir
4. Build a tree and click "Print" -> the dialog appears. It has no printers and "Microsoft Print to PDF" option

The correct solution is to include this library into the build. Please fix the build and release scripts/checksums.

yalgaer (changes requested):
This change will break release build.
Please update release-bundle.txt

yalgaer (approved):

PR #910, opened 2022-08-01

rasputinkirill:
QT error in log after cancelling PHY tree creation.


rasputinkirill:
Info message updated.

yalgaer (approved):

PR #913, opened 2022-08-02
UGENE-7649. [Mac] The version in Info.plist is outdated

### LINKED ISSUE #904: CFBundleShortVersionString and CFBundleVersion of macOS Application not up to date
The Version numbering `CFBundleShortVersionString` and `CFBundleVersion` in the file `Unipro UGENE.app/Contents/Info.plist` inside the DMG of the MacOS releases do not match up with the version numbers of the releases. They remained unchanged since version 40.0. Fixing that would make it easier to compare different versions.

evelinabis:
Info for the reviewers.
1. I didn't do
    - XML test, because as far as I understand our current XML testing system doesn't consider OS,
    - GUI test because we don't currently test the Mac.

    Is a test required for this issue?
2. [Release](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/ReleaseMacIntelDmg/244181) build doesn't break.

https://ugene.dev/tracker/browse/UGENE-7649

### DIFF
--- a/src/ugeneui/ugeneui.pri
+++ b/src/ugeneui/ugeneui.pri
@@ -64,9 +64,13 @@ macx {
 
     # Configure Info.plist
     PLIST_DIR = _tmp/plist
+    UGENE_VERSION_FOR_PLIST=$${UGENE_VER_MAJOR}.$${UGENE_VER_MINOR}
     infoPlistTarget.target = $${PLIST_DIR}/Info.plist
     infoPlistTarget.depends = FORCE
-    infoPlistTarget.commands = mkdir -p $${PLIST_DIR}; sed \"s/\\\$${UGENE_VERSION}/$${UGENE_VER_MAJOR}.$${UGENE_VER_MINOR}/g\" $$absolute_path(../../etc/script/mac/dmg/Info.plist) > $${infoPlistTarget.target}
+    infoPlistTarget.commands = mkdir -p $${PLIST_DIR}; \
+                               sed -e \"s/\\\$${UGENE_VERSION_FULL}/$${UGENE_VERSION_FOR_PLIST}.0/g\" \
+                                   -e \"s/\\\$${UGENE_VERSION_SHORT}/$${UGENE_VERSION_FOR_PLIST}/g\" \
+                                   $$absolute_path(../../etc/script/mac/dmg/Info.plist) > $${infoPlistTarget.target}
     PRE_TARGETDEPS += $${infoPlistTarget.target}
     QMAKE_EXTRA_TARGETS += infoPlistTarget
     QMAKE_INFO_PLIST = $$OUT_PWD/$${infoPlistTarget.target}
--- a/src/ugeneui/ugeneui.pri
+++ b/src/ugeneui/ugeneui.pri
@@ -61,6 +61,17 @@ win32 {
 
 macx {
     RC_FILE = images/ugeneui.icns
+
+    # Configure Info.plist
+    PLIST_DIR = _tmp/plist
+    infoPlistTarget.target = $${PLIST_DIR}/Info.plist
+    infoPlistTarget.depends = FORCE
+    infoPlistTarget.commands = mkdir -p $${PLIST_DIR}; sed \"s/\\\$${UGENE_VERSION}/$${UGENE_VER_MAJOR}.$${UGENE_VER_MINOR}/g\" $$absolute_path(../../etc/script/mac/dmg/Info.plist) > $${infoPlistTarget.target}
+    PRE_TARGETDEPS += $${infoPlistTarget.target}
+    QMAKE_EXTRA_TARGETS += infoPlistTarget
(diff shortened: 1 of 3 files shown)


yalgaer (changes requested):
Please update the script that uses/generates this file to put the current version automatically

yalgaer (changes requested):
I checked MacOS build #3395 and see the following content in the Info.plist file included into the bundle:
```
    <key>CFBundleVersion</key>
    <string>$(QMAKE_FULL_VERSION)</string>

    <key>CFBundleShortVersionString</key>
    <string>$(QMAKE_SHORT_VERSION)</string>
```
Looks like the tokens were not updated.

evelinabis:
> I checked MacOS build #3395 and see the following content in the Info.plist file included into the bundle:
> 
> ```
>     <key>CFBundleVersion</key>
>     <string>$(QMAKE_FULL_VERSION)</string>
> 
>     <key>CFBundleShortVersionString</key>
>     <string>$(QMAKE_SHORT_VERSION)</string>
> ```
> 
> Looks like the tokens were not updated.

Yes. On a pure template Qt project using our Info.plist and
```
VERSION=43.0.0.0
QMAKE_INFO_PLIST = Info.plist
```
in a .pro file works great. I'm trying to find the reasons for our behavior.
I'm drafting PR, I'll call you when it's ready.

src/ugeneui/ugeneui.pri
```diff
@@ -61,7 +61,17 @@ win32 {
...
+
+    # Configure Info.plist
+    PLIST_DIR = _tmp/plist
+    infoPlistTarget.target = $${PLIST_DIR}/Info.plist
+    infoPlistTarget.depends = FORCE
+    infoPlistTarget.commands = mkdir -p $${PLIST_DIR}; sed \"s/\\\$${UGENE_VERSION}/$${UGENE_VER_MAJOR}.$${UGENE_VER_MINOR}/g\" $$absolute_path(../../etc/script/mac/dmg/Info.plist) > $${infoPlistTarget.target}
+    PRE_TARGETDEPS += $${infoPlistTarget.target}
+    QMAKE_EXTRA_TARGETS += infoPlistTarget
+    QMAKE_INFO_PLIST = $$OUT_PWD/$${infoPlistTarget.target}
+    QMAKE_CLEAN += $${PLIST_DIR}
```
evelinabis: These changes are needed to maintain versioning of the ugeneui executable. I didn't change the behavior that existed before my changes

yalgaer (changes requested):
I'm ready to approve the patch if you remove unrelated changes from ugeneui.pri

If you believe these changes are needed for something else please create another PR with a description why we need them.

evelinabis:
> unrelated changes

Why do you write that these are unrelated changes?



> If you believe these changes are needed for something else

It's false.

Please select your prefered option:
1. Use this script. With it ugeneui version will look like UGENE_VERSION from ugene_version.pri file.  This was the case in all previous release versions of UGENE.
2. Remove the script, but leave `QMAKE_INFO_PLIST = ../../etc/script/mac/dmg/Info.plist` (as in the master). After that, ugeneui will have "${UGENE_VERSION}" version.
3. Remove `QMAKE_INFO_PLIST = ../../etc/script/mac/dmg/Info.plist`. After that, ugeneui will have no version. Different from previous UGENE releases, but not bad behavior: macOS bundles usually don't version their internal executables.

yalgaer:
> > unrelated changes
> 
> Why do you write that these are unrelated changes?
> 
> > If you believe these changes are needed for something else
> 
> It's false.
> 
> Please select your prefered option:
> 
>     1. Use this script. With it ugeneui version will look like UGENE_VERSION from ugene_version.pri file.  This was the case in all previous release versions of UGENE.
> 
>     2. Remove the script, but leave `QMAKE_INFO_PLIST = ../../etc/script/mac/dmg/Info.plist` (as in the master). After that, ugeneui will have "${UGENE_VERSION}" version.
> 
>     3. Remove `QMAKE_INFO_PLIST = ../../etc/script/mac/dmg/Info.plist`. After that, ugeneui will have no version. Different from previous UGENE releases, but not bad behavior: macOS bundles usually don't version their internal executables.

I see your point. 

If we can remove this line completely from the *.pri file
```QMAKE_INFO_PLIST = .```

and it will not break break the build/release I would prefer using this option: it is the simplest. (solution 3)

Otherwise your current variant with "replace" in pri should work (please use some default qmake build output folder for the _tmp dir)

evelinabis:
> If we can remove this line completely from the *.pri file `QMAKE_INFO_PLIST = .`
> and it will not break break the build/release I would prefer using this option: it is the simplest. (solution 3)

It won't break, everything will be fine. I also prefer this solution.

The final version is ready. I looked at the assembled bundle for both High Sierra and Big Sur, everything suited me.

yalgaer (approved):

PR #916, opened 2022-08-04

rasputinkirill:


yalgaer (commented):
Please run GUI tests and I will approve

yalgaer (approved):

PR #931, opened 2022-08-12
UGENE-7644 [Sequence view] Crash after deleting sequence

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7644


src/corelibs/U2Gui/src/util/GObjectComboBoxController.cpp
```diff
@@ -79,7 +79,10 @@ void GObjectComboBoxController::addDocumentObjects(Document* d) {
     // checks whether you need to add a new annotations table
     QString docUrl = settings.relationFilter.ref.docUrl;
     if (d->getURLString() == docUrl) {
-        connect(d->getObjectById(settings.relationFilter.ref.entityRef.entityId), SIGNAL(si_lockedStateChanged()), SLOT(sl_lockedStateChanged()));
+        GObject* seqObj = d->getObjectById(settings.relationFilter.ref.entityRef.entityId);
+        CHECK(seqObj != nullptr, );
+
+        connect(seqObj, SIGNAL(si_lockedStateChanged()), SLOT(sl_lockedStateChanged()));
```
yalgaer: All affected old-style `connect`s should be replaced with function pointer variants.

yalgaer (commented):
Minor issues only. Will approve after the tests

src/corelibs/U2Gui/src/util/GObjectComboBoxController.cpp
DmitriiSukhomlinov: OK

smlnkcv (dismissed):

DmitriiSukhomlinov:
@yalgaer review

src/corelibs/U2View/src/ov_sequence/ADVSingleSequenceWidget.cpp
```diff
@@ -628,6 +628,9 @@ void ADVSingleSequenceWidget::sl_onSelectRange() {
             getSequenceContext()->getSequenceSelection()->setSelectedRegions(curRegions);
         }
     }
+    CHECK(fw != nullptr, );
+
+    fw->setFocus();
```
yalgaer: The originally focused widget may be already destroyed: you should use QPointer.

src/corelibs/U2Gui/src/util/GObjectComboBoxController.cpp
```diff
@@ -79,7 +79,10 @@ void GObjectComboBoxController::addDocumentObjects(Document* d) {
     // checks whether you need to add a new annotations table
     QString docUrl = settings.relationFilter.ref.docUrl;
     if (d->getURLString() == docUrl) {
-        connect(d->getObjectById(settings.relationFilter.ref.entityRef.entityId), SIGNAL(si_lockedStateChanged()), SLOT(sl_lockedStateChanged()));
+        GObject* seqObj = d->getObjectById(settings.relationFilter.ref.entityRef.entityId);
+        CHECK(seqObj != nullptr, );
```
yalgaer: If you return here you disable a bunch of code below that does not depend on presence of the sequence object.
Examples:
```QList<GObject*> listAnnotations = d->findGObjectByType(GObjectTypes::ANNOTATION_TABLE)```
``` foreach (GObject* obj, d->getObjects()) {```

src/corelibs/U2View/src/ov_sequence/ADVSingleSequenceWidget.cpp
yalgaer: A better solution would be not changing the code here at all: the focus must be returned to the original widget automatically and if it is not returned: this is a bug.

In your test you can close "Pan View" first, so the focus will remain in DetView.

yalgaer (changes requested):
See the comments: the current fix changes original behavior + uses unsafe workarounds for the focus handling

src/corelibs/U2View/src/ov_sequence/ADVSingleSequenceWidget.cpp
DmitriiSukhomlinov: Unfortunately, closing PanView won't help. For some reason, on Linux, when you close the "Select sequence region" dialog focus is being lost from the widget it was on before. This problem doesn't appear on Windows and macOS. 
I'm not sure why is this happening, but, as far as focusing has the same code on each OS I may suppose, that this problem relates to Qt somehow. 
Please, check it. I don't like this solution either and would be glad if you suggest another one.

DmitriiSukhomlinov: Just a scenario to reproduce, to be sure that we understand each other:

1. Open any sequence.
2. Enable details view if it's not.
3. Enable editing mode.
4. Click on DetView.
5. Click "Ctrl + A" to open the "Select sequence region" dialog.
6. Select any region or you may even just click "OK"
7. Click "A" to insert a new base to the place where the edit pointer is located at.
On Windows and macOS: focus is still on DetVIew, the "A" base has been inserted
On Linux: focus got lost, the "A" base hasn't been inserted

src/corelibs/U2Gui/src/util/GObjectComboBoxController.cpp
DmitriiSukhomlinov: Yes, there are some code below this line, but just check it carefully - if `seqObj` here is null, then nothing will happens. 
In the mentioned `foreach` we check some relationships, if we have some we just set some local variable `hasAnnotationTable` to `true`. Nothing meaningful happens. 
The all other code depends on the `seqObj` which is `nullptr`. 
I could rewrite it, but I there's no point.

src/corelibs/U2View/src/ov_sequence/ADVSingleSequenceWidget.cpp
yalgaer: I checked on Linux and if I 

a) close PanView 
and 
b) click on DetView 

between steps 2 and 3 the focus after step 6 will remain on the DetView

src/corelibs/U2Gui/src/util/GObjectComboBoxController.cpp
yalgaer: The correct rewrite of this method should remove the top-level 'if' block code completely (which is an old workaround) and use a separate explicit API to add custom options into the combo-box. 
In the case of sequence view it will be a ''Create-New-Feature-Table-In-The-Sequence-Object' option. This option should be added only if the format supports writable features.

This is a much bigger task than the problem in this issue and it will affect (require a review) of all places this combo controller is used.

To fix only this issue I'm asking not changing behavior of this method: because the old and the new code are not equal and review of all side effects is out of scope of this PR.

Why the behaviors are not equal:
Sequence object may be null, but there may still be annotations in the matched document (GenBank with no sequences but with annotations). In this case the nested `if (!hasAnnotationTable)`check  won't work: the execution will proceed to the loop at the end of method which contains a good code.

yalgaer (commented):
See comments. I think we should keep this fix as simple as possible.

src/corelibs/U2View/src/ov_sequence/ADVSingleSequenceWidget.cpp
DmitriiSukhomlinov: Well, OK, but the problem is still with us, what shell we do? Just skip it?

yalgaer: Lets create a bug with this issue description and fix focus management in sequence view. It may be a small or a big fix. This requires investigation and a solution that may be very different from the current one.

yalgaer (dismissed):

src/corelibs/U2View/src/ov_sequence/ADVSingleSequenceWidget.cpp
DmitriiSukhomlinov: https://ugene.dev/tracker/browse/UGENE-7673

yalgaer (approved):

PR #932, opened 2022-08-14
Fix object-by-name sort in buildMoveSelectionToAnotherObjectMenu

yalgaer:
I found this fix while reviewing Igor's PR.
I do not know how to reproduce the problem from UI, but incorrect compactors in `std::sort` may cause crashes in UGENE.

This PR includes the fix + I reviewed all other usages of std::sort/stable_sort in UGENE.


DmitriiSukhomlinov (changes requested):
I don't see, how it works, `GObject::objectLessThan` has the same code. Of course, it makes  code a bit shorter and prettier, but, I don't see how it could help to fix any crash.

yalgaer:
> I don't see, how it works, `GObject::objectLessThan` has the same code. Of course, it makes code a bit shorter and prettier, but, I don't see how it could help to fix any crash.

Why do you think that the code in `objectLessThan` is equal to the original one? It is not.

PR #935, opened 2022-08-24
UGENE-7668. Support sync mode for phy-trees in Circular and Unrooted layouts

yalgaer:


src/corelibs/U2View/src/ov_phyltree/GraphicsRectangularBranchItem.cpp
```diff
@@ -57,12 +57,12 @@ void GraphicsRectangularBranchItem::collapse() {
             }
 
             GraphicsRectangularBranchItem* childItem = dynamic_cast<GraphicsRectangularBranchItem*>(graphItem);
-            if (nullptr == childItem) {
+            if (childItem == nullptr) {
```
DmitriiSukhomlinov: `CHECK_CONTINUE`?

DmitriiSukhomlinov (dismissed):
Minor

src/corelibs/U2View/src/ov_phyltree/GraphicsRectangularBranchItem.cpp
yalgaer: Fixed

PR #936, opened 2022-08-26
[UGENE-7630] Replace recursive calls by loop

ichebyki:
See [UGENE-7630](https://ugene.dev/tracker/browse/UGENE-7630)


yalgaer (commented):
The idea to replace recursion with a loop is great.

Two comments/questions:

1) We need a couple of tests to check the fix: 
  a) To check that we have no regressions: 2-3 small API tests with all data in test in C++ code
  b) To check that the problem from the bug is fixed: it can be checked with an XML test that calls  find algorithm with the data from the bug (you can extract/copy-paste the sequences into separate files for that).

2) This recursion never goes beyond max(x,y) which is ~5k in the scenario described in the bug, which can't cause crashes with  native code. Do I understand correctly that the problem is only reproducible on Mac emulator (with large stack frames)?

ichebyki:
сб, 27 авг. 2022 г., 08:00 Yuliya Algaer ***@***.***>:

yalgaer:
> сб, 27 авг. 2022 г., 08:00 Yuliya Algaer ***@***.***>:

'nothing' here means that stack trace = 5k for a trivial method & compiled code can't cause such errors. But such error can happen of it is some kind of 'emulator' or interpreted that can't handle large recursion.

ichebyki:
Don't understand what "emulator" you speak about. I can reproduce the crash
with last ugene version debug build on our Mac laptop. MacOS is Monterey.
Anyway one function call is equal to about ~10 integer "add" instructions
processor's ticks. It's not good for my strong opinion.

вс, 28 авг. 2022 г., 06:09 Yuliya Algaer ***@***.***>:

yalgaer:
> Don't understand what "emulator" you speak about. 

Our Mac is arm64 and UGENE build is x84-64?  Please read: https://en.wikipedia.org/wiki/Rosetta_(software)
This error is only possible because of it.

>Anyway one function call is equal to about ~10 integer "add" instructions processor's ticks. It's not good for my strong opinion. 
How many extra instructions you need to avoid the recursion?

Overall I think the patch is great and is ready to approve it. But it needs tests.

ichebyki:
пн, 29 авг. 2022 г., 00:40 Yuliya Algaer ***@***.***>:

> Don't understand what "emulator" you speak about.
>
> Our Mac is arm64 and UGENE build is x84-64? Please read:
> https://en.wikipedia.org/wiki/Rosetta_(software)
> This error is only possible because of it.
>
I know about this
The error is possible for any user which starts ugene on same laptop.

ichebyki:
Just FYI
This is not "emulator", this is "binary compiler".
The same feature exists in Elbrus processor - it can execute x86 code on the fly

yalgaer:
> Just FYI This is not "emulator", this is "binary compiler". 

Thank you. The name does not matter here, since we do not know all internal details of this product. 

What really matters here that with a native binaries produced by gcc or MSVC 'binary compilers' recursion level of 1-million calls for small functions like this is OK, while for 'rosetta binary compiler' on MacM1 even 1k may be too much.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2803,6 +2808,72 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
     GTUtilsLog::checkContainsError(os, logTracer, "All input reads contain gaps or Ns only, abort");
 }
 
+GUI_TEST_CLASS_DEFINITION(test_7630) {
+    // Open CVU55762.gb and murine.gb in separate sequence mode.
+    GTFileDialog::openFile(os, dataDir + "/samples/Genbank/", "CVU55762.gb");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    // In CVU55762 select region 1001-1000.
+    SelectSequenceRegionDialogFiller* filler = new SelectSequenceRegionDialogFiller(os, 1001, 1000);
```
yalgaer, line 2857: nit: do not use type name twice in the same line. Use 'auto'

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2803,6 +2808,72 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+    filler = new SelectSequenceRegionDialogFiller(os, 1000, 5830);
+    filler->setCircular(true);
+    GTUtilsDialog::waitForDialog(os, filler);
+    GTKeyboardDriver::keyClick('a', Qt::ControlModifier);
+
+    // Right click on the sequence->Replace subsequence... "Replace sequence" dialog appears.
+    // Paste clipboard into text field (Cmd-V).
+    // Press Enter.
+    // Dialog closed, sequence changed. Now only 1 annotation remains (5' terminal repeat).
+    class ReplaceSequenceScenario : public CustomScenario {
```
yalgaer, line 2880: Do we need such a complex test? 

Can the recursion issue be reproduced with a simpler data like 5-k of 'A' characters searched in the sequence that contains only 'A' characters?
In this case you can create a sequence from data (generate a string of 5-k 'A') and next search the same generated files: no 'select-region'/'replace-sequence'/'click-annotation' actions are needed at all.

yalgaer:
We also need several API level tests when the changed function is called directly with input from C++ with a small human checkable input: empty table, table of size 1, 2, 3...

Check `tests/unit_tests/util/MsaUtil/MsaUtils_compareRowsAfterAlignment.xml` as an example of API tests

ichebyki:
> We also need several API level tests when the changed function is called directly with input from C++ with a small human checkable input: empty table, table of size 1, 2, 3...
> 
> Check `tests/unit_tests/util/MsaUtil/MsaUtils_compareRowsAfterAlignment.xml` as an example of API tests

The are ~9377 calls of the getLen() while xml testing
The sort -u shows:
getLen(1, 1
getLen(2, 1
getLen(2, 2
getLen(3, 3
getLen(4, 3
getLen(4, 4
getLen(5, 4
getLen(5, 5
getLen(6, 5
getLen(6, 6
getLen(7, 5
getLen(7, 7
getLen(8, 7
getLen(8, 8
getLen(10, 10
getLen(20, 20
getLen(28, 28

yalgaer:
> > We also need several API level tests when the changed function is called directly with input from C++ with a small human checkable input: empty table, table of size 1, 2, 3...
> > Check `tests/unit_tests/util/MsaUtil/MsaUtils_compareRowsAfterAlignment.xml` as an example of API tests
> 
> The are ~9377 calls of the getLen() while xml testing The sort -u shows: getLen(1, 1 getLen(2, 1 getLen(2, 2 getLen(3, 3 getLen(4, 3 getLen(4, 4 getLen(5, 4 getLen(5, 5 getLen(6, 5 getLen(6, 6 getLen(7, 5 getLen(7, 7 getLen(8, 7 getLen(8, 8 getLen(10, 10 getLen(20, 20 getLen(28, 28

This is right, the method is very important and used in a lot of places and we have 0 direct tests for it.

Are you sure that that calls you see in the log validate the results related to this method call, but not only check that 'some bigger task finished with no errors'? are 10k calls all independent? It can be a single call with 10k matrix inside. How can you sure that boundary checks work with no errors?

The fix is good and is important and will be included into the upcoming release. Writing tests will take about 1 hour. If you do not want to add the tests by yourself please re-assign the bug to me and I will add the tests by myself and keep your implementation.

ichebyki:
I will create the test
But your question is a very strange: if any of the 9,377 calls is not verified then what is the quality of the our xml-tests ?

yalgaer:
> I will create the test But your question is a very strange: if any of the 9,377 calls is not verified then what is the quality of the our xml-tests ?

Every XML test verify a limited scope: for example a workflow task may run a lot of code, but the test will verify only some parameters of 1 actor result. The rest of the side effects may be completely ignored: out of scope of the test. Some tests are even verify that "UGENE does not crash". These are valid checks based on historical issues, but not related to the current change.

src/plugins/api_tests/src/core/util/MsaUtilsUnitTests.cpp
```diff
@@ -97,4 +101,23 @@ IMPLEMENT_TEST(MsaUtilsUnitTests, all_names_with_spaces) {
...
+    int newLen = dynTable.getLastLen();
+    CHECK_NO_ERROR(os);
+    CHECK_EQUAL(4733, newLen, "getLastLen() return value");
+
+    DynTable dynTable2(3733, 4733, false);
+    newLen = dynTable2.getLastLen();
+    CHECK_NO_ERROR(os);
+    CHECK_EQUAL(3733, newLen, "getLastLen() return value");
+
+    DynTable dynTable3(44733, 44733, false);
```
yalgaer: How this check is different from the first one?

Can you please add more tests for a small table with custom paths inside of the table. 
All 3 tests you added check only degenerated case of empty table, right?

yalgaer (changes requested):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
ichebyki: I think yes

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
ichebyki: changed

src/plugins/api_tests/src/core/util/MsaUtilsUnitTests.cpp
ichebyki: The test checks the long loops, which were failed on the Mac
All other real search are in xml tests.
There are more then nine thousands calls of the method
Do you really need to repeat the same calls (maybe with different args) ?

yalgaer: Right, it is enough a single test to check the long loop problem on Mac OS and you already have a regression test for that. You can add 1 more XML - this is OK too.

But you have rewritten the whole method: the method depends on the internal matrix content: it computes the result length using content of the matrix.

The tests you added do not deal with it: you always fill the matrix with 0 values. Could you please add tests with some meaningful and non-degenerated content?

ichebyki:
Broke the dispute about the tests.
Added Yuliya to Assignees

yalgaer:
> Broke the dispute about the tests. Added Yuliya to Assignees

Ok, I will add the tests until EOW and will ping you for the review.

yalgaer:
I added the tests. Please check

PR #940, opened 2022-08-30
UGENE-7549 Crash when opening a file while closing UGENE

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7549

If some dialog (as the dialog in the scenario, called in ProjectLoaderImpl.cpp, line 231) has Main Window as parent and main window closes, then some difficulties appears into Qt core and everything crashes. Actually, each `static` function like `QFileDialog::getOpenFileName`, which was used here, has a special note for this case in documentation:
image

I replaced the static call with the call using the `QFileDialog` constructor and added a connection, which sets parent as `nullptr` if this parent is about to be destroyed.

By the way, I'm not sure about GUI test. It works, but I'm not sure, that it's OK to close the Application from this test.


src/corelibs/U2Gui/src/util/U2FileDialog.cpp
```diff
@@ -94,11 +108,68 @@ QString U2FileDialog::getSaveFileName(QWidget* parent,
...
+                                     QFileDialog::AcceptSave,
+                                     QFileDialog::ExistingFile);
 }
 
+QStringList U2FileDialog::getFileNames(QWidget* parent,
+                                       const QString& caption,
+                                       const QString& dir,
+                                       const QString& filter,
+                                       const QString& selectedFilter,
+                                       QFileDialog::Options options,
```
yalgaer: Use `const Type&` for all parameters of non primitive types (these are 3rd party type and a reader is not aware if they are classes  or enums).

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
```diff
@@ -94,11 +108,68 @@ QString U2FileDialog::getSaveFileName(QWidget* parent,
...
+}
+
+QString U2FileDialog::getFileName(QWidget* parent,
+                                  const QString& caption,
+                                  const QString& dir,
+                                  const QString& filter,
+                                  const QString& selectedFilter,
+                                  QFileDialog::Options options,
+                                  QFileDialog::AcceptMode acceptMode,
+                                  QFileDialog::FileMode fileMode) {
```
yalgaer: Same here: use `const Type&`

src/corelibs/U2Gui/src/util/U2FileDialog.h
```diff
@@ -28,24 +28,20 @@
...
- * This class should be removed after switching to Qt 5.1 or newer
- */
 class U2GUI_EXPORT U2FileDialog {
 public:
     static QString getOpenFileName(QWidget* parent = 0,
                                    const QString& caption = QString(),
                                    const QString& dir = QString(),
                                    const QString& filter = QString(),
-                                   QString* selectedFilter = 0,
+                                   const QString& selectedFilter = QString(),
```
yalgaer, line 37: nit: use "" instead of QString()

src/corelibs/U2Gui/src/util/U2FileDialog.h
```diff
@@ -57,8 +53,28 @@ class U2GUI_EXPORT U2FileDialog {
...
                                    const QFileDialog::Options& options = 0);
+
+private:
+    static QStringList getFileNames(QWidget* parent,
```
yalgaer: nit: can be moved to file local static function. Nobody cares about private static interface of the class - so there is no reason to have it the header: it will only increase compilation time on changes to these methods.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2436,6 +2437,25 @@ GUI_TEST_CLASS_DEFINITION(test_7548) {
...
+
+    // Press shutdown button "File -> Exit" in the right corner.
+    // Press Yes in the dialog.
+    // Press Save in "Save project as" dialog.
+    GTUtilsDialog::waitForDialog(os, new SaveProjectAsDialogFiller(os, "New Project", sandBoxDir + "test_7549.proj"));
+    GTUtilsDialog::waitForDialog(os, new SaveProjectDialogFiller(os, QDialogButtonBox::Yes));
+    GTMenu::clickMainMenuItem(os, { "File", "Exit" });
+
+    // After the dialogs disappear and UGENE starts to close, press "File -> Open..."
+    GTMenu::clickMainMenuItem(os, { "File", "Open..." });
```
yalgaer: I think this test is not reliable and depends on the timings (how fast does UGENE close.
I would suggest not having it at all.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
```diff
@@ -63,22 +63,36 @@ QString U2FileDialog::getOpenFileName(QWidget* parent,
...
+                                     effectiveOptions,
+                                     QFileDialog::AcceptOpen,
+                                     QFileDialog::ExistingFile);
 }
 
 QStringList U2FileDialog::getOpenFileNames(QWidget* parent,
                                            const QString& caption,
                                            const QString& dir,
                                            const QString& filter,
-                                           QString* selectedFilter,
```
yalgaer, line 77: nit: this is a good change

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
```diff
@@ -94,11 +108,68 @@ QString U2FileDialog::getSaveFileName(QWidget* parent,
...
+                                       QFileDialog::FileMode fileMode) {
+
+    QFileDialog fileDialog(parent, caption, dir, filter);
+    if (!selectedFilter.isEmpty()) {
+        fileDialog.selectNameFilter(selectedFilter);
+    }
+    fileDialog.setOptions(options);
+    fileDialog.setFileMode(fileMode);
+    fileDialog.setAcceptMode(acceptMode);
+    auto deleteConnection = QObject::connect(parent, &QWidget::destroyed, [&]() {
```
yalgaer: So we want to allow deletion of parent of QFileDialog, right? 
In this case we should follow the following logic: when a parent of a widget is deleted the widget is deleted too.

Here you keep the dialog alive for a longer period, than the parent: the parent is deleted, but the dialog stays alive.

This does not look a correct behavior and has side-effects: close() will cause CloseEvent to be sent to all parties first, next the dialog will process CloseEvent and send signals (rejected/done) to all parties and only next it will be destroyed. This all happens when the application hierarchy is already partially destroyed: destroyed() is emitted from inside of the main window object destructor.  I would follow a simple rule: never listen to 'destroyed' within the same child<->parent hierarchy.

I see 2 simpler solutions to check:

1) We can disable file dialog invocation during UGENE shutdown completely: either by disabling the related actions or/and, by creating a special flag that shutdown is in progress and using this flag inside U2FileDialog factory methods: the dialog won't be shown at all if UGENE is closing (CHECK(!isShuttingDown)).

2) Allow destroy of the dialog with the parent. Use QPointer<> to track if the dialog was destroyed inside of the U2FileDialog methods and return empty result if it was.

yalgaer (changes requested):
Check the comments + a lot of tests failed

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: Passing primitive types (`enum` is, basically, just an usual `int`) by the const reference doesn't make any sense and, even, incorrect, because it causes some complications (this was asked and answered numerous amount of times, like [here](https://stackoverflow.com/a/10770431/7273346), [here](https://stackoverflow.com/a/4112947/7273346), [here](https://stackoverflow.com/a/14013189/7273346) and [here](https://stackoverflow.com/a/6834608/7273346)). I have never done this and see no point to begin.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: Same

src/corelibs/U2Gui/src/util/U2FileDialog.h
DmitriiSukhomlinov: Why?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
DmitriiSukhomlinov: Well, OK, I had some concerns about this test too

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: > these are 3rd party type and a reader is not aware if they are classes or enums

Actually, this is the reason why it should be without `const &` - to aware reader, that this is not a class

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: I don't see a point into a not-trusting the `destroyed` signal, which is called right before the object is deleted and supposed to be used just for the cases like this. This is a Qt very-very base architecture, if we cannot trust it we cannot trust anything in our code at all. That means, that if we have a rule "never listen to 'destroyed'", than we should have a rule "never listen to anything which is produced by Qt".
And we cannot - just really cannot - delete the dialog when `exec()` is still working - it will definitely make a crash. Where is nothing bad in **existing** of the `QFileDialog fileDialog` object, it doesn't creates any problems. The problem appears because it's **executing** when parent dyes, not because it just **exists**.
So, my solution looks like it uses the exact functions which should be used in this exact case (actually, I spied it somewhere on stackoverflow, so this is a pretty well-known solution), I see what you said, but see no point of changing it.

src/corelibs/U2Gui/src/util/U2FileDialog.h
DmitriiSukhomlinov: Ok

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
yalgaer: We need a common style for the whole application: please update to `const T&`, this apporach does not make any assumptions if the reader knows that 'options' is a 'class' or an 'int'

src/corelibs/U2Gui/src/util/U2FileDialog.h
yalgaer: This is 'nit' and you can ignore it if you want.
If you want to know why: "" is better than QString() because it is shorter and gives the same level of info.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
```diff
@@ -33,6 +33,59 @@
 
 namespace U2 {
 
+
+namespace {
```
yalgaer: Use static on the function level, not a static namespace. 
Reason: the reader can't see the scope in 2-3 pages away from the function, so using 'static' is a required style for UGENE.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
yalgaer: What will happen if you allow deletion of the dialog together with the parent (done automatically)? 

If it works (like it works with any other dialog we have in UGENE) we can throw away the 1) connect 2) re-parent 3) manual close 4) disconnect parts in your code. The closing will happen naturally.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
```diff
@@ -33,6 +33,59 @@
...
+
+QStringList getFileNames(QWidget* parent,
+                         const QString& caption,
+                         const QString& dir,
+                         const QString& filter,
+                         const QString& selectedFilter,
+                         QFileDialog::Options options,
+                         QFileDialog::AcceptMode acceptMode,
```
yalgaer, line 45: Do we really need the last 2 options? The list of parameters is already very long and we can safely use defaults for some of them that  we do not change.
We can re-introduce them back when we will have the first use-case. Until that lets not adding any boilerplate with no real need.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: You cannot delete parent while the dialog is executing, this fact is even documented:
image

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
yalgaer: Right, and you already do what is written here: you create the dialog using QFileDialog constructor. 
Now it is safe to have it auto-deleted with parent.

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
DmitriiSukhomlinov: Well, as far as I said - we do not need it, its just some complication you've thought up for I don't know why reason (sorry, but that'' true). The **existing** of the `fileDialog` object doesn't make any problems, its **execution** makes and I handled the execution. The dialog will be destroyed on itself on `getFileNames` closing. It's not a pointer, it's just an object, we do not need to do anything about it, it will be handled on its own when it pops out of the call stack, this is how the C++ language works, I do not know how to explain it anymore.

Let's do this - please, gimme a scenario where we still have a crash of `SAFE_POINT` or whatever and I will fix it.

yalgaer: Here is a thing that is better than an example: a complete code snippet of the solution:

```c++
    QObjectScopedPointer<QFileDialog> fileDialog(new QFileDialog(parent, caption, dir, filter));
    if (!selectedFilter.isEmpty()) {
        fileDialog->selectNameFilter(selectedFilter);
    }
    fileDialog->setOptions(options);
    fileDialog->setFileMode(fileMode);
    fileDialog->setAcceptMode(acceptMode);

    CHECK(fileDialog->exec() == QFileDialog::Accepted && !fileDialog.isNull(), {});
    return fileDialog->selectedFiles();
```

We already use this pattern across UGENE code everywhere to handle dialogs and QT docs for QFileDialog says this approach is legit for file dialogs too.

yalgaer (approved):

src/corelibs/U2Gui/src/util/U2FileDialog.cpp
yalgaer: I see my fix does not work on 5.12.12 on Windows but works on 5.15.2
Lets use your fix in this case.
Please fix other issues (reference types) before submitting.

PR #976, opened 2022-09-21
UGENE-7652 Crash when exporting consensus from AE + adding it to MSA

rasputinkirill:


src/corelibs/U2Gui/src/OpenViewTask.cpp
```diff
@@ -84,13 +84,16 @@ static Task* createOpenViewTask(const MultiGSelection& ms) {
             ls.removeAll(f);
         }
     }
-
+    Task* t = nullptr;
+    if (QApplication::activeModalWidget() != nullptr) {
+        coreLog.info(QObject::tr("Unable to open view because of active modal widget."));
+        return t;
+    }
     if (ls.size() == 1) {
```
yalgaer: nit: all lines below can be a 1 liner: `return ls.size() == 1 ?  ls[0]->createViewTask()  : nullptr;`

src/corelibs/U2Gui/src/OpenViewTask.cpp
```diff
@@ -84,13 +84,16 @@ static Task* createOpenViewTask(const MultiGSelection& ms) {
             ls.removeAll(f);
         }
     }
-
+    Task* t = nullptr;
+    if (QApplication::activeModalWidget() != nullptr) {
+        coreLog.info(QObject::tr("Unable to open view because of active modal widget."));
```
yalgaer, line 90: This is a big limitation and we should allow creation of views in the background.
May be it worth to create a separate bug for that.

yalgaer:
The fix looks reasonable. Please run GUI tests and I will approve.

yalgaer:
Too many tests failed on Linux, please fix

yalgaer (approved):

PR #982, opened 2022-09-25
UGENE-7682. Support 'curvature' option for rectangular tree visualization

yalgaer:


src/corelibs/U2View/src/ov_phyltree/GraphicsRectangularBranchItem.cpp
```diff
@@ -153,11 +153,48 @@ QRectF GraphicsRectangularBranchItem::boundingRect() const {
 
 void GraphicsRectangularBranchItem::paint(QPainter* painter, const QStyleOptionGraphicsItem*, QWidget*) {
     setUpPainter(painter);
+    CHECK(!qFuzzyCompare(width, 0), );
```
DmitriiSukhomlinov: `Note that comparing values where either p1 or p2 is 0.0 will not work, nor does comparing values where one of the values is NaN or infinity. If one of the values is always 0.0, use qFuzzyIsNull instead.` - from the documentation. Use [qFuzzyIsNull](https://doc.qt.io/qt-6/qtglobal.html#qFuzzyIsNull)

DmitriiSukhomlinov (dismissed):
Minor comment

src/corelibs/U2View/src/ov_phyltree/GraphicsRectangularBranchItem.cpp
yalgaer: I think it is safe to remove this check at all (I kept it from the original code)

PR #994, opened 2022-10-11
UGENE-7667. Fix QueryDesigner

evelinabis:
[7667](https://ugene.dev/tracker/browse/UGENE-7667)


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3078,6 +3081,178 @@ GUI_TEST_CLASS_DEFINITION(test_7659) {
...
+    GTUtilsDialog::add(os, new Filler(os, "RunQueryDlg", new RunSchemaDialogScenario(inputPath, addToProject)));
+    GTMenu::clickMainMenuItem(os, {"Actions", "Run Schema..."});
+}
+
+// Sets the cell (`row`, 1) of `table` to `value`.
+static void setDouble(GUITestOpStatus& os, QTableView* table, int row, double value) {
+    QModelIndex modelIndex;
+
+    class FindModelIndexScenario : public CustomScenario {
+        QTableView* table_;
```
yalgaer: Why not using `table` instead of `table_` here? It won't be a name shadowing in this case.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3078,6 +3081,178 @@ GUI_TEST_CLASS_DEFINITION(test_7659) {
     CHECK_SET_ERR(barWidget->tabText(0) == "NewSet", "Actual dataset name on 'Read Sequence' worker is not expected 'NewSet'.");
 }
 
+static void clickOn(const QPoint& point) {
```
yalgaer: Use `GTWidget::click` or create another `GTWidget::moveToAndClick`  with some documentation that explains why the first is not good enough.

The method looks very generic so it is not clear why this test needs it and others don't.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3078,6 +3081,178 @@ GUI_TEST_CLASS_DEFINITION(test_7659) {
...
+            : table_(table_), rowNum(rowNum), modelInd(modelInd) {
+        }
+        void run(GUITestOpStatus&) override {
+            modelInd = table_->model()->index(rowNum, 1);
+        }
+    };
+    GTThread::runInMainThread(os, new FindModelIndexScenario(table, row, modelIndex));
+
+    GTWidget::scrollToIndex(os, table, modelIndex);
+    clickOn(GTTableView::getCellPosition(os, table, 1, row));
```
yalgaer: This should be GTTableView::click(....) method. If may even include `scrollToIndex` as the first step.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3078,6 +3081,178 @@ GUI_TEST_CLASS_DEFINITION(test_7659) {
...
+    GTUtilsAnnotationsTreeView::checkAnnotationRegions(os, "Result 4  (0, 2)", {{452, 471}, {610, 629}});
+    GTUtilsAnnotationsTreeView::checkAnnotationRegions(os, "Result 5  (0, 2)", {{610, 629}, {786, 805}});
+    GTUtilsProject::closeProject(os, true);
+    GTUtilsMdi::activateWindow(os, "Query Designer - NewSchema");
+
+    clickOn(GTUtilsQueryDesigner::getItemCenter(os, "Primer"));
+    auto table = GTWidget::findTableView(os, "table");
+    {  // Product size ranges.
+        clickOn(GTTableView::getCellPosition(os, table, 1, 2));
+        GTLineEdit::setText(os, GTWidget::findWidgetByType<QLineEdit*>(os, table, "7667"), "100-300");
```
yalgaer, line 3204: Use `static void setText(GUITestOpStatus& os, const QString& lineEditName...` so it will be 1 method call.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3078,6 +3081,178 @@ GUI_TEST_CLASS_DEFINITION(test_7659) {
...
+    setDouble(os, table, 6, 10);  // Max 3' stability.
+    setDouble(os, table, 7, 20);  // Pair max repeat mispriming.
+    setDouble(os, table, 8, 20);  // Pair max template mispriming.
+    runSchema(os);
+    GTUtilsAnnotationsTreeView::checkAnnotationRegions(os, "Result 1  (0, 2)", {{199, 218}, {297, 316}});
+    GTUtilsAnnotationsTreeView::checkAnnotationRegions(os, "Result 2  (0, 2)", {{40, 59}, {297, 316}});
+    GTUtilsAnnotationsTreeView::checkAnnotationRegions(os, "Result 3  (0, 2)", {{39, 58}, {297, 316}});
+}
+
+GUI_TEST_CLASS_DEFINITION(test_7667_0) {
```
yalgaer: Naming is non-symmetrical: I would suggest using `test_7667_1` and `test_7667_2`

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
```diff
@@ -65,6 +60,12 @@ QString QDPrimerActor::getText() const {
 
 Task* QDPrimerActor::getAlgorithmTask(const QVector<U2Region>& /*location*/) {
     Task* t = nullptr;
+
+    auto settings = new Primer3TaskSettings;
```
yalgaer: add () to the constructor call (use the same style everywhere in UGENE)

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
```diff
@@ -101,27 +102,27 @@ Task* QDPrimerActor::getAlgorithmTask(const QVector<U2Region>& /*location*/) {
     int numRet = cfg->getParameter(NUM_RETURN_ATTR)->getAttributeValueWithoutScript<int>();
     settings->setIntProperty("PRIMER_NUM_RETURN", numRet);
 
-    /*qreal maxMispriming = cfg->getParameter(MAX_MISPRIMING_ATTR)->getAttributeValueWithoutScript<double>();
-    settings->setAlignProperty("PRIMER_MAX_MISPRIMING", maxMispriming);
-    assert(settings->getAlignPropertyList().contains("PRIMER_MAX_MISPRIMING"));*/
+    qreal maxMispriming = cfg->getParameter(MAX_MISPRIMING_ATTR)->getAttributeValueWithoutScript<double>();
```
yalgaer: nit: Use double, not qreal

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
```diff
@@ -101,27 +102,27 @@ Task* QDPrimerActor::getAlgorithmTask(const QVector<U2Region>& /*location*/) {
...
+    settings->setDoubleProperty("PRIMER_MAX_LIBRARY_MISPRIMING", maxMispriming);
+    assert(settings->getDoublePropertyList().contains("PRIMER_MAX_LIBRARY_MISPRIMING"));
```
yalgaer: Do not use asserts, use 'SAFE_POINTS' so we detect the problem in release tests

src/plugins_3rdparty/primer3/src/Primer3Task.cpp
```diff
@@ -663,12 +663,19 @@ void Primer3Task::selectPairsSpanningIntron(p3retval* primers, int toReturn) {
...
     : Task("Pick primers SW task", TaskFlags_NR_FOSCOE | TaskFlag_CollectChildrenWarnings),
-      settings(_settings) {
+      settings(_settings),
+      ownsSettings(_ownsSettings) {
     median = settings->getSequenceSize() / 2;
 }
 
+Primer3SWTask::~Primer3SWTask() {
+    if (ownsSettings) {
+        delete settings;
```
yalgaer, line 599: Can you just pass the settings by value instead of these complex memory management protocol?

src/plugins_3rdparty/primer3/src/Primer3Task.h
```diff
@@ -196,6 +197,7 @@ class Primer3SWTask : public Task {
     Primer3Task* primer3Task = nullptr;
     int median;
     Primer3TaskSettings* settings;
+    bool ownsSettings;
```
yalgaer: initialize fields with default values.

yalgaer (changes requested):

src/plugins_3rdparty/primer3/src/Primer3Task.h
evelinabis: What is the default value here?

evelinabis: This default value will never be used and it only confuses the code.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: I'll do it, but I'm not sure if it should be the `GTWidget` method.

evelinabis: Should I make changes to `GTWidget` and `GTTableView` within this issue and this branch, or is it better to do another issue and/or another branch?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: The test falls.

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
evelinabis: And I've never used asserts :)

src/plugins_3rdparty/primer3/src/Primer3Task.cpp
evelinabis: I'd like to do this, but don't know how... I can do a workaround like "create a separate task for QD that takes a `QScopedPointer<Primer3TaskSettings>` and has a `Primer3SWTask` as a subtask". Or "try to inherit from `Primer3SWTask` and also have `QScopedPointer<Primer3TaskSettings>` as a field". I haven't tried this, but I hope there will be no pitfalls.

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
evelinabis: This remark applies to the entire UGENE and, in my opinion, should be resolved in a separate commit.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
evelinabis: ✓

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
evelinabis: ✓

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
evelinabis: ✓

src/plugins_3rdparty/primer3/src/Primer3Task.h
evelinabis: ✓

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: I would make it in the GTWidget, because it is very generic.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: I do not understand what will be the difference, but OK.
Just change `findWidgetByType` to `findLineEdit`

src/plugins_3rdparty/primer3/src/Primer3Query.cpp
yalgaer: I usually add comments like these about changed lines: once the line is changed it may be fixed too.
Not a blocker anyway.

src/plugins_3rdparty/primer3/src/Primer3Task.cpp
yalgaer: What will happen if you pass `Primer3TaskSettings` by value? So it will always be copied.
Will it break anything?

src/plugins_3rdparty/primer3/src/Primer3Task.h
yalgaer: You may have multiple constructors far from the variable declaration and forget to initialize the field on some of them. If you initialize it by default the error will be clear: `null` is always better than some random value.

Most of other languages do it by default. With C++ & default memory allocator you need this boilerplate code. And having a good and safe code (like other languages do) is more important than side-effects of the boilerplate code it require.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: `GTLineEdit::setText(GUITestOpStatus& os, const QString& lineEditName, ...` uses `GTWidget::findLineEdit` which calls `GTWidget::findExactWidget` which calls `GTWidget::findWidget`. What's the difference between `findWidget` and `findWidgetByType` which I am using? `findWidget` fails if it finds more than 1 widget with the given name. This method also expects a widget name. `findWidgetByType` works the other way around. It is used in `GTUtilsWorkflowDesigner::setCellValue`: https://github.com/ugeneunipro/ugene/blob/86e15e13fc5d7caf1c2806e27e8cad5a8759b44c/src/plugins/GUITestBase/src/GTUtilsWorkflowDesigner.cpp#L1098

DmitriiSukhomlinov:
Rebase your branch with master

src/plugins_3rdparty/primer3/src/Primer3Task.cpp
evelinabis: Dima purposefully changed pass by value to pass by pointer. I don't think it's correct to change this behavior back... So, I only have the QD that uses `Primer3SWTask`, but this task now takes a pointer to the settings, not the settings themselves. And I needed to change the behavior according to the new logic. This is how I see my task.

Why is a pointer used? I think Primer3's settings constructor became [non-trivial](https://github.com/ugeneunipro/ugene/blob/d407dc1a0cce97d9a3d597fab117eac8b5ec1cbd/src/plugins_3rdparty/primer3/src/Primer3TaskSettings.cpp#L48-L49), so copying became [forbidden](https://github.com/ugeneunipro/ugene/blob/d407dc1a0cce97d9a3d597fab117eac8b5ec1cbd/src/plugins_3rdparty/primer3/src/Primer3TaskSettings.h#L55-L58). It was decided to work around this with raw pointers.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: I agree that the method is generic, I don't agree that it is related to widgets. I don't know a better class though.

Should I make changes to `GTWidget` and `GTTableView` within this issue and this branch, or is it better to do another issue and/or another branch?

yalgaer: GTWidget

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: ok

src/plugins_3rdparty/primer3/src/Primer3Task.cpp
yalgaer: ok

yalgaer (dismissed):

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

yalgaer (approved):

PR #1000, opened 2022-10-16
UGENE-7459. SAFE_POINT when inserting a URL as the reference in the s…

yalgaer:
…equence generator


src/corelibs/U2Core/src/io/HttpFileAdapter.cpp
```diff
@@ -171,8 +166,7 @@ qint64 HttpFileAdapter::readBlock(char* data, qint64 size) {
 }
 
 qint64 HttpFileAdapter::writeBlock(const char*, qint64) {
-    SAFE_POINT(0, "Operation is not supported!", 0);
-    return 0;
+    FAIL("Operation is not supported!", 0);
```
DmitriiSukhomlinov, line 169: I'd like to add `return 0` because  the compiler must be worried about not all paths are reachable

DmitriiSukhomlinov (approved):

smlnkcv (approved):

src/corelibs/U2Core/src/io/HttpFileAdapter.cpp
yalgaer: The compiler sees an after pre-processor source and FAIL expands into 'return'

It is the opposite happens: the static analyzer in IDE detects that the 'return' after the FAIL() is unreachable and shows an error.

PR #1003, opened 2022-10-18
UGENE-7406. Crash when execute workflow Read annotation->Write anno…

yalgaer:
…tation with different formats. Part 2: The fix.


src/corelibs/U2Core/src/util/U2FeatureUtils.cpp
```diff
@@ -94,7 +94,12 @@ QList<U2Feature> U2FeatureUtils::exportAnnotationDataToFeatures(const QList<Shar
 
         // add sub-features
         if (annotation->location->isMultiRegion()) {
-            U2FeatureUtils::addSubFeatures(annotation->location->regions, annotation->location->strand, feature.id, rootFeatureId, dbiRef, os);
+            U2FeatureUtils::createSubFeatures(annotation->location->regions.mid(1),
```
DmitriiSukhomlinov, line 97: Do not get it - why did you remove the first element?

DmitriiSukhomlinov (changes requested):
A question

src/corelibs/U2Core/src/util/U2FeatureUtils.cpp
yalgaer: The first region is always saved as a part of the parent/top-level annotation.

DmitriiSukhomlinov (approved):

PR #1005, opened 2022-10-18
Translations update for U2Core

yalgaer:


src/corelibs/U2Core/src/dbi/U2DbiRegistry.cpp
```diff
@@ -277,7 +277,7 @@ U2DbiPool::~U2DbiPool() {
 
 U2Dbi* U2DbiPool::createDbi(const U2DbiRef& ref, bool create, U2OpStatus& os, const QHash<QString, QString>& properties) {
     U2DbiFactory* dbiFactory = AppContext::getDbiRegistry()->getDbiFactoryById(ref.dbiFactoryId);
-    CHECK_EXT(nullptr != dbiFactory, os.setError(tr("Invalid database type: %1").arg(ref.dbiFactoryId)), nullptr);
+    CHECK_EXT(nullptr != dbiFactory, os.setError(QString("Invalid database type: %1").arg(ref.dbiFactoryId)), nullptr);
```
DmitriiSukhomlinov: As far as I know, we do not translate `SAFE_POINT`s (because they are asserts and shouldn't appear) and translate `CHECK`s (because they really can appear in a usual situation). It this case, either get `tr` back or make it `SAFE_POINT_EXT`

DmitriiSukhomlinov (changes requested):
See comment, this is important

src/corelibs/U2Core/src/dbi/U2DbiRegistry.cpp
yalgaer: OK

DmitriiSukhomlinov (dismissed):

PR #1008, opened 2022-10-24
UGENE-7629 Long sequence filter resulted in a crash

rasputinkirill:
We do not need this assert. Bad memory allocation handling already present  in FindTask (this is the only place where RollingMatrix class is used).


src/corelibs/U2Algorithm/src/misc/RollingMatrix.h
```diff
@@ -30,11 +30,10 @@ namespace U2 {
 
 class U2ALGORITHM_EXPORT RollingMatrix {
 public:
+    // Always check memory overflow on the client side.
     RollingMatrix(int _sizeX, int _sizeY)
         : sizeX(_sizeX), sizeY(_sizeY), column0(0) {
-        // TODO: check memory overflow on the client side!
         U2_ASSERT(sizeX >= 0 && sizeY >= 0);
-        U2_ASSERT(getMatrixSizeInBytes(sizeX, sizeY) < INT_MAX);
```
yalgaer, line 37: How this change is related to the problem? If you want to fix the problem in the 'TODO' please to what is written in the TODO: fix the caller side (find algorithm).

src/plugins/GUITestBase/src/GUITestBasePlugin.cpp
```diff
@@ -1798,6 +1798,7 @@ void GUITestBasePlugin::registerTests(UGUITestBase* guiTestBase) {
     REGISTER_TEST(GUITest_regression_scenarios::test_7609);
     REGISTER_TEST(GUITest_regression_scenarios::test_7616);
     REGISTER_TEST(GUITest_regression_scenarios::test_7623);
+    REGISTER_TEST_WITH_TIMEOUT(GUITest_regression_scenarios::test_7629, 720000);
```
yalgaer: You know that I will never approve this right? Make the test smart and fast. (<1 min)

yalgaer (changes requested):
I do not see the fix in the patch (client side changes that will prevent exception). 
The test uses unacceptable timeout.

smlnkcv (dismissed):

yalgaer (changes requested):
About U2OpStatus: it could be a good solution for the factory method that return nullptr in case of error, but not for a constructor that always create an object.

If you want to use a constructor use the following solution

1) Use SAFE_POINT in the constructor. If SAFE_POINT happens the created matrix must be correctly initialized as an empty matrix
2) Add isValid/isEmpty method to the matrix. Check it on the caller side: if the matrix is empty or not.
3) Re-check that Find algorithm makes correct estimation before DynTable construction. FindAlgorithm (or any other algorithm) must never pass arguments that trigger safe points.

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
```diff
@@ -155,6 +156,10 @@ static void findInAmino(FindAlgorithmResultsListener* rl,
     QByteArray revPattern(pattern);
     TextUtils::reverse(revPattern.data(), patternLen);
 
+    if (!DynTable::isAcceptableMatrixDimensions(width, height)) {
+        coreLog.error(QObject::tr("Matrix with given dimensions too big for calculation."));
```
yalgaer: User does not know what is the 'matrix'. Change the text: 'The search pattern is too long.'

src/corelibs/U2Algorithm/src/misc/RollingMatrix.h
```diff
@@ -32,9 +34,8 @@ class U2ALGORITHM_EXPORT RollingMatrix {
 public:
     RollingMatrix(int _sizeX, int _sizeY)
         : sizeX(_sizeX), sizeY(_sizeY), column0(0) {
-        // TODO: check memory overflow on the client side!
-        U2_ASSERT(sizeX >= 0 && sizeY >= 0);
-        U2_ASSERT(getMatrixSizeInBytes(sizeX, sizeY) < INT_MAX);
+        CHECK(sizeX >= 0 && sizeY >= 0, );
```
yalgaer: This should be assertion because it creates a matrix in invalid state.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2886,6 +2886,48 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+    GTUtilsDialog::waitForDialog(os, new PopupChooserByText(os, QStringList() << "Copy/Paste" << "Copy selected sequence"));
+    GTMenu::showContextMenu(os, GTWidget::findWidget(os, "ADV_single_sequence_widget_0"));
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    //3. Paste it to project filter
+    //Expected: no crash, here is error in log
+    GTLogTracer logTracer;
+    auto nameFilterEdit = GTWidget::findLineEdit(os, "nameFilterEdit");
+    GTLineEdit::setText(os, nameFilterEdit, GTClipboard::text(os), true, true);
+    //necessary sleep to activate search task(s) after pasting sequence
+    GTGlobals::sleep();
```
yalgaer: Use checkTaskIsPresent if you need to wait for a task.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2886,6 +2886,48 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    CHECK_SET_ERR(logTracer.hasErrors(), "Log should contain errors");
```
yalgaer: Check for exact error message

yalgaer (changes requested):

rasputinkirill:
It is not possible to replace first sleep() occurrence with checkTaskIsPresent(), because with too long pattern, the task disappears from the tree too quickly for the method to catch it.
[build with both sleep() replaced](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/GuiLinux/249000)

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
yalgaer: The text was not fixed.
Do not use QObject::tr when you have a better context: use some of FindAlgorithm QObject classes.

src/corelibs/U2Algorithm/src/misc/RollingMatrix.h
```diff
@@ -32,7 +34,6 @@ class U2ALGORITHM_EXPORT RollingMatrix {
 public:
     RollingMatrix(int _sizeX, int _sizeY)
         : sizeX(_sizeX), sizeY(_sizeY), column0(0) {
-        // TODO: check memory overflow on the client side!
         U2_ASSERT(sizeX >= 0 && sizeY >= 0);
```
yalgaer: Why not SAFE_POINT as we discussed?

src/corelibs/U2Algorithm/src/misc/RollingMatrix.h
```diff
@@ -97,6 +98,10 @@ class U2ALGORITHM_EXPORT RollingMatrix {
     }
 
 protected:
+    static qint64 getMatrixSizeInBytes(int sizeX, int sizeY) {
```
yalgaer: Keep it public as before: other algorithms that create a lot of matrices and have memory limits may call it

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: Why this comment was resolved with no changes?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2886,6 +2886,47 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+    GTGlobals::sleep();
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    CHECK_SET_ERR(logTracer.getJoinedErrorString().contains("The search pattern is too long."), "Log should contain error");
+    
+    //4. Copy region with acceptable length 23 170 ( length * length < MAX_INT) 
+    GTUtilsDialog::waitForDialog(os, new SelectSequenceRegionDialogFiller(os, 1, 23170));
+    GTUtilsDialog::waitForDialog(os, new PopupChooser(os, {"Select", "Sequence region"}));
+    GTMenu::showContextMenu(os, GTWidget::findWidget(os, "ADV_single_sequence_widget_0"));
+
+    GTUtilsDialog::waitForDialog(os, new PopupChooserByText(os, QStringList() << "Copy/Paste"
```
yalgaer: Use the short form of lists: {"Menu1", "Menu2"}...

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2886,6 +2886,47 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+                                                                              << "Copy selected sequence"));
+    GTMenu::showContextMenu(os, GTWidget::findWidget(os, "ADV_single_sequence_widget_0"));
```
yalgaer: Dont we have a more readable way to show context menu for the sequence view?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2886,6 +2886,47 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    //5. Paste it to project filter
+    //Expected: no crash, no error in log
+    GTLogTracer logTracer2;
+    GTUtilsTaskTreeView::openView(os);
+    GTLineEdit::clear(os, nameFilterEdit);
+    GTLineEdit::setText(os, nameFilterEdit, GTClipboard::text(os), true, true);
+    GTUtilsTaskTreeView::checkTaskIsPresent(os, "Filtering project content");
```
yalgaer, line 2926: This check will work correctly (with no flakyness) only for a long living tasks. How long does the filtering task is active?

yalgaer (changes requested):
More issues + some older issues are not resolved. Please check.

src/corelibs/U2Algorithm/src/misc/RollingMatrix.h
rasputinkirill: Because you wrote "This should be assertion because it creates a matrix in invalid state.", if you mean c++ assert - it is useless in release mode, so i decided what you talking about U2_ASSERT. I still can't understand you want here.

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
rasputinkirill: FindAlgorithm is not a QObject successor. It contains only static functions inside. Other classes referred to it: FindAlgorithmResult, FindAlgorithmSettings also have nothing to do with QObject.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: ~65 sec

rasputinkirill:
my [answer ](https://github.com/ugeneunipro/ugene/pull/1008#issuecomment-1303020572 )to [this](https://github.com/ugeneunipro/ugene/pull/1008#discussion_r1014557875)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: It is not possible to replace first sleep() occurrence with checkTaskIsPresent(), because with too long pattern, the task disappears from the tree too quickly for the method to catch it.
[build with both sleep() replaced](http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/GuiLinux/249000)

src/corelibs/U2Algorithm/src/misc/RollingMatrix.h
yalgaer: SAFE_POINT is the way  UGENE' asserts in release mode and restores from internal errors. 
Please use SAFE_POINTS and never user U2_ASSERT/assertion.

If this is the last use of U2_ASSERT you can remove it from UGENE codebase

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: In this case replace with some other targeted waiter, like waitTasksFinished

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: So you expect user's computer/UGENE to hang for 65 seconds and then fail? :1st_place_medal:

yalgaer (changes requested):

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
```diff
@@ -155,6 +156,10 @@ static void findInAmino(FindAlgorithmResultsListener* rl,
     QByteArray revPattern(pattern);
     TextUtils::reverse(revPattern.data(), patternLen);
 
+    if (!DynTable::isAcceptableMatrixDimensions(width, height)) {
+        coreLog.error(QObject::tr("The search pattern is too long."));
```
yalgaer, line 160: Use FindAlgorithm...::tr() not QObject::tr

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
```diff
@@ -815,6 +820,10 @@ void FindAlgorithm::find(
     }
 
     try {
+        if (!DynTable::isAcceptableMatrixDimensions(width, height)) {
+            coreLog.error(QObject::tr("The search pattern is too long."));
```
yalgaer, line 824: Same here

yalgaer:
Closed as non-active. Please reopen when updated

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
rasputinkirill: 'tr': is not a member of 'U2::FindAlgorithm'

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
rasputinkirill: 'tr': is not a member of 'U2::FindAlgorithm'

rasputinkirill:
GUI test failed because there is not enogh RAM to run search with such large pattern.

src/corelibs/U2Algorithm/src/misc/FindAlgorithm.cpp
yalgaer: Ok

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2887,6 +2887,48 @@ GUI_TEST_CLASS_DEFINITION(test_7623) {
...
+    GTMenu::showContextMenu(os, GTUtilsMdi::activeWindow(os));
+
+    GTUtilsDialog::waitForDialog(os, new PopupChooserByText(os, QStringList() << "Copy/Paste"
+                                                                              << "Copy selected sequence"));
+    GTMenu::showContextMenu(os, GTUtilsSequenceView::getPanOrDetView(os));
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    // 3. Paste it to project filter
+    // Expected: no crash, here is info message in log and warning message box
+    GTLogTracer logTracer;
+    GTUtilsDialog::waitForDialog(os, new MessageBoxDialogFiller(os, QMessageBox::Ok, "The search pattern is too long. Patten's length will be decreased to 1000 symbols."));
```
yalgaer: You have typo in the test: patten -> pattern
I would suggest using this text: 

The search pattern is too long. Pattern was truncated to 1000 symbols.

yalgaer (changes requested):
Please fix the typo in the error message and replace U2_ASSERT with a safe-point

src/ugeneui/src/project_view/ProjectViewImpl.cpp
```diff
@@ -1095,9 +1096,15 @@ void ProjectViewImpl::sl_openStateView() {
 
 void ProjectViewImpl::sl_filterTextChanged(const QString& str) {
     SAFE_POINT(nullptr != projectTreeController, "NULL controller", );
-
     ProjectTreeControllerModeSettings settings = projectTreeController->getModeSettings();
-    settings.tokensToShow = str.split(QRegExp("\\s+"), QString::SkipEmptyParts);
+    if (str.length() > MAX_SEARCH_PATTERN_LENGTH) {
+        settings.tokensToShow = str.mid(0, 1000).split(QRegExp("\\s+"), QString::SkipEmptyParts);
```
yalgaer: 1. use `mid(0, MAX_SEARCH_PATTERN_LENGTH)`
2. move `settings.tokensToShow = str.split(QRegExp("\\s+")` out of the if (avoid code duplication)

yalgaer (approved):

PR #1009, opened 2022-10-25
UGENE-7697. Tree viewer settings are not saved correctly #995

dkandrov:
- Return backward compatibility with Qt 5.12


yalgaer:
We intentionally dropped Qt 5.12 support in the current release.

The only supported/tested version of Qt today is 5.15.

PR #1018, opened 2022-11-09
Remove unused si_updateBranch signal subscription that causes error i…

yalgaer:
…n log


evelinabis:
These connections were needed before Wrap mode. Why don't we need them now? If it connects from/to nullptr, then perhaps some functionality is lost.

yalgaer:
> These connections were needed before Wrap mode. Why don't we need them now? If it connects from/to nullptr, then perhaps some functionality is lost.

This patch has nothing to do with Wrap Mode. It only removes dead code from the tree viewer.

PR #1019, opened 2022-11-10
Use unified icon size on toolbar

yalgaer:
Without this patch the icon on MacOS looks bigger.

@EvelinaBiserova could you please check this fix, because I have no access to MacOS today.

PR #1021, opened 2022-11-12
Use unified icon size on toolbar-2

evelinabis:
Left -- view in v44, right -- on the current branch.
image
image
I'm currently waiting for a Linux build to check out the view there. #1019 can be closed.
Can I create an issue in Jira and add it to the pre-release table?

P.S. Your icon looks perfect even if it's cropped and scaled down to 16x16.


evelinabis:
Screenshot (28)
On Linux, in my case, everything also looks good.

yalgaer (approved):

PR #1024, opened 2022-11-17
Fix GeneCut icon on windows

DmitriiSukhomlinov:
GeneCut icon has **.ico** format, which was never used before. I added windows libs


yalgaer:
Please convert the icon into png, so we do not need add new libraries.

yalgaer (changes requested):
See the comment above

yalgaer (approved):

PR #1026, opened 2022-11-23
UGENE-7700. [Sam, crash] Long path

evelinabis:
https://ugene.dev/tracker/browse/UGENE-7700
GUI tests are not needed for this issue, we can do with only xml tests. But I didn't find a way to generate and set a custom ini file for a xml test. I can try to implement this functionality.


yalgaer:
> But I didn't find a way to generate and set a custom ini file for a xml test. I can try to implement this functionality.

You need an API XML test that will call the new function only.
See DynTableTests.cpp as an example

src/corelibs/U2Core/src/globals/GUrl.cpp
```diff
@@ -58,6 +58,11 @@ static QString makeFilePathCanonical(const QString& originalUrl) {
         prefix = ":";
         result = result.mid(1);
     } else {
+        // https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#:~:text=%5C%5C.%5CC%3A%5CTest%5CFoo.txt%20%5C%5C%3F%5CC%3A%5CTest%5CFoo.txt
+        // Paths starting with "\\.\" and "\\?\" appear to be the same, but Qt treats them differently.
+        if (result.startsWith("\\\\?\\")) {
```
yalgaer, line 63: this code should be called only on Windows, right?

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3445,6 +3445,82 @@ GUI_TEST_CLASS_DEFINITION(test_7697) {
...
+    QString coveredRegionsText;
+    class GetLabelTextScenario : public CustomScenario {
+        QLabel* label;
+        QString& text;
+
+    public:
+        GetLabelTextScenario(QLabel* label, QString& text)
+            : label(label), text(text) {
+        }
+        void run(GUITestOpStatus&) override {
```
yalgaer: We call `label->text()` from the current thread everywhere in the tests. This call is const and should be safe to be called from non-main thread. If you think it is unsafe create a GTLabel::getText reusable method

src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -71,7 +72,9 @@ static wchar_t* toWideCharsArray(const QString& text) {
 
 FILE* BAMUtils::openFile(const QString& fileUrl, const QString& mode) {
 #ifdef Q_OS_WIN
-    QScopedPointer<wchar_t> unicodeFileName(toWideCharsArray(fileUrl));
+    CHECK(!mode.contains('r') || QFileInfo::exists(fileUrl), nullptr);
```
yalgaer: Why do we need this check? Looks like unrelated to the main patch

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -3445,6 +3445,82 @@ GUI_TEST_CLASS_DEFINITION(test_7697) {
...
+            text = label->text();
+        }
+    };
+    GTThread::runInMainThread(
+        os, new GetLabelTextScenario(GTWidget::findLabel(os, "CoveredRegionsLabel"), coveredRegionsText));
+    CHECK_SET_ERR(coveredRegionsText.contains("<td><a href=\"0\">45 890 375</a></td><td align=\"center\">2 316</td>"),
+                  "Position 45 890 375 with 2 316 coverage isn't the first well-covered region");
```
yalgaer: Add the current coveredRegionsText value to the error message

src/corelibs/U2Core/src/globals/GUrl.cpp
evelinabis: It would be very strange if such a path came here on Unix. I don't know what this function will return on Unix with or without my changes. However, I can add a check on Windows here :)

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
evelinabis: I'll create 👍

src/corelibs/U2Formats/src/BAMUtils.cpp
evelinabis: For us, this is a common situation when a non-existent file is opened for reading. The syscall documentation is too vague: I don't think the _wfopen syscall is appropriate for such situations. It seems that in this case we should at least clear errno at the end of the function.

Indeed, this line doesn't apply to the current bug. I can put this in a separate pr without creating an issue in Jira, or do something else (including ignoring) that you write to me.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: Ok, but so far the current convention in the GUI tests is

1. Is the method `const` -> call it directly from either thread.
2. Is the method non-const -> call it from the main thread only.

Of course Qt says that any UI object methods can be called only from the UI thread.

src/corelibs/U2Formats/src/BAMUtils.cpp
yalgaer: You can keep it here and add a comment or some very explicit variable: 
```
bool isReadOnlyMode = ... (implement it correctly, I feel 'r-only' check is not complete.) 
CHECK(!isReadOnlyMode || File::exists())....
```

The easiest way to make the check correct is to provide 2 booleans to the method: isRead  and isBinary. But in this case you have to fix all caller sites.

As you see a correct implementation + migration of this part may be bigger than this fix. So may be you will want to put it into a separate patch (up to you).

evelinabis: Why is the `isBinary` flag needed?

evelinabis: > So may be you will want to put it into a separate patch (up to you).

Sounds right. I will do so.

yalgaer (commented):
The patch looks good now.
Please make Linux GUI tests green and I will approve

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

src/corelibs/U2Formats/src/BAMUtils.cpp
evelinabis: ✓

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp

yalgaer (approved):

evelinabis:
While working on this task, some tests failed. I analyzed them:
1. [GUI Linux 2651](https://github.com/ugeneunipro/ugene/blob/35dc2b0997f7c47373dd9c420e68570407b78da9/src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_2001_3000.cpp#L3629). I've checked many version (current debug, release 45) and sometimes, usually on the first run, the EntrezQueryTask for "AB797204.1 AB797210.1 AB797201.1" fails: "No results found corresponding to the query". This only happens on Linux, it has happened a few times. My guess is that this is not a ncbi issue, but something is going wrong on our side. It's very hard to reproduce, I have no idea what's going on.
2. [GUI Windows 4588(_2)](https://github.com/ugeneunipro/ugene/blob/35dc2b0997f7c47373dd9c420e68570407b78da9/src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_4001_5000.cpp#L3362). This test also fails from time to time. Failures happen due to the fact that in the save file dialog, unfortunately, sometimes the text starts typing after 10 sec (the time sometimes varies), but we have a limit on waiting for the end of text input up to 5 sec. The very similar test [2470](https://github.com/ugeneunipro/ugene/blob/35dc2b0997f7c47373dd9c420e68570407b78da9/src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_2001_3000.cpp#L2694) has no problem with this. These tests differ only in one place: I created an ["attempt to fix the test" patch](https://github.com/ugeneunipro/ugene/tree/attempt-to-fix-4588).
3. [GUI Windows 7448_4](https://github.com/ugeneunipro/ugene/blob/35dc2b0997f7c47373dd9c420e68570407b78da9/src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp#L1633). A bug, easy to reproduce, I created an [issue](https://ugene.dev/tracker/browse/UGENE-7738) in Jira.
4. [XML Linux GenomeAligner/workflow/test_0004](https://github.com/ugeneunipro/ugene/blob/35dc2b0997f7c47373dd9c420e68570407b78da9/tests/genome_aligner.xml#L8). I propose to start by [improving](https://github.com/ugeneunipro/ugene/tree/improving-xml-testing) the logging of timeout tests.
5. XML Linux Assert: prepareStage ? !taskRes.locked : taskRes.locked at src/[TaskSchedulerImpl.cpp:455](https://github.com/ugeneunipro/ugene/blob/35dc2b0997f7c47373dd9c420e68570407b78da9/src/corelibs/U2Private/src/TaskSchedulerImpl.cpp#L455). [Patch](https://github.com/ugeneunipro/ugene/tree/fix-xml-linux-assert).
6. XML Windows timeout error. No idea. Maybe the agents are too overloaded.

Also the [patch](https://github.com/ugeneunipro/ugene/tree/checking-opening-modes-bam-utils) for the [comment](https://github.com/ugeneunipro/ugene/pull/1026#discussion_r1031237981).

PR #1029, opened 2022-11-29
UGENE-6176 update maximum primers value

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-6176

This issue is a bit obsolete. In [UGENE-7193](https://ugene.dev/tracker/browse/UGENE-7193) the limit of primers to found was added (50 forward and 50 reverse). Which means, that there are 50*50=2500 products maximum we may have. I added the corresponding info to the manual (see [here](https://doc.ugene.net/wiki/display/UM/In+Silico+PCR)) and made a limit 2500 for products. That's a minor fix, not sure that GUI test is required

### DIFF
--- a/src/plugins/pcr/src/InSilicoPcrOptionPanelWidget.ui
+++ b/src/plugins/pcr/src/InSilicoPcrOptionPanelWidget.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>203</width>
+    <width>230</width>
     <height>656</height>
    </rect>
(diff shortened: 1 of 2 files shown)


yalgaer (dismissed):

DmitriiSukhomlinov:
Sorry, my bad. It happened, that "Maximum product" means "Maximum product size", not "Maximum products number". I reverted changes about this field and fixed the field name (also translated and fixed in docs)

yalgaer (approved):

PR #1030, opened 2022-11-29
UGENE-7677 Crash rerunning Extract consensus from assemblies with big sam data

DmitriiSukhomlinov:
The scenario is exaggerated. Problem appears when you import a SAM or BAM file with more then 1 000 000 reads to UGENEDB.


src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -346,7 +346,7 @@ static QString createNumericSuffix(int n) {
 
 static void bamSortBlocks(int n, int k, bam1_p* buf, const QString& prefix, const bam_header_t* h) {
     QString sortedFileName = n < 0 ? prefix + ".bam" : prefix + "." + createNumericSuffix(n) + ".bam";
-    const char* mode = n < 0 ? "w" : "w1";
+    const char* mode = n < 0 ? "w" : "w+";
```
yalgaer: `w1` comes from the original samtools package.

I would use keep a simple `w` for both cases (n<0 & n>=0)
or
pass `w1` to bam_dopen and `w` to BAMUtils::openFile

There is no reason to use `w+` here

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/Assembly_browser/GTTestsAssemblyBrowser.cpp
```diff
@@ -1045,5 +1045,15 @@ GUI_TEST_CLASS_DEFINITION(test_0037) {
...
 }
 
+GUI_TEST_CLASS_DEFINITION(test_0038) {
+    // Based on UGENE-7677
+    // 1. open view for _common_data\bam\more_then_100000_reads.bam
+    // Expected state: conversion finished without error
+    GTUtilsDialog::waitForDialog(os, new ImportBAMFileFiller(os));
+    GTFileDialog::openFile(os, testDir + "_common_data/bam/", "more_then_100000_reads.bam");
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    GTUtilsDialog::checkNoActiveWaiters(os);
```
yalgaer, line 1056: Could you please add a check that 
1) there are no errors in the log 
and 
2) that the active assembly browser window contains expected number of reads?

Otherwise the test will pass even if nothing has happened.

yalgaer (changes requested):

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/tests/common_scenarios/Assembling/Assembly_browser/GTTestsAssemblyBrowser.cpp
DmitriiSukhomlinov: OK

yalgaer:
The code looks good.
Please run GUI tests and I will approve

yalgaer (approved):

PR #1033, opened 2022-12-12
samtools/src/samtools/bgzf.c: Include "ugene_custom_io.h"

fweimer-rh:
This is necessary to obtain a prototype for the ugene_custom_open function.  Otherwise, the build fails with compilers which do not support implicit function declarations (which were removed from C in 1999).

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


yalgaer:
Thank you, I need one day to check this PR.

yalgaer:
Will be fixed here https://github.com/ugeneunipro/ugene/pull/1034

fweimer-rh:
@yalgaer Great, thank you!

PR #1035, opened 2022-12-15
UGENE-7745 Remove U2_ASSERT from project.

rasputinkirill:


src/corelibs/U2Private/src/TaskSchedulerImpl.cpp
```diff
@@ -147,7 +147,7 @@ bool TaskSchedulerImpl::processFinishedTasks() {
         TaskInfo* ti = priorityQueue[i];
         TaskInfo* pti = ti->parentTaskInfo;
         Task::State state = ti->task->getState();
-        U2_ASSERT(state != Task::State_Finished);
+        SAFE_POINT(state != Task::State_Finished, "Task state is not 'finished'.", hasFinished);
```
yalgaer: Add task details here, so the error message will be more informative (same in all other places)

src/corelibs/U2Private/src/TaskSchedulerImpl.cpp
```diff
@@ -382,10 +383,10 @@ QString TaskSchedulerImpl::tryLockResources(Task* task, bool prepareStage, bool&
     for (int i = 0, n = tres.size(); i < n; i++) {
         TaskResourceUsage& taskRes = tres[i];
         if (taskRes.prepareStageLock != prepareStage) {
-            U2_ASSERT(prepareStage ? !taskRes.locked : taskRes.locked);
+            SAFE_POINT(prepareStage ? !taskRes.locked : taskRes.locked, "Task lock state is not correct.", L10N::internalError());
```
yalgaer, line 386: No need to use L10N in case of SAFE_POINTs. Add the task name instead

yalgaer (commented):
Please check that all tests run successfully, add task names to the log messages and I will approve

yalgaer (approved):

PR #1036, opened 2022-12-15
Attempt to fix 7700

evelinabis:
Sometimes test 7700 fails because of timeout


evelinabis:
Moved to another branch.

PR #1043, opened 2022-12-27
UGENE-5036 U2SequenceImporter ambiguously detects Ext. RNA and Ext. D…

rasputinkirill:
…NA alphabets


rasputinkirill:
Test will be added after passing current xml tests.

src/corelibs/U2Core/src/util/U2SequenceUtils.h
```diff
@@ -135,6 +135,8 @@ class U2CORE_EXPORT U2SequenceImporter {
 
     const DNATranslation* aminoTT = nullptr;
     const DNATranslation* complTT = nullptr;
+
+    QByteArray alphabetCharacterHit = QByteArray();
```
yalgaer: `= QByteArray()` is not needed

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
```diff
@@ -463,23 +463,13 @@ void U2SequenceImporter::startSequence(U2OpStatus& os,
...
-        } else {
-            resAl = U2AlphabetUtils::deriveCommonAlphabet(blockAl, oldAl);
+    for (int i = 0; i < len; i++) {
+        if (Q_UNLIKELY(!alphabetCharacterHit.contains(data[i]))) {
+            alphabetCharacterHit.append(data[i]);
         }
-        CHECK_EXT(resAl != nullptr, os.setError(U2SequenceUtils::tr("Failed to derive sequence alphabet!")), );
     }
-
+    const DNAAlphabet* resAl = U2AlphabetUtils::findBestAlphabet(alphabetCharacterHit);
```
yalgaer, line 471: Why do you need `alphabetCharacterHit` if `U2AlphabetUtils::findBestAlphabet(alphabetCharacterHit)` will return the same ressult as `U2AlphabetUtils::findBestAlphabet(sequence-block-data)` ?

yalgaer (changes requested):
Current solution is good, but should be optimized: 
1) Keep array of unique symbols to detect alphabet like you do now.
2) Use QBitArray to check if a character is inside of the unique symbols array (see TextUtils as an example). Do not use (byteArray.contains())
3) You can use raw alphabet by default and compute the result alphabet only once: in finilizeSequence() method.

Please also add an API test for the code.

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
```diff
@@ -463,23 +463,19 @@ void U2SequenceImporter::startSequence(U2OpStatus& os,
...
-        } else {
-            resAl = U2AlphabetUtils::deriveCommonAlphabet(blockAl, oldAl);
+    for (int i = 0; i < len; i++) {
+        if (Q_UNLIKELY(!alphabetCharacterHit[data[i]])) {
+            alphabetCharacterHit[data[i]] = true;
         }
-        CHECK_EXT(resAl != nullptr, os.setError(U2SequenceUtils::tr("Failed to derive sequence alphabet!")), );
     }
-
+    QByteArray bytes;
```
yalgaer, line 471: Use TextUtils::selectIdx256() utility method: it does what you need.

tests/doc_format/embl/sequence/test_0004.xml
```diff
@@ -7,7 +7,7 @@
 
     <find-object-by-name index="obj3" doc="doc" name="AB000265" type="OT_SEQUENCE"/>
 
-    <check-sequence-alphabet-by-id obj="obj3" value="NUCL_DNA_DEFAULT_ALPHABET"/>
+    <check-sequence-alphabet-by-id obj="obj3" value="NUCL_DNA_EXTENDED_ALPHABET"/>
```
yalgaer: Which character in AB000265 makes the alphabet 'extended' ?

tests/doc_format/genbank/sequence/test_0008.xml
```diff
@@ -7,6 +7,6 @@
 
     <find-object-by-name index="obj2" doc="doc" name="B" type="OT_SEQUENCE"/>
 
-    <check-sequence-alphabet-by-id obj="obj2" value="AMINO_DEFAULT_ALPHABET"/>
+    <check-sequence-alphabet-by-id obj="obj2" value="RAW_ALPHABET"/>
```
yalgaer: Same question here: why the alphabet is RAW now?

rasputinkirill:
Tests restored, solution updated.

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
```diff
@@ -463,23 +463,19 @@ void U2SequenceImporter::startSequence(U2OpStatus& os,
...
-    if (oldAl != nullptr) {
-        if (oldAl->getType() == DNAAlphabet_AMINO && resAl->getType() == DNAAlphabet_NUCL) {
-            resAl = oldAl;
-        } else if (resAl->getType() == DNAAlphabet_AMINO && oldAl->getType() == DNAAlphabet_NUCL) {
-            oldAl = resAl;
-        } else {
-            resAl = U2AlphabetUtils::deriveCommonAlphabet(blockAl, oldAl);
+    for (int i = 0; i < len; i++) {
+        if (Q_UNLIKELY(!alphabetCharacterHit[data[i]])) {
+            alphabetCharacterHit[data[i]] = true;
```
yalgaer: you do not need if() here. Just use `alphabetCharacterHit[data[i]] = true with no if.

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
```diff
@@ -463,23 +463,19 @@ void U2SequenceImporter::startSequence(U2OpStatus& os,
...
         }
-        CHECK_EXT(resAl != nullptr, os.setError(U2SequenceUtils::tr("Failed to derive sequence alphabet!")), );
     }
-
+    QByteArray bytes;
+    for (int i = 0; i < alphabetCharacterHit.size(); i++) {
+        if (Q_UNLIKELY(alphabetCharacterHit[i])) {
```
yalgaer: Remove Q_UNLIKELY. This part is not a hotspot

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
```diff
@@ -463,23 +463,19 @@ void U2SequenceImporter::startSequence(U2OpStatus& os,
...
+            bytes.append(QChar(i));
```
yalgaer: Do not create an object. Use direct cast: `(char)i`

src/corelibs/U2Formats/src/EMBLGenbankAbstractDocument.cpp
```diff
@@ -155,6 +155,10 @@ void EMBLGenbankAbstractDocument::load(const U2DbiRef& dbiRef, IOAdapter* io, QL
             fullSequenceSize += gapSize;
         }
 
+        if (!merge) {
+            seqImporter.resetAlphabetCharacterHit();
```
yalgaer: Why don't you reset in seqImporter.startSequence()?
With your current solution you need to re-check that `resetAlphaberCharacterHit` is called in all usages of seqImporter (for all formats)

tests/doc_format/embl/sequence/test_0008.xml
```diff
@@ -7,7 +7,7 @@
 
     <find-object-by-name index="kkk" doc="doc" name="AB000264" type="OT_SEQUENCE"/>
 
-    <check-sequence-alphabet-by-id obj="kkk" value="AMINO_DEFAULT_ALPHABET"/>
+    <check-sequence-alphabet-by-id obj="kkk" value="AMINO_EXTENDED_ALPHABET"/>
```
yalgaer: Do not change the test. Make your algorithm case insensitive.

yalgaer (changes requested):

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
rasputinkirill: It is faster not to overwrite with the same value.

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
rasputinkirill: Why not? In every case there are less hits than skips.

src/corelibs/U2Formats/src/EMBLGenbankAbstractDocument.cpp
rasputinkirill: :thumbsup:

tests/doc_format/embl/sequence/test_0008.xml
rasputinkirill: I'ts not about case sensivity, i just edit wrong test.

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
yalgaer: If you can explain why we do not use Q_UNLIKELY with all other `if`s in our code you can explain this case too :)

src/corelibs/U2Core/src/util/U2SequenceUtils.cpp
yalgaer: Could you please share a performance test that shows that the optimization you do (and extra logic for it) has a positive effect? 

What I see here:
1) You read a memory  address
2) You have a branching instruction based on the value read.

In my suggested variant:
1) You read & write the same memory address
2) You do not have branching (if) - the loop contains less instructions and is easier to optimize by the compiler.

In case if you wrote the same value to memory that it stored before it may a no-op from the CPU memory management POV.

src/corelibs/U2Core/src/util/U2SequenceUtils.h
```diff
@@ -135,6 +137,8 @@ class U2CORE_EXPORT U2SequenceImporter {
 
     const DNATranslation* aminoTT = nullptr;
     const DNATranslation* complTT = nullptr;
+
+    std::bitset<256> alphabetCharacterHit;
```
yalgaer, line 141: nit: this is not a blocker, but why 256 but not CHAR_MAX (127) ?

rasputinkirill: added conversion to uchar

yalgaer (commented):
The fix looks good now. Please make the tests green and I will approve

yalgaer (approved):

PR #1044, opened 2022-12-28
UGENE-7551 SAFE_POINT on "Extract consensus from assemby" workflow wi…

rasputinkirill:
…th *.apr file


rasputinkirill:
Wrong input file in test, as result incorrect solution.
Actually hint added in previous commit do not need now. But in future it could prevent error\crash. So i decided to leave it.

yalgaer (commented):
Looks good, please run GUI tests and I will approve.

yalgaer (approved):

PR #1053, opened 2022-12-30
Fix test test_7154 and fix drag-n-drop GUI tests api

ichebyki:
Just add a middle point between start and stop
The new point is at distance >= system delta for drag-n-drop
The problem is that if stop-start < delta then internal system drag-n-drop state is invalid
and some tests may fail or hang


src/libs_3rdparty/QSpec/src/drivers/GTMouseDriver.cpp
```diff
@@ -47,14 +47,12 @@ bool GTMouseDriver::dragAndDrop(const QPoint& start, const QPoint& end) {
...
 
-#ifdef Q_OS_WIN
-    // TODO: check if the 'middle point' logic is still needed on Windows.
     QPoint middlePoint = (end + start) / 2;
-    bool useMiddlePoint = (end - middlePoint).manhattanLength() > 2 * QApplication::startDragDistance();
-    if (useMiddlePoint) {
-        DRIVER_CHECK(moveTo(middlePoint), QString("Mouse could not be moved to point (%1, %2)").arg(middlePoint.x()).arg(middlePoint.y()));
+    bool useMiddlePoint = (middlePoint - start).manhattanLength() > 2 * QApplication::startDragDistance();
+    if (!useMiddlePoint) {
+        middlePoint = start + QPoint(3 * QApplication::startDragDistance(), 3 * QApplication::startDragDistance());
```
yalgaer, line 54: As the result of this math `middlePoint` may be out of the screen. 

Add a check that the middlePoint is still a valid point or use a more advanced algorithm to select mid-point in this case.

src/libs_3rdparty/QSpec/src/drivers/GTMouseDriver.cpp
```diff
@@ -47,14 +47,12 @@ bool GTMouseDriver::dragAndDrop(const QPoint& start, const QPoint& end) {
     DRIVER_CHECK(press(), "Mouse button was not be pressed");
     GTThread::waitForMainThread();
 
-#ifdef Q_OS_WIN
-    // TODO: check if the 'middle point' logic is still needed on Windows.
     QPoint middlePoint = (end + start) / 2;
-    bool useMiddlePoint = (end - middlePoint).manhattanLength() > 2 * QApplication::startDragDistance();
-    if (useMiddlePoint) {
-        DRIVER_CHECK(moveTo(middlePoint), QString("Mouse could not be moved to point (%1, %2)").arg(middlePoint.x()).arg(middlePoint.y()));
+    bool useMiddlePoint = (middlePoint - start).manhattanLength() > 2 * QApplication::startDragDistance();
```
yalgaer: nit: rename `useMiddlePoint` to something like 'isValidDefaultMiddlePointValue' because you use middle-point below regardless of the `useMiddlePoint` value.

yalgaer (changes requested):

src/libs_3rdparty/QSpec/src/drivers/GTMouseDriver.cpp
ichebyki: fixed

src/libs_3rdparty/QSpec/src/drivers/GTMouseDriver.cpp
ichebyki: fixed

yalgaer (dismissed):

yalgaer (approved):

PR #1080, opened 2023-01-10
UGENE-7617 Unable to save dotplot when removing sequence

rasputinkirill:


yalgaer (changes requested):
This proposed solution is not the best option.
The best option is not to loose DP and support saving.

In order to save DP when any of the sequence is removed you will need to cache sequence length & name only.

yalgaer (changes requested):
What if the sequence was renamed? We need to listen for the sequence modification events and update cached details.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -2898,6 +2900,25 @@ GUI_TEST_CLASS_DEFINITION(test_7616) {
...
+    GTUtilsDialog::waitForDialog(os, new MessageBoxDialogFiller(os, QMessageBox::Yes));
+
+    //2. Click "Remove sequence" (gray cross) on sequence widget toolbar
+    //3. Answer "Yes" in "Save dot-plot" dialog and choose valid path
+    //Expected state: no crash, no errors in the log
+    GTLogTracer lt;
+    auto toolbar = GTWidget::findWidget(os, "views_tool_bar_human_T1 (UCSC April 2002 chr7:115977709-117855134)");
+    GTWidget::click(os, GTWidget::findWidget(os, "remove_sequence", toolbar));
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+    CHECK_SET_ERR(!lt.hasErrors(), "Errors in log: " + lt.getJoinedErrorString());
```
yalgaer, line 2919: Add a check that the file was saved and is not empty.

yalgaer:
> What if the sequence was renamed? We need to listen for the sequence modification events and update cached details.

Please ignore this comment. There is no problem with names.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
rasputinkirill: done

yalgaer (approved):

PR #1087, opened 2023-01-11
UGENE-7515 Crash if run several workfow Calculate assembly consensus

DmitriiSukhomlinov:
https://ugene.dev/tracker/browse/UGENE-7515

This is the same problem we have with Primer 3, but for Samtools - the tool doesn't support parallel calculation.


yalgaer (changes requested):
When I explained the Primer3 problem/solution before I attached several examples of code that were unsafe (used global variables). Since we do not want to modify Primer3 code the solution was to add a lock.

Could you please point to the same code location in Samtools so we can estimate which solution is better: 1) Lock based 2) Fix of the bad variable.

Note, that with Samtools option 2 is more 'real' than with 'Primer3' because we already did a lot of modification to the original Samtools code and have no opportunity upgrading to some newer version: only a complete rewrite.

Lock based variable should also be checked correctly for 'memory-leaks' or 'old-value-reuse'

DmitriiSukhomlinov:
> Could you please point to the same code location in Samtools so we can estimate which solution is better: 1) Lock based 2) Fix of the bad variable.
- I've never done any modification of samtools guts and sure, that this is the worst idea ever (as the idea of modification of any 3rd party tool). This is the code, which was written by other people - you do not know its architecture and you may add more unpredictable problems than you fix.
- We have a big issue about samtools updating - you spend time for figuring the code out, but it'll be rewritten soon.

> Lock based variable should also be checked correctly for 'memory-leaks' or 'old-value-reuse'

Do not have any idea what are you talking about.

yalgaer:
> I've never done any modification of samtools guts and sure, that this is the worst idea ever (

This started from the first days we have Samtools and  "why it happened?" is out of the scope of the issue. Lets focus on the current state problem.

> We have a big issue about Samtools updating

This is a separate task not related to the current one. Since the whole Samtools code was rewritten the update task does not depend on the current embedded Samtools code: everything must be replaced.
We have no exact time when it will happen. May be never (?)

> Do not have any idea what are you talking about.

We need to understand the root cause before fixing it. You used very conservative (a last-chance) solution.
My question is: can we do better and keep the Samtools parallel?

If you can't answer this question please re-assign the bug to me.

DmitriiSukhomlinov:
> This started from the first days we have Samtools and "why it happened?" is out of the scope of the issue. Lets focus on the current state problem.
> 
> >

This is a large possibility that current problems appeared because of these fixes. See no way to continue this bad politic.



> We have no exact time when it will happen. May be never (?)
> 
> >

This is a VERY difficult task, so it required time, but should be done, because the old version is not crash-safe.



> My question is: can we do better and keep the Samtools parallel?

Can't say it right now

yalgaer:
> > My question is: can we do better and keep the Samtools parallel?
> 
> Can't say it right now

Ok, lets proceed once we have the full answer.

Otherwise how can we be sure that methods that are not safe to be executed in parallel are safe to be executed in a sequence (have no side-effects)?

More than that how can we be sure that the problem is in the Samtools, but not in our adapter code?

Closing this bug with no merge. 

Please feel free to reopen with more details.

PR #1094, opened 2023-01-13
Fix illegal memory access in TextUtils.cpp

yalgaer:


DmitriiSukhomlinov (changes requested):
What errors did it fix? Any scenarios, GUI tests, issues, just anything? For me, this commit looks like some spherical horse in a vacuum

yalgaer:
> For me, this commit looks like some spherical horse in a vacuum

Let me quote this for a collection. Thank you.

The patch is a trivial cleanup that does need any bug/issue and it is impossible to have a test because this illegal memory access does not trigger a crash. 

The patch is a 7 changed-lines long. I can explain any part of it if you want. Which part of the patch you want me to explain?

PR #1101, opened 2023-01-17
UGENE-7768 Too long CIGAR line is out of border

DmitriiSukhomlinov:
We do not have any features to read pop-up hints from Assembly Browser (even we can't target some specific read), so I decided not to make a test.

### DIFF
--- a/src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
+++ b/src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
@@ -66,19 +66,16 @@ AssemblyReadsAreaHint::AssemblyReadsAreaHint(QWidget* parent)
     setObjectName("AssemblyReadsAreaHint");
 }
 
-static QString getCigarString(const QString& ci, bool trimmed = false) {
+QString getCigarString(const QString& ci) {
     if (ci.isEmpty()) {
         return QObject::tr("no information");
     }
 
     QString cigar;
-    int cigarSize = ci.size();
-    bool sizeBecameLess = false;
-    if (trimmed) {
-        cigarSize = qMin(cigarSize, AssemblyReadsAreaHint::LETTER_MAX_COUNT);
-        if (cigarSize < ci.size()) {
-            sizeBecameLess = true;
-        }
+    int cigarSize = qMin(ci.size(), AssemblyReadsAreaHint::LETTER_MAX_COUNT);
+    bool isTrimmed = false;
+    if (cigarSize < ci.size()) {
+        isTrimmed = true;
     }
     for (int i = 0; i < cigarSize; ++i) {
         QChar ch = ci.at(i);
@@ -88,7 +85,7 @@ static QString getCigarString(const QString& ci, bool trimmed = false) {
             cigar.append(QString("<b>%1 </b>").arg(ch));
         }
     }
-    if (sizeBecameLess) {
+    if (isTrimmed) {
         cigar.append("...");
     }
     return cigar;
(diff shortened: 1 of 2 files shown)


src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
```diff
@@ -66,20 +66,31 @@ AssemblyReadsAreaHint::AssemblyReadsAreaHint(QWidget* parent)
...
-static QString getCigarString(const QString& ci) {
+static QString getCigarString(const QString& ci, bool trimmed = false) {
     if (ci.isEmpty()) {
         return QObject::tr("no information");
     }
 
     QString cigar;
-    for (int i = 0; i < ci.size(); ++i) {
+    int cigarSize = ci.size();
+    bool sizeBecameLess = false;
```
yalgaer: rename to something more readable like 'isTrimmed'
Or remove it completely and use 
`if (cigarSize > AssemblyReadsAreaHint::LETTER_MAX_COUNT) cigar.append("...");`

src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
```diff
@@ -66,20 +66,31 @@ AssemblyReadsAreaHint::AssemblyReadsAreaHint(QWidget* parent)
     setObjectName("AssemblyReadsAreaHint");
 }
 
-static QString getCigarString(const QString& ci) {
+static QString getCigarString(const QString& ci, bool trimmed = false) {
```
yalgaer: The method is over-designed. 
It has a single use in the code, so you do not need different states (trimmed/not-trimmed) for it: it is dead code.

You can just rename it to getCigarStringForTooltip() and always do the trimming.

yalgaer (changes requested):
The patch is correct, please fix the readability/boilerplate issues noted in comments.

src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
DmitriiSukhomlinov: Renamed as `isTrimmed`. You fix won't work - `int cigarSize = qMin(ci.size(), AssemblyReadsAreaHint::LETTER_MAX_COUNT);`, which means, that `cigarSize` is always less or equal than `AssemblyReadsAreaHint::LETTER_MAX_COUNT`

src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
DmitriiSukhomlinov: Yep, really, didn't notice

src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
yalgaer: In my pseudo-code I by the `cigarSize` I meant size of the original cigar: `ci.size()`

yalgaer (changes requested):
Lets resolve "cigarSize" comment, pass the tests and I will approve.

src/corelibs/U2View/src/ov_assembly/AssemblyReadsAreaHint.cpp
DmitriiSukhomlinov: OK

yalgaer:
Looks good. Please run GUI tests and I will approve

yalgaer (approved):

PR #1102, opened 2023-01-17
UGENE-7769 UGENECL has an execution error

DmitriiSukhomlinov:


yalgaer (changes requested):
AFAIU you do not need to delete `viewCtx` object at all, because `viewCtx` has the plugin instance as a QObject parent.

yalgaer:
Look good. Will approve after the tests.

yalgaer (approved):

PR #1105, opened 2023-01-19
Gcc warning fix

rasputinkirill:


rasputinkirill:
gcc warning check is green
http://ugene-teamcity.unipro.ru:8111/ugene-teamcity/buildConfiguration/CodeAnalysisGccWarnings?branch=gcc-warning-fix&mode=builds

yalgaer (changes requested):
Could you please use size_t but not uint?
On 64 bit systems these types are different

yalgaer (approved):

PR #1106, opened 2023-01-21
UGENE-7771. Crash if BAMUtils::openFile returns nullptr

yalgaer:
I not only fixed the bug but created a micro-framework (Nullable template) to detect if a pointer that can be a `nullptr` is used before any checks.


src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -209,15 +209,10 @@ static samfile_t* samopen_ugene(int fd, const char* mode, const QString& faiUrl,
...
-    FILE* file = BAMUtils::openFile(url, fileMode);
-    if (file == nullptr) {
-        return nullptr;
-    }
-    samfile_t* samfile = samopen_ugene(fileno(file), samMode, faiUrl, header);
-    if (samfile == nullptr) {
-        closeFileIfOpen(file);
-        return nullptr;
-    }
+    NP<FILE> file = BAMUtils::openFile(url, fileMode);
```
yalgaer, line 212: Using `NP<FILE>` instead of `FILE*` make it very visible for a code reviewer that the pointer can be `nullptr` and must be checked before used.

The unchecked access is guarded by a SAFE_POINT (runtime issue)

I can't find any non-boilerplate way to make this check compile-time.

PR #1109, opened 2023-01-23
Switch build to c++17 standard

yalgaer:
This change will allow us using more advanced language features and newer std libraries.

The negative side effect is that we have to bump a minimal supported gcc version from 5.4 to 7.0 (we are building with 7.5 for a couple of years)

Alternatively we can stay longer with C++14 but pick the features from the <experimental> libraries.


yalgaer:
Closing as failed to pass MacOS compatibliity: we have to upgrade our MacOS requirements to be able to switch to C++17 safely

PR #1113, opened 2023-01-25
UGENE-7778. QObject::connect(QObject, Unknown): invalid nullptr param…

yalgaer:
…eter in U2FileDialog


DmitriiSukhomlinov:
GUI test?

yalgaer:
> GUI test?

There are no measurable side-effects except the log.
As for the logging I'm planning to create a generic test that will fail if there is any QT warning in the log (need to find a way to catch it still).

PR #1114, opened 2023-01-25
UGENE-7741 Use Primer 3 melting temperature calculation in UGENE

DmitriiSukhomlinov:
I have created a registry for temperature calculation algorithms and picked out the formula which was in use in UGENE before as the first algorithm (called it "Rough") and the second one has been implemented with Primer3 functions. Also 50 XML test and 4 GUI tests have been added (all calculations with different possible parameters are checked in XML, in GUI test only GUI implementation features).

Shortly about implementation - all temperature parameters are stored to the placed they are used in. That means, that you can open several different sequences and use different temperature settings in each sequence. Primer Library stores settings in the database with primers. Workflow designer has its own architecture for storing parameters, so I even haven't touched that.


src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalc.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <U2Core/U2SafePoints.h>
+
+namespace U2 {
+
+
+RoughTempCalc::RoughTempCalc(TempCalcSettings* settings)
+    : BaseTempCalc(settings) {}
+
+double RoughTempCalc::getMeltingTemperature(const QByteArray& sequence) {
+    CHECK(!sequence.isEmpty(), 0.0)
```
yalgaer: Return INVALID_TM?

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.cpp
```diff
@@ -0,0 +1,56 @@
...
+const QString RoughTempCalcCmdFactory::ID = "Rough";
+
+RoughTempCalcCmdFactory::RoughTempCalcCmdFactory()
+    : TempCalcFactory(ID) {}
+
+BaseTempCalc* RoughTempCalcCmdFactory::createTempCalculator(TempCalcSettings* settings) const {
+    return new RoughTempCalc(settings);
+}
+
+BaseTempCalc* RoughTempCalcCmdFactory::createTempCalculator(QMap<QString, QVariant> mapSettings) const {
```
yalgaer: use const&

src/corelibs/U2Core/src/datatype/primers/PrimerStatistics.h
```diff
@@ -102,7 +105,9 @@ class U2CORE_EXPORT PrimerStatisticsCalculator {
 
 class U2CORE_EXPORT PrimersPairStatistics {
 public:
-    PrimersPairStatistics(const QByteArray& forward, const QByteArray& reverse);
+    PrimersPairStatistics(const QByteArray& forward, const QByteArray& reverse, BaseTempCalc* temperatureCalculator);
+    ~PrimersPairStatistics() = default;
+    Q_DISABLE_COPY_MOVE(PrimersPairStatistics);
```
yalgaer, line 108: Could you please document why do we need this special behavior?
It is looks like it should be safe to copy this structure by value.

src/corelibs/U2Formats/src/sqlite_dbi/SQLiteUdrDbi.cpp
```diff
@@ -199,6 +201,20 @@ OutputStream* SQLiteUdrDbi::createOutputStream(const UdrRecordId& recordId, int
     return new SQLiteBlobOutputStream(db, tableName(recordId.getSchemaId()).toLatin1(), field.getName(), recordId.getRecordId(), (int)size, os);
 }
 
+void SQLiteUdrDbi::createTable(const UdrSchemaId& schemaId, U2OpStatus& os) {
+    CHECK(!SQLiteUtils::isTableExists(tableName(schemaId), db, os), );
+    CHECK_OP(os, );
+
+    UdrSchemaRegistry* udrRegistry = AppContext::getUdrSchemaRegistry();
+    SAFE_POINT_EXT(nullptr != udrRegistry, os.setError(L10N::nullPointerError("UdrSchemaRegistry")), );
```
yalgaer: do not use yoda style

src/corelibs/U2Formats/src/sqlite_dbi/SQLiteUdrDbi.cpp
```diff
@@ -199,6 +201,20 @@ OutputStream* SQLiteUdrDbi::createOutputStream(const UdrRecordId& recordId, int
...
+
+    const UdrSchema* schema = udrRegistry->getSchemaById(schemaId);
+    SAFE_POINT_EXT(nullptr != schema, os.setError(L10N::nullPointerError("UdrSchema")), );
+    
+    createTable(schema, os);
+    CHECK_OP(os, );
```
yalgaer: this CHECK_OP is not needed

src/corelibs/U2Private/src/AppContextImpl.h
```diff
@@ -437,6 +438,11 @@ class U2PRIVATE_EXPORT AppContextImpl : public AppContext {
         projectFilterTaskRegistry = value;
     }
 
+    void setTempCalcRegistry(TempCalcRegistry* value) {
+        assert(tcr == nullptr || value == nullptr);
```
yalgaer: Replace assert with SAFE_POINT

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
```diff
@@ -368,14 +373,19 @@ void DNAStatisticsTask::computeStats() {
...
-        QList<U2Region> blocks = U2Region::split(region, 1024 * 1024);
+        QList<U2Region> blocks = U2Region::split(region, MAX_SIZE);
         for (const U2Region& block : qAsConst(blocks)) {
             if (isCanceled() || hasError()) {
                 break;
             }
             const QByteArray seqBlock = sequenceDbi->getSequenceData(seqRef.entityId, block, os);
             CHECK_OP(os, );
+
+            meltTempSeq.append(seqBlock, qMin(MAX_SIZE - meltTempSeq.size(), seqBlock.size()));
```
yalgaer, line 388: It does not makes sense to compute melting temperature for huge blocks. Return INVALID_TM if the sequence is above some reasonable limit (less than 1 block size). This way you can move `result.meltingTemp=` out of the blocks loop.
Like: `result.meltingTemp  = sequenceSize > 10k ? INVALID : calculateTm()`
(if TM is INVALID (or 0) it should not be shown in UI)

src/corelibs/U2Algorithm/src/registry/TempCalcRegistry.h
```diff
@@ -0,0 +1,76 @@
...
+class U2ALGORITHM_EXPORT TempCalcRegistry : public IdRegistry<TempCalcFactory> {
+public:
+    bool registerEntry(TempCalcFactory* t) override;
+    /**
+     * Get the default temperature calculator
+     * Get the factory with the corresponding ID @saveId
+     * If @saveId is empty - get the first registred factory
+     * @return pointer to the temperature calculator
+     */
+    BaseTempCalc* getDefaultTempCalculator(const QString& saveId = "") const;
```
yalgaer: nit: rename "get" to "create"? The one who calls the method is responsible for the resource de-allocation.

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.cpp
```diff
@@ -533,6 +552,22 @@ bool SequenceInfo::eventFilter(QObject* object, QEvent* event) {
     return false;
 }
 
+void SequenceInfo::statisticLabelLinkActivated(const QString& link) {
+    if (link == CAPTION_SEQ_MELTING_TEMPERATURE) {
+        auto dialog = new TempCalcDialog(annotatedDnaView->getActiveSequenceWidget(), temperatureCalculator->getSettings());
```
yalgaer: Why don't use use the standard UGENE way to deal with dialogs like:
```
    QObjectScopedPointer<MyDialog> dialog(new MyDialog());
    int res = dialog->exec();
    CHECK(!dlg.isNull() && res == QDialog::Accepted );
```
This default method is better because in the same line with `new` operator it defines the life range of the allocated resource (both from the memory & code span point of viewes).

Do you want to make the dialog non-blocking/allow multiple dialogs?
Note that you call updateCommonStatisticsData() here while the dialog is still present, so if updateCommonStatisticsData() will trigger other UI it will have a conflict.

I would recommend just to rework it to the standard pattern to simplify the review of side-effects

src/corelibs/U2View/src/temperature/TempCalcDialog.h
```diff
@@ -0,0 +1,62 @@
...
+     */
+    TempCalcSettings* getTemperatureCalculatorSettings() const;
+
+private:
+    TempCalcWidget* tempCalcWidget = nullptr;
+};
+
+}
+
+#endif
```
yalgaer: change to 'pragma once' all ifdefs on new headers

src/corelibs/U2View/src/temperature/methods/RoughTempCalcWidget.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <U2Algorithm/BaseTempCalc.h>
+
+namespace U2 {
+
+RoughTempCalcWidget::RoughTempCalcWidget(QWidget* parent, const QString& id) :
+    BaseTempCalcWidget(parent, id) {}
+
+
+TempCalcSettings* RoughTempCalcWidget::getSettings() const {
+    auto settings = new TempCalcSettings;
```
yalgaer: Rename the method to new/create from get -> it creates a resource that caller must manage after that.
Or change the settings from pointer to a value type

src/plugins/CoreTests/src/DnaStatisticsTests.cpp
```diff
@@ -132,16 +177,42 @@ void GTest_DnaStatisticsTest::init(XMLTestFormat*, const QDomElement& element) {
...
+        regions << reg;
+    }
 
-    task = new DNAStatisticsTask(sequenceObject->getAlphabet(), sequenceObject->getEntityRef(), regions);
+    task = new DNAStatisticsTask(sequenceObject->getAlphabet(), sequenceObject->getEntityRef(), regions, temperatureCalculator);
     addSubTask(task);
 }
 
+bool checkDoubleProperty(double value, double expectedValue) {
+    if (qAbs(value - expectedValue) > qMax(qAbs(value / 1000), 0.005)) {
```
yalgaer: nit: can be a 1-liner return;

src/plugins/pcr/src/FindPrimerPairsWorker.h
```diff
@@ -79,7 +84,8 @@ class FindPrimerPairsWorkerFactory : public DomainFactory {
 class FindPrimersTask : public Task {
     Q_OBJECT
 public:
-    FindPrimersTask(const QString& outputFileUrl, const QList<DNASequence>& sequences);
+    FindPrimersTask(const QString& outputFileUrl, const QList<DNASequence>& sequences, BaseTempCalc* temperatureCalculator);
+    ~FindPrimersTask();
```
yalgaer: I see a lot of boilerplate code do deallocate 'temperatureCalculator' across the code. 
Use shared pointer?

src/plugins/pcr/src/InSilicoPcrTask.h
```diff
@@ -22,31 +22,33 @@
...
 
 namespace U2 {
 
-class InSilicoPcrTaskSettings {
-public:
-    InSilicoPcrTaskSettings();
+struct InSilicoPcrTaskSettings {
+    InSilicoPcrTaskSettings() = default;
+    ~InSilicoPcrTaskSettings() = default;
+    Q_DISABLE_COPY_MOVE(InSilicoPcrTaskSettings)
```
yalgaer: Will making 'temperatureCalculator' a shared pointer help to keep the settings a plain data structure with no copy-magic?

src/corelibs/U2Algorithm/src/registry/TempCalcRegistry.h
yalgaer: It is a good example where you should return a shared pointer instead of raw pointer. I see that the code below has a lot of non-related changes just because it has to deal with memory management.

src/plugins/GUITestBase/src/GTUtilsOptionPanelSequenceView.cpp
```diff
@@ -360,6 +362,42 @@ void GTUtilsOptionPanelSequenceView::openAnnotationParametersShowHideWidget(HI::
...
+                found = true;
+                break;
+            }
+        }
+        if (found) {
+            break;
+        }
+    }
+
+}
```
yalgaer, line 393: Fail if not found?

src/plugins/GUITestBase/src/GTUtilsOptionPanelSequenceView.cpp
```diff
@@ -360,6 +362,42 @@ void GTUtilsOptionPanelSequenceView::openAnnotationParametersShowHideWidget(HI::
...
+    for (int column = 0; column < statisticsLabel->height(); column += 10) {
+        bool found = false;
+        for (int row = 0; row < statisticsLabel->width(); row += 10) {
+            GTThread::waitForMainThread();
+            QPoint newMousePos = labelGlobalPos + QPoint(column, row);
+            GTMouseDriver::moveTo(newMousePos);
+            Qt::CursorShape shape = statisticsLabel->cursor().shape();
+            if (shape != Qt::ArrowCursor) {
+                GTMouseDriver::click();
+                found = true;
```
yalgaer: no need to use `bool found`: just return

src/plugins/pcr/src/PrimerLibraryWidget.cpp
```diff
@@ -136,11 +143,42 @@ void PrimerLibraryWidget::sl_exportPrimers() {
...
 }
 
+void PrimerLibraryWidget::sl_openTemperatureSettings() {
+    U2OpStatusImpl os;
+    PrimerLibrary* library = PrimerLibrary::getInstance(os);
+    CHECK_OP_UI(os, );
+
+    auto dialog = new TempCalcDialog(this, library->getTemperatureSettings());
+    dialog->open();
+    connect(dialog, &QDialog::finished, this, [this, library](int result) {
```
yalgaer: Use a simpler standard code to manage dialogs (see another comment about dialogs above)

src/plugins/pcr/src/TempCalcDelegate.h
```diff
@@ -0,0 +1,51 @@
...
+public:
+    TempCalcDelegate(QObject* parent = nullptr);
+
+    QVariant getDisplayValue(const QVariant& value) const override;
+    PropertyDelegate* clone() override;
+    QWidget* createEditor(QWidget* parent, const QStyleOptionViewItem& option, const QModelIndex& index) const override;
+    PropertyWidget* createWizardWidget(U2OpStatus& os, QWidget* parent) const override;
+
+    void setEditorData(QWidget* editor, const QModelIndex& index) const override;
+    void setModelData(QWidget* editor, QAbstractItemModel* model, const QModelIndex& index) const;
```
yalgaer: nit: document non-overriden methods

src/plugins/pcr/src/TempCalcPropertyWidget.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto settingsId = settingsVariantMap.value(TempCalcSettings::KEY_ID).toString();
+    lineEdit->setText(settingsId);
+    auto newSettigns = AppContext::getTempCalcRegistry()->getById(settingsId)->createDefaultTempCalcSettings();
+    newSettigns->fromVariantMap(settingsVariantMap);
+    delete tempSettings;
+    tempSettings = newSettigns;
+}
+
+void TempCalcPropertyWidget::sl_showDialog() {
+    auto dialog = new TempCalcDialog(this, tempSettings);
```
yalgaer: Use smart pointer to manage the dialog

src/plugins/pcr/src/TempCalcPropertyWidget.h
```diff
@@ -0,0 +1,60 @@
...
+    TempCalcPropertyWidget(QWidget* parent = nullptr, DelegateTags* tags = nullptr);
+    ~TempCalcPropertyWidget();
+
+    QVariant value() override;
+
+public slots:
+    void setValue(const QVariant& value) override;
+
+private slots:
+//    void sl_textEdited();
```
yalgaer: nit: cleanup

yalgaer (changes requested):
Looks good, please fix 2 main blockers and I will approve

1) Use shared pointer for the calculators - so you do not need to handle deletion everywhere and change settings structures to pointers

2) Use QDialogs the way we use it in other places: with smart pointers instead of signal/slots/delete

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalc.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/datatype/primers/PrimerStatistics.h
DmitriiSukhomlinov: OK

src/corelibs/U2Formats/src/sqlite_dbi/SQLiteUdrDbi.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Formats/src/sqlite_dbi/SQLiteUdrDbi.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Private/src/AppContextImpl.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
DmitriiSukhomlinov: Don't get your point.

> Return INVALID_TM if the sequence is above some reasonable limit (less than 1 block size).

What is "one block size"? If you open human_T1.fa and have no selection, the there is one block and its size is about 200'000 bases long, but if you select "1..10,20..30,40..50" there are three block, but just 30 bases long.

> This way you can move result.meltingTemp= out of the blocks loop.

 `result.meltingTemp=` is already out of block loop.

> result.meltingTemp  = sequenceSize > 10k ? INVALID : calculateTm()

`sequenceSize` could be as long as possible, but the selection size is several bases, these variables aren't related to each other.

src/corelibs/U2Algorithm/src/registry/TempCalcRegistry.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.cpp
DmitriiSukhomlinov: I use this way because Qt developers notes, that this solution is better. This is a part of [exec()](https://doc-snapshots.qt.io/qt5-5.15/qdialog.html#exec) documentation:

image

And, by the way, we already faced bugs, which appeared because of this problem (e.g. see [UGENE-7549](https://ugene.dev/tracker/browse/UGENE-7549)). I think, the argument that I need to rewrite it to `exec()` because we have it everywhere isn't that good, it just means that we have been using the incorrect solution for the whole time.

src/corelibs/U2View/src/temperature/TempCalcDialog.h
DmitriiSukhomlinov: Really, now? You decided to start doing that exactly from this issue? Listen, this could be easily automated - it would be much easier to write a separate script, which will change it in the whole UGENE source files. I can create the issue and do it. It's a bit tedious to do it manually right now, I'd rather not.

src/corelibs/U2View/src/temperature/methods/RoughTempCalcWidget.cpp
DmitriiSukhomlinov: OK

src/plugins/CoreTests/src/DnaStatisticsTests.cpp
DmitriiSukhomlinov: OK

src/plugins/pcr/src/InSilicoPcrTask.h
DmitriiSukhomlinov: Yes, it will, I added this restriction because I added a usual pointer as a member, and if you have a pointer member, you should overall copy-move operators-constructors and a destructor - otherwise it could cause memory-pointing errors (deletion is a sort of it). Both my and your solution works good, but yours is a bit more modern, I think

src/plugins/GUITestBase/src/GTUtilsOptionPanelSequenceView.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/GTUtilsOptionPanelSequenceView.cpp
DmitriiSukhomlinov: OK

src/plugins/pcr/src/PrimerLibraryWidget.cpp
DmitriiSukhomlinov: See [my comment](https://github.com/ugeneunipro/ugene/pull/1114#discussion_r1087581524)

src/plugins/pcr/src/TempCalcDelegate.h
DmitriiSukhomlinov: Overriden too

src/plugins/pcr/src/TempCalcPropertyWidget.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.cpp
yalgaer: You can create another bug that will focus on the problem like
1) Rewrite 100+ places of using ->exec()
2) Rewrite 100+ places of using QMessageBox

This is not related to Primer3 and requires additional discussion.

About fix for UGENE-7549: that was a bug in Qt 5.12.x but not in UGENE.
Now with Qt 5.15  that code was already migrated (and tested) to the standard way:  using ->exec

I suggest removing this question from the Primer3 PR discussion: I want to proceed with approval and this unrelated issue will block the approval. The issue you described may be much bigger than this PR and may require a special helper code to avoid connect/delete boilerplate everywhere.

src/corelibs/U2View/src/temperature/TempCalcDialog.h
yalgaer: Please check the code: all corelibs are already using pragmas. I'm asking to follow the common style in newly added files.

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
yalgaer: You changed the original algorithm and introduced a limit 1MB for the melting temperature to work.
I suggest to change 1MB (default buffer size) to a different constant that has a special meaning: MAX_LENGTH_TO_ESTIMATE_TEMPERATURE. I think the reasonable value here will be much less than 1MB.

src/plugins/pcr/src/FindPrimerPairsWorker.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
DmitriiSukhomlinov: Geneious and SnapGene doesn't have this kind of limitation, I checked right now

src/corelibs/U2View/src/temperature/TempCalcDialog.h
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.cpp
DmitriiSukhomlinov: What if I say "No"? Because nothing convincing I read from above, just common phrases, which means in generally, "I'm a teamleader, so you should do what I said"

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
yalgaer: But now we do have the limitation, right? I see that you compute TM only for the first 1Mb right?

Here is my logic: the old code didn't have this limitation. The new code has and set it to some non-related value. Could you please document this constant, why 1Mb? I'm OK with your size limit,  but would limit it to much smaller value reasonable for tasks where the temperature is needed.

data/workflow_samples/Scenarios/pcr.uwl
```diff
@@ -126,6 +126,8 @@ workflow "In silico PCR"{
                         in-silico-pcr.max-product {
                         }
                         in-silico-pcr.use-ambiguous-bases {
+                        }
+						in-silico-pcr.temperature-settings {
```
yalgaer: nit: align blocks

src/corelibs/U2Algorithm/src/registry/TempCalcRegistry.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    return result;
+}
+
+void TempCalcRegistry::saveSettings(const QString& saveId, TempCalcSettings* settings) {
+    savedSettings.insert(saveId, settings->toVariantMap());
+}
+
+}
```
yalgaer: nit: do not keep unterminated lines. While for *.cpp this is just a bad style, for the headers it is an error.

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
```diff
@@ -0,0 +1,97 @@
...
+     * It's required it wee need to save this settings somewhere
+     * @return a map with the parameter id as a key and the corresponding parameter value as value
+     */
+    virtual QMap<QString, QVariant> toVariantMap() const;
+    /**
+     * Load settings from variant map
+     * Variant map could be loaded from some store
+     * @mapSettings a map with the parameter id as a key and the corresponding parameter value as value
+     */
+    virtual void fromVariantMap(const QMap<QString, QVariant>& mapSettings);
```
yalgaer: This method does not say what happens with the current state. It is completely reset or appended?

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
```diff
@@ -0,0 +1,97 @@
...
+
+    QString id;
```
yalgaer: Need some docs here. Is it algorithm type or some storage key per object?

src/corelibs/U2View/src/temperature/methods/RoughTempCalcWidget.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+class RoughTempCalcWidget : public BaseTempCalcWidget {
+    Q_OBJECT
+public:
+    RoughTempCalcWidget(QWidget* parent, const QString& id);
+
+    TempCalcSettings* createSettings() const override;
+    void restoreFromSettings(TempCalcSettings* settings) override;
+
+private:
```
yalgaer: nit: remove empty private scope

yalgaer (changes requested):
Important open topics are

1) Why settings are passed as raw pointer?
It creates a lot of boilerplate code with memory management and unclear ownership.
Could you please wrap it with a shared ptr too?

A better fix would be to remove this type completely - because it is not a real data type: it is QVariantMap that is packed/unpacked inside of the private code of the temp-algorithm.

I'm Ok with either fix: shared-ptr or a by-value passed QVariantMap - both are better than the raw-ptr


2) Check comment about 1Mb buffer limit for temp calc in sequence statistics widget

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
```diff
@@ -0,0 +1,97 @@
...
+#include <QVariant>
+
+#include <U2Core/global.h>
+
+namespace U2 {
+
+/**
+ * Settings for the temperature calculations
+ */
+struct U2ALGORITHM_EXPORT TempCalcSettings {
```
yalgaer: You expect this struct to be inherited and use it a lot as a pointer to the parent class: 
-> make it a class and add a virtual destructor, so deallocation will work correctly.

src/corelibs/U2View/src/temperature/BaseTempCalcWidget.h
```diff
@@ -0,0 +1,61 @@
...
+    Q_OBJECT
+public:
+    BaseTempCalcWidget(QWidget* parent, const QString& id);
+
+    /**
+     * Get settings, which were set up.
+     * Use this function when settings up is done.
+     * @return pointer to the settings object
+     */
+    virtual TempCalcSettings* createSettings() const = 0;
```
yalgaer: If there any reason to create empty settings? 

You can pass QVariantMap to `createSettings` and return settings in initialized state:

1) no need to have  'fromVariantMap()': settings can accept QVariantMap in constructor.
2) there will be no such thing as 'uninitialized settings': created but not loaded. Less states and less code is better.

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
DmitriiSukhomlinov: I can compute it for 1, 10 or 100 MB, doesn't matter. The point is the longest the sequence the less meling temperature changes.  I mean:
- If you have sequence 10 characters long its MT, 26.00 °C. If you add 1 character (length is 11 now) MT changes to 28.00 °C (2.00 °C difference).
- 100 characters long its MT, 75.81 °C. Add 10 characters - now MT 77.05 °C (1.24 °C difference).
- 1000 characters long its MT, 80.63 °C. Add 100 characters - now MT 80.43 °C (0.2 °C difference).
- 10000 characters long its MT, 80.46 °C. Add 1000 characters - now MT 80.49 °C (0.03 °C difference).

(this is the real data from human_T1.fa). You got the point, right? I take the sequence, add 10% of its length and compare MTs. The longer the sequence, the less MT differences we have. After some length this difference is so insignificantly, that there is no point even to calculate it. I decided to have a limit 1MB just because this is not a large amount of RAM and could be captured in one calculation process. The MT differences between 1MB and 10 MB would be so little, that there is no need of its calculation.

Also there is no need to write N\A because:
- Not a one app do this
- It's just weird and not clear what it means. We can calculate MT for the sequence of every length, the calculation algorithm has no upper sequence length limitation and allows us to do that, so why N\A?

data/workflow_samples/Scenarios/pcr.uwl
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/registry/TempCalcRegistry.cpp
DmitriiSukhomlinov: OK but, what  error? A compile one?

src/corelibs/U2View/src/temperature/methods/RoughTempCalcWidget.h
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
DmitriiSukhomlinov: > It is completely reset or appended?

Don't really understand what you mean, but I updated the description

src/corelibs/U2View/src/temperature/BaseTempCalcWidget.h
DmitriiSukhomlinov: Maybe you have a point, but I really can't get it. 

> You can pass QVariantMap to createSettings and return settings in initialized state 

What you mean? `createSettings()` does not have arguments? And why do I need `QVariantMap` here?

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
DmitriiSukhomlinov: > add a virtual destructor

I can assume, that I forgot some C++ bases, but why do we need virtual destructor here? We do not have pointers in a base class, so even if some heir won't call it when delete it nothing bad happens.

> make it a class

`class` and `struct` is absolutely the same but the fact that `class` is `private` by default, and `struct` is `public`. And it's better to use `struct` in case of objects like this - when we have a lot of public members, which could vary independently. [Bjarne Stroustrup said this](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c2-use-class-if-the-class-has-an-invariant-use-struct-if-the-data-members-can-vary-independently).

src/corelibs/U2View/src/temperature/BaseTempCalcWidget.h
DmitriiSukhomlinov: I removed `TempCalcSettings` at all

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
DmitriiSukhomlinov: `TempCalcSettings` and heirs are removed, settings are stored in `QVariantMap`

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
yalgaer: OK here is the question: the old had no limit, you have a limit and your limit comes from a very unrelated domain: buffer reader size.

A good solutions I see
1) Have no limit as it was before.
2) Put a reasonable limit that does not depend on buffer size

According to the data you put above the reasonable limit is 10k, right? 
So lets limit (and may be even speed up this way) TM calculation by 10k instead of using the current limit of 1Mb: which is read buffer size.

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
yalgaer: This is good: QVariantMap is the best option here.

About virtual destructors: If B extends A and you do not have a virtual destructor in the code like this: `A* a = createB(); delete a;` the destructor from B won't be called at all (behavior is undefined) -> memory leak.

src/corelibs/U2Algorithm/src/registry/TempCalcRegistry.cpp
yalgaer: Right: this is a convention every C++ file must end with a line break.
This is why GH shows these symbols in diffs.

Not having the line break may, especially in headers may cause unexpected results: usually compile time errors.

In CPP files it is less important but still may be a problem if somebody includes the file as a header (and we have the code like this in UGENE).

DmitriiSukhomlinov: OK, fixed

src/corelibs/U2Algorithm/src/temperature/BaseTempCalc.h
DmitriiSukhomlinov: Ok. Are we finished here?

yalgaer: Yes. Current variant of the solution is safe and simple

yalgaer (dismissed):

src/corelibs/U2View/src/ov_sequence/sequence_info/DNAStatisticsTask.cpp
DmitriiSukhomlinov: Having no limit as it was before isn't a solution, because `Primer3` should have the whole sequence as an argument, it is just impossible to set it in parts - and if you had a 1GB sequence it may cause some complications. And I do not like 10k because this is much smaller than even sample sequences we have. For example, even our main sample file `human_T1` will change it's temperature, so an this length is too small. If you want to increase this number, let it be `200'000`.

yalgaer (approved):

PR #1123, opened 2023-01-30
Support 100% zoom level in tree viewer.

yalgaer:
100% zoom level is enabled by default when tree is opened.

This is default behavior or other phy-tree viewers.
To switch to the 'fit-to-screen' mode use a dedicated button on toolbar.


DmitriiSukhomlinov:
Something wrong with "src/corelibs/U2Gui/images/zoom_out_tree.png"

yalgaer:
> Something wrong with "src/corelibs/U2Gui/images/zoom_out_tree.png"

There is no issue with the image in UGENE. I reconverted it again so GitHub shows no error now

PR #1133, opened 2023-02-06
UGENE-7788 Update melting temperature documentation

DmitriiSukhomlinov:


src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.cpp
```diff
@@ -24,10 +24,10 @@
 
 namespace U2 {
 
-const QString RoughTempCalcCmdFactory::ID = "Rough";
+const char* RoughTempCalcCmdFactory::ID = QT_TR_NOOP("Rough");
```
yalgaer: Do not mix IDs and Display names.

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.h
```diff
@@ -42,7 +44,8 @@ class U2ALGORITHM_EXPORT RoughTempCalcCmdFactory : public TempCalcFactory {
     TempCalcSettings createDefaultTempCalcSettings() const override;
     BaseTempCalcWidget* createTempCalcSettingsWidget(QWidget* parent, const QString& id) const override;
 
-    static const QString ID;
+protected:
+    static const char* ID;
```
yalgaer: nit: If it is not public you can move it into the cpp. Public interface will be cleaner this way.

src/plugins/pcr/transl/russian.ts
```diff
@@ -1106,7 +1106,7 @@ Use &quot;Tools -&gt; Primer -&gt; Primer Library&quot; for managing the library
     <message>
         <location filename="../src/PrimerLibraryWidget.cpp" line="76"/>
         <source>Temperature</source>
-        <translation>Температура</translation>
+        <translation>Температура плавдения</translation>
```
yalgaer: fix typo

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.h
```diff
@@ -164,25 +164,25 @@ private slots:
     QSharedPointer<BaseTempCalc> temperatureCalculator;
 
     static const int COMMON_STATISTICS_TABLE_CELLSPACING;
```
yalgaer: nit: all these private constants can be moved into cpp. It will reduce code size and re-compilation time caused by change in this private data (like in your example now)

yalgaer (changes requested):

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.cpp
DmitriiSukhomlinov: Now this is the same and it won't be easy to change that behavior. The only thing I can is to rename "ID" with "NAME" (because, there is no need to use this parameter as an actual ID, its more the algorithm name)

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.h
DmitriiSukhomlinov: OK

src/plugins/pcr/transl/russian.ts
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/sequence_info/SequenceInfo.h
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/temperature/methods/RoughTempCalcCmdFactory.cpp
yalgaer: This is not a name but ID: you can store it into settings (as a last selected Tm algorithm id), use it in serialized WD schemas.

If support of the pair: visual name + persistent ID is too complex you can drop this part (translation of ID) from the patch and I will approve. Next I will help to make it the right way: with a separate ID & display text

DmitriiSukhomlinov: It would be difficult to fix now

yalgaer: I see no technical difficulty about this fix. 
You can drop it translation of ID from the patch and once merged I will add a patch on top of it.

DmitriiSukhomlinov: OK

yalgaer (approved):

PR #1134, opened 2023-02-06
Remove correspondingRectangularBranchItem from Tree Viewer data model

yalgaer:
No-op code cleanup


DmitriiSukhomlinov (changes requested):
What this fix should lead to?

yalgaer:
> What this fix should lead to?

It fixes nothing. It is a no-op cleanup.

PR #1135, opened 2023-02-07
Update QT_DISABLE_DEPRECATED_BEFORE from 5.7 to 5.9

yalgaer:


CMakeLists.txt
```diff
@@ -5,17 +5,17 @@
 #   * add Qt5_DIR variable to File->Settings->Build,Execution,Deployment->CMake pointing to valid QT dir.
 #   * add UGENE_TRANSLATION_FILE environment variable to the RUN configuration and point it to *.qm file produced by qmake build
 
-cmake_minimum_required(VERSION 3.9)
```
DmitriiSukhomlinov, line 8: Why did you changed CMAKE version?

yalgaer: It is not required. It was never updated since I introduced it before, so I put an LTS Linux version Linux here.

PR #1174, opened 2023-03-14
UGENE-7842 Can't add several fragments from the "Construct molecule" dialog

DmitriiSukhomlinov:


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -4254,5 +4257,36 @@ GUI_TEST_CLASS_DEFINITION(test_7806) {
...
+    // 3. Click "From project..."
+    // 4. Choose both sequences
+    // Expected: only one sequence could be selected, only one "Create Fragment" dialog has appeared, only one fragment added
+
+    GTUtilsDialog::waitForDialog(os, new GTSequenceReadingModeDialogUtils(os));
+    GTFileDialog::openFileList(os, dataDir + "samples/GenBank", { "murine.gb", "sars.gb" });
+    GTUtilsTaskTreeView::waitTaskFinished(os);
+
+    class Scenario : public CustomScenario {
+        void run(HI::GUITestOpStatus& os) {
```
yalgaer: nit: add override

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -4254,5 +4257,36 @@ GUI_TEST_CLASS_DEFINITION(test_7806) {
...
+            QWidget* dialog = GTWidget::getActiveModalWidget(os);
+
+            GTUtilsDialog::waitForDialog(os, new CreateFragmentDialogFiller(os));
+            GTUtilsDialog::waitForDialog(os, new ProjectTreeItemSelectorDialogFiller(os, { { "sars.gb", { "NC_004718" } }, { "murine.gb", { "NC_001363" } } } ));
+            GTWidget::click(os, GTWidget::findWidget(os, "fromProjectButton"));
+
+            auto fragmentListWidget = GTWidget::findListWidget(os, "fragmentListWidget", dialog);
+            CHECK_SET_ERR(fragmentListWidget->count() == 1, QString("Unexpected fragments size, expected: 1, current: %1").arg(fragmentListWidget->count()));
```
yalgaer, line 4302: You ask to select 2 and result with 1. 
The filler should fail, but not silently return 1, because it can't fulfill the request.

src/plugins/enzymes/src/ConstructMoleculeDialog.cpp
```diff
@@ -332,35 +332,30 @@ void ConstructMoleculeDialog::sl_onItemClicked(QTreeWidgetItem* item, int column
...
-                    DNAFragment frag = dlg->getFragment();
-                    QString fragItem = QString("%1 (%2) %3").arg(frag.getSequenceName()).arg(frag.getSequenceDocName()).arg(frag.getName());
-                    fragments.append(frag);
-                    fragmentListWidget->addItem(fragItem);
-                    break;
-                }
-            }
-        }
-    }
+    auto obj = objects.first();
```
yalgaer, line 344: The old code was very explicit to support multiple selection.
Why did you decide to change it to support only 1 element at a time?

yalgaer (changes requested):
The decision made in the patch is not unclear. Please provide more details.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
DmitriiSukhomlinov: Don't get, what you mean and what you want to change

src/plugins/enzymes/src/ConstructMoleculeDialog.cpp
DmitriiSukhomlinov: Current behaviout is weird and confusing - you can select several sequences, but the fhagment will be created just to the first one (because of the `break`, which ends the `for` cycle). So, obviously, we have two options:

1. Remove breake and make possible to create as many fragments as you want from the first try.
2. Make selection singular and user'll be avaliable to select only one sequnce - and create only one fragment.

Here you should understand, that the special "Create fragment" dialog should appear every selected sequence for. So, if you make multiple selection, several dialogs will appeare - one after another. If you select 10 sequence - 10 dialogs will appear. If select 20 - 20 dialogs. This is not good. Besides, this dialog has some strict view and not connected to any sequence - which means, that user can't even define, what sequence for he's creating a fragment in this dialog.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
yalgaer: You select 2 objects in Project view:
```
GTUtilsDialog::waitForDialog(os, new ProjectTreeItemSelectorDialogFiller(os, { { "sars.gb", { "NC_004718" } }, { "murine.gb", { "NC_001363" } } } ));
```

and say in the description:

```
Only one sequence could be selected
```

Could you confirm that ProjectTreeItemSelectorDialogFiller selects both, right? In this case you need to change the description

src/plugins/enzymes/src/ConstructMoleculeDialog.cpp
yalgaer: Ok sounds reasonable

PR #1175, opened 2023-03-14
UGENE-7825 Primer3 has shift in the result primers

DmitriiSukhomlinov:


src/plugins_3rdparty/primer3/src/Primer3TaskSettings.cpp
```diff
@@ -49,6 +49,18 @@ Primer3TaskSettings::Primer3TaskSettings() {
...
+    // Sequence character: ACGT...
+    //
+    // Primer3 numerates characters (by default) in a programming mode - the first character has index 0:
+    // Index number:       0123...
+    // Sequence character: ACGT...
+    //
+    // It causes one character shift is the result (UGENE-7825)
+    // But Primer3 has a special parameter to set a number to start indexing with - FIRST_BASE_INDEX (default value = 0)
+    // By settings this parameter to 1 we make Primer3 index characters from number 1
+    primerSettings->first_base_index = 1;
```
yalgaer, line 63: Does it mean that UGENE in internal model for primer coordinates in sequence uses math (U2Regions) that is 1 based, but not 0 based?

UGENE should use 1-based coordinates only during rendering values to UI (formatted values). 
Internally any coordinates are always 0-based.

DmitriiSukhomlinov: Yes, I'm talking about UI rendering in this case. Previous version of Primer3 also has this value set to 1 to avoid this problem, I just missed it when was doing upgrading

yalgaer: I thought Primer3TaskSettings is a data structure to configure how Primer3 algorithm works (non-ui).
And the default mode for primer3 is 0.

This way we can completely remove the method Primer3TaskSettings::getFirstBaseIndex and all logic around it.
But you do it the opposite way: set it to a non default value and adds logic around this parameter.

Check this diff: https://github.com/ugeneunipro/ugene/compare/primer3-first-base-index?expand=1
Looks like we do not really use getFirstBaseIndex() because the default primer3 settings always have this value set to 0 and we never change it.

WDYT? Can we make this patch other way, using default primer3 mode and 0 first base offset? 
If there is no way to avoid it we can use your approach.

yalgaer: I found that we do change this property in our code - so it is not readonly. Resolving.

PR #1188, opened 2023-03-24
UGENE-7836 Update restriction sites

DmitriiSukhomlinov:


yalgaer:
Please fix the GUI test and I will approve

DmitriiSukhomlinov:
Some additional improvements required:

- The "Find restriction sited" dialog became a bit slow. It happened because number of sites increased in ~4 times. This is because of `QTextEdit`, which contains all selected enzymes. The problem is that this field is slow a bit if you want to fill it with the really large text (just for estimation - in previous version of sites the all sites line was ~36'000 chars, for the current version - ~160'000 chars). I made two things: 1. morphed it to `QPlainTextEdit` - this class works faster (but still not instantly) 2. add 50'000 chars limit. Unfortunately, there is no simple way to increase speed of Qt text widgets, they are just slow (see [this question](https://stackoverflow.com/questions/63127788/qplaintextedit-display-slow-performace), and, especially, [this answer](https://stackoverflow.com/a/63129566/7273346) for details)
- Forbid to show sites if to many of them selected. This is not sites problem, but auto-annotation system (see [here](https://github.com/ugeneunipro/ugene/blob/49cfbae93d64a949ea245b9e6ad2a8a727a5abce/src/corelibs/U2Core/src/globals/AutoAnnotationsSupport.h#L101) for details). Actually, this is OK because not a single person would try to see 15'000 sites at the same time - it's just too many to recognize

yalgaer:
>     QTextEdit -> QPlainTextEdit

Ok. In case of the limit you can set it to a smaller value: 1k, 10k, etc and for the values beyond the limit show a text: 'N enzymes selected'

>     * Forbid to show sites if to many of them selected. This is not sites problem, but auto-annotation system (see [here](https://github.com/ugeneunipro/ugene/blob/49cfbae93d64a949ea245b9e6ad2a8a727a5abce/src/corelibs/U2Core/src/globals/AutoAnnotationsSupport.h#L101) for details). Actually, this is OK because not a single person would try to see 15'000 sites at the same time - it's just too many to recognize

This is already implemented: if I remember correctly there is a limit of enzyme count in the task. I see you made it not a warning, but an error -> OK

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -565,10 +582,8 @@ void FindEnzymesDialog::accept() {
 
     if (FindEnzymesAutoAnnotationUpdater::isTooManyAnnotationsInTheResult(advSequenceContext->getSequenceLength(), selectedEnzymes.size())) {
         QString message = tr("Too many results to render. Please reduce the search region or number of selected enzymes.");
-        int ret = QMessageBox::question(this, tr("Warning!"), message, QMessageBox::Cancel | QMessageBox::Ignore);
-        if (ret == QMessageBox::Cancel) {
-            return;
-        }
+        QMessageBox::information(this, tr("Warning!"), message, QMessageBox::Ok);
```
yalgaer: Change the title & type to error: it is not a warning anymore.

src/plugins/enzymes/src/FindEnzymesDialog.h
```diff
@@ -74,6 +74,7 @@ private slots:
     void saveFile(const QString& url);
     void setEnzymesList(const QList<SEnzymeData>& enzymes);
     void updateStatus();
+    int gatherCheckedNamesListString(QString& checkedNamesList);
```
yalgaer: nit: add `const`?

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -374,7 +387,11 @@ void EnzymesSelectorWidget::sl_saveSelectionToFile() {
     LastUsedDirHelper dir;
     dir.url = U2FileDialog::getSaveFileName(this, tr("Select enzymes selection"), dir.dir);
     if (!dir.url.isEmpty()) {
-        QFile data(dir.url);
+        QString urlToSave = dir.url;
+        if (!urlToSave.endsWith(".txt")) {
```
yalgaer: You should change the file name: a user was very explicit which name he wants.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -274,17 +290,13 @@ void EnzymesSelectorWidget::sl_filterTextChanged(const QString& filterText) {
...
-    for (int i = 0, n = tree->topLevelItemCount(); i < n; i++) {
-        auto gi = static_cast<EnzymeGroupTreeItem*>(tree->topLevelItem(i));
-        nChecked += gi->checkedEnzymes.size();
-        foreach (const EnzymeTreeItem* ci, gi->checkedEnzymes) {
-            checkedNamesList.append(ci->enzyme->id);
-        }
+    QString checkedNamesListString;
+    int nChecked = gatherCheckedNamesListString(checkedNamesListString);
+    if (checkedNamesListString.size() > 50'000) {
+        checkedEnzymesEdit->setPlainText(tr("Too many sites selected to show, click \"Save selection\" to export them to the separate file"));
```
yalgaer: "Too many" sounds like an error. Show the count instead.

yalgaer (changes requested):
Minor issues only + waiting for the green tests

src/plugins/enzymes/src/FindEnzymesDialog.h
DmitriiSukhomlinov: Nope, this is a hint to avoid extra copying (in out case this is important, because `checkedNamesList` could be very long). If your function is:
```
QString gatherCheckedNamesListString() {
    QString result;
    // do something, result is very large
    return result;
};
```
you call it lake this:
`QString checkedNamesList = gatherCheckedNamesListString(); // copy from return gatherCheckedNamesListString() r-value to checkedNamesList l-value `

To avoid it I pass `checkedNamesList` by reference:
```
int gatherCheckedNamesListString(QString& checkedNamesList) {
    // do something, checkedNamesList is very large
};

QString checkedNamesList;
gatherCheckedNamesListString(checkedNamesList); // no extra copying
```

Actually, this could be handled a bit more modern by using `std::move()`, but I just forgot about it and resolved it in this way.

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: You mean, you should not? Or what?

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.h
yalgaer: I meant const to the method, not to the parameter. Can the method be const? (not a blocker, a style issue)

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: Right, changing the file name is bad here because:
1) You give no chance to user to select anything else
2) This change worth a separate discussion: how UGENE behaves with file save dialogs everywhere: does it auto-fixes or just use the user provided input. Right now it does not auto-fixes and just follows the user input.

src/plugins/enzymes/src/FindEnzymesDialog.h
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
@yalgaer please, approve, the last time GUI test have just some random failure

DmitriiSukhomlinov:
2 GUI test failed - some flaky and the one, which has no relation to the code I wrote (again, some random)

yalgaer (approved):

PR #1207, opened 2023-04-10
UGENE-7849 Save dialog parameters when it's closed

DmitriiSukhomlinov:


src/corelibs/U2Gui/src/U2SavableWidget.cpp
```diff
@@ -188,10 +190,30 @@ MWMDIWindow* U2SavableWidget::getContextWindow() const {
...
+}
+
 QSet<QWidget*> U2SavableWidget::getCompoundChildren() const {
     return QSet<QWidget*>();
 }
 
+bool U2SavableWidget::isExcluded(const QString& childId) const {
+    auto wgt = getChildWidgetById(childId);
+    for (auto excludeWgt : qAsConst(excludeWidgets)) {
+        CHECK_CONTINUE(excludeWgt == wgt || excludeWgt->findChild<QWidget*>(childId) != nullptr);
```
yalgaer: nit: use std::any_of

yalgaer: q: why do you need this part: `excludeWgt->findChild<QWidget*>(childId)` ? I see that excluded widget can only come from ` auto excludeWgt = getChildWidgetById(excludeWgtId);`

src/corelibs/U2Gui/src/U2SavableWidget.h
```diff
@@ -44,17 +44,24 @@ class U2GUI_EXPORT U2SavableWidget {
...
+    bool isExcluded(const QString& childId) const;
 
     QWidget* wrappedWidget;
     MWMDIWindow* contextWindow;
     bool widgetStateSaved;
+    /*
+     * Widgets, which state shouldn't be saved.
+     * This behaviour works for the widget and all it's children
+     */
+    QList<QWidget*> excludeWidgets;
```
yalgaer: What is the widgets are dynamic (added/removed). 
It may be safier to keep ids only and check for savable widget flag only while saving

src/plugins/GUITestBase/src/runnables/ugene/plugins_3rdparty/primer3/Primer3DialogFiller.cpp
```diff
@@ -128,7 +128,11 @@ void Primer3DialogFiller::commonScenario() {
     }
 
     if (settings.notRun) {
-        GTKeyboardDriver::keyClick(Qt::Key_Escape);
+        if (isOsMac()) {
+            dialog->close();
```
yalgaer: You should not call API (main thread) methods in GUI tests. Click something like  'Cancel' or 'OK' button instead

src/plugins/GUITestBase/src/tests/common_scenarios/primer3/GTTestsCommonScenariosPrimer3.cpp
```diff
@@ -520,6 +522,41 @@ GUI_TEST_CLASS_DEFINITION(test_0023) {
...
+        Scenario(HI::GUITestOpStatus& _os)
+            : Filler(_os, "Primer3Dialog") {
+        }
+        void run() override {
+            QWidget* dialog = GTWidget::getActiveModalWidget(os);
+            GTUtilsDialog::add(os, new GTFileDialogUtils(os, sandBoxDir, "test_0024.txt", GTFileDialogUtils::Save));
+            GTWidget::click(os, GTWidget::findWidget(os, "saveSettingsButton", dialog));
+
+            if (isOsMac()) {
+                dialog->close();
```
yalgaer: Same here: mimic a real user behavior in UI tests. A user can't call API.

src/corelibs/U2Gui/src/U2SavableWidget.cpp
DmitriiSukhomlinov: This part is required if you want to exclude a bunch of widgets, which all have a common ancestor. For example, here I excluded 2 widgets - region selector and annotation settings. But, region selector contains several widgets - 2 labels, 2 line edits and popup menu:
image

And annotation settings have more than 10 widgets. I do not need to enumerate them all - just to add their common parent.

src/corelibs/U2Gui/src/U2SavableWidget.h
DmitriiSukhomlinov: That what I did first, but I haven't find any problems or unsafe scenarious about using pointers. It's just faster - wee call `getChildWidgetById(...)` just once for each widget instead of doing it every time this widget needed.

src/plugins/GUITestBase/src/runnables/ugene/plugins_3rdparty/primer3/Primer3DialogFiller.cpp
DmitriiSukhomlinov: That's a good point, fixed

src/plugins/GUITestBase/src/tests/common_scenarios/primer3/GTTestsCommonScenariosPrimer3.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Gui/src/U2SavableWidget.h
yalgaer: `I haven't find any problems` does not mean the problem can't exist: a good code makes the problem impossible to appear.
In this case it is very easy to write a complex widget (for example with tabs) where widgets are added/removed dynamically and your solution will produce a crash.

src/corelibs/U2Gui/src/U2SavableWidget.cpp
yalgaer: I see. Why do you call findChild twice for the same id? 
Once here `auto wgt = getChildWidgetById(childId);` and once here `excludeWgt->findChild<QWidget*>(childId)`
Are these 2 different hierarchies?

DmitriiSukhomlinov: I do not get this question, actually. I'll try to explain:

1. `excludeWgt == wgt` - check, that `excludeWgt` itself is not a widget, which should be exclude.
2. `excludeWgt->findChild<QWidget*>(childId) != nullptr` - check, that `excludeWgt` does not have children, which should be exclude

src/corelibs/U2Gui/src/U2SavableWidget.h
DmitriiSukhomlinov: I think, my solution is good and have no problems. You want me to rewrite it - just say.

yalgaer: Question about your solution: what if the cached widget is destroyed and replaced with another widget?
Do you still think that your solution is good?

src/corelibs/U2Gui/src/U2SavableWidget.cpp
yalgaer: I mean that your code iterates the whole `excludeWidgets` list subtree multiple times: once per each widget in the list.

This code is not good from the algorithmic POV:  all these widgets are within the same hierarchy, right? A single findChild should be enough. Next, for the child->parent relation you can check by parent: there is no need to iterate all children tree for it.

Next: check 'wgt' for nullptr. You solution will crash in user mode if the child does not exist (getChildWidgetById return nullptr).

DmitriiSukhomlinov: Not sure, that I did exactly what you wanted, check it

src/corelibs/U2Gui/src/U2SavableWidget.h
DmitriiSukhomlinov: Ok, fixed

DmitriiSukhomlinov:
All tests passed except some flaky one in "GUI Linux", @yalgaer please approve

yalgaer (approved):

PR #1213, opened 2023-04-13
UGENE-7862 Can't import large alignment

DmitriiSukhomlinov:


src/corelibs/U2Core/src/globals/Nullable.h
```diff
@@ -45,10 +45,6 @@ class NP {
         return value != anotherValue;
     }
 
-    operator T*() const {
```
yalgaer: This method should stay. Reason: all smart pointers have it.

You may think that using explicit .get() instead of auto-cast will help but this is not correct: you do not see the whole life-range of the pointer.
NP is designed to find errors in runtime (during GUI testing)

yalgaer:
Will review the rest of the patch tomorrow

src/corelibs/U2Core/src/globals/Nullable.h
DmitriiSukhomlinov: I'll take it back, but I had SO many errors because of it, that this is the less friendly interface I've ever met

DmitriiSukhomlinov:
It's still a draft

src/corelibs/U2Core/src/globals/Nullable.h
yalgaer: I'm happy that this utility class helped you to find bugs: unchecked access to nullable pointers.
Except this I see no other non-friendly usage patterns: the NP pointer behaves just like a regular pointer with a 0 boilerplate.

DmitriiSukhomlinov: The only thing this class helped me is to create bugs and spend time. But OK, I'll take it back

yalgaer: What kind of bugs did you make while using this class? 
This info may help to improve this class.

DmitriiSukhomlinov: The main problem is that it's very difficult to track either you used `operator==` (or `operator!=`) or not before some line. You need to check, in some cases, more than several hundred lines to to find out if some `NL<...> obj` has been  compared to something - and only after it you can decide if you can get it by overloaded `operator*` or by `getNullable()`. 

The second problem - you can use `operator*` accidentally instead of  `getNullable()`. Even when you added this class you had several places like this - [here](https://github.com/ugeneunipro/ugene/blob/db15b659827f6e76533d0fd33178ad7565f0b858/src/corelibs/U2Formats/src/BAMUtils.cpp#L374), [here](https://github.com/ugeneunipro/ugene/blob/db15b659827f6e76533d0fd33178ad7565f0b858/src/corelibs/U2Formats/src/BAMUtils.cpp#L567) and [here](https://github.com/ugeneunipro/ugene/blob/db15b659827f6e76533d0fd33178ad7565f0b858/src/corelibs/U2Formats/src/BAMUtils.cpp#L670). In all these spots you use a usual `operator*` instead of `getNullable()`, but the parameter haven't been used with `operator==` or `operator!=` yet, so it causes an unfair `SAFE_POINT`. It's difficult to produce a scenario to get these lines and our tests do not cover it, so it was missed and already released, which isn't good, but not critical.

What about me - I stumbled several times on that, and, as far as I have been using a **very** large file in this issue, it was a typical case, when I'd been waiting for 4-5 hours for calculation and then had a `SAFE_POINT` on some of this pointers.

DmitriiSukhomlinov:
I have found three problems:

1. Too many opened files at the same time. The file, which is opened [in this function](https://github.com/ugeneunipro/ugene/blob/db15b659827f6e76533d0fd33178ad7565f0b858/src/corelibs/U2Core/src/util/FileAndDirectoryUtils.cpp#L208) should be closed. If it's not, when approximately 500 files are opened, you'll get an error "EMFILE, too many open files" next time you try to open file. This problem was fixed by `fp->owned_file = 1;` where it's needed - now `bamFile` closes the file when dies.
2. When we have unsorted SAM/BAM file to open, UGENE splits it to several small file, sort each of them separately and then merge them all. `bamMergeCore` algorithm required ALL files opened at the same time. Our file was split to about 1500 files, so about 1500 files should be opened at the same time - problem number 1 again. I decided not to rewrite algorithm at all, but made it split files to bunches 100 files, merge them  - now UGENE has 15 files, each is the result of merging - and then merge these 15 files.
3. `ftello` and `fseeko` uses `int` to point the order number of the byte in the file. This value was too small for the large file I used, replaced with functions, which use `long`

DmitriiSukhomlinov:
@yalgaer please, review. No test, because it requires about a day to open a file we have - just check it manually (see issue description)

src/corelibs/U2Core/src/globals/Nullable.h
yalgaer: The scenario you described demonstrates problems in your code, but not with a class: you tried to use a nullable pointer with no null checks. This is exactly why this pointer exists: to reveal such problems during testing.

The pointer has 0 effect in the production code: it is designed to behave like a raw pointer. It should add a custom behavior only in debug or testing builds.

The only "problem" I know about NP class that it can't come up with results during compile time and requires test runtime.

The rule of thumb for this class: see NP in the method: make a null check. Tired of the null checks -> convert it to a raw pointer and use the raw pointer (but you have to make the null check during the conversion).

src/plugins/dbi_bam/src/SamtoolsBasedDbi.cpp
```diff
@@ -176,7 +176,7 @@ bamFile SamtoolsBasedDbi::openNewBamFileHandler() const {
     QString filePath = url.getURLString();
     NP<FILE> file = BAMUtils::openFile(filePath, "rb");
     bamFile bFile = bgzf_fdopen(file.getNullable(), "r");
-    CHECK_EXT(bFile != nullptr, BAMUtils::closeFileIfOpen(file), nullptr);
+    CHECK_EXT(bFile != nullptr, BAMUtils::closeFileIfOpen(file.getNullable()), nullptr);
```
yalgaer: nit: it is not recommended to use getNullable at all, NP should be used like a regular raw pointer.
In this case, you do it like `bamFile bamHandler = file != nullptr ? bgzf_fdopen(file, "r"): nullptr;` and use `file` directly everywhere below without `getNullable`

src/corelibs/U2Formats/src/tasks/BgzipTask.cpp
```diff
@@ -122,7 +122,12 @@ Task::ReportResult BgzipTask::report() {
 
 bool BgzipTask::checkBgzf(const GUrl& fileUrl) {
     NP<FILE> file = FileAndDirectoryUtils::openFile(fileUrl.getURLString(), "r");
-    int checkResult = file == nullptr ? -1 : bgzf_check_bgzf(file);
+    int checkResult = 0;
```
yalgaer: Here you replaced 1 line of code with 5. Does it change anything? If not, please keep the old code

src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -606,27 +631,40 @@ static int bamMergeCore(const QString& outFileName, const QList<QString>& filesT
...
+    }
 
-void* BAMUtils::loadIndex(const QString& filePath) {
-    // See bam_index_load_local.
-    QString mode = "rb";
-    NP<FILE> file = BAMUtils::openFile(filePath + ".bai", mode);
-    if (file == nullptr && filePath.endsWith("bam")) {
-        file = BAMUtils::openFile(filePath.chopped(4) + ".bai", mode);
+    QStringList newOutFileNameList;
+    for (int i = 0; i < iterationNumber; i++) {
```
yalgaer: A big part of the method here is a boilerplate that deals with chunks of files of size `MAX_FILES_OPENED`.
We already have a common pattern to deal with chunks with a minimal boilerplate. 
Could you please the manual chunk/remainder/iteration to `U2Region::split()` ?

src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -606,27 +631,40 @@ static int bamMergeCore(const QString& outFileName, const QList<QString>& filesT
...
+        newOutFileNameList << newOutFileName;
+        int res = recursiveBamMergeCore(newOutFileName, newFilesToMerge);
+        CHECK(res >= 0, res);
     }
-    CHECK(file != nullptr, nullptr);
-    bam_index_t* idx = bam_index_load_core(file);
-    closeFileIfOpen(file);
-    return idx;
+
+    return recursiveBamMergeCore(outFileName, newOutFileNameList);
```
yalgaer: Could you please document why we need the recursion here? Can it be done with no recursion if we use U2Region:split?

yalgaer (changes requested):
This is a new complex/useful functionality and we should not lose it by chance in the future.

Could you please add a small XML or GUI test for it with a manually created set of small BAM/SAMs that will work fast?

src/plugins/dbi_bam/src/SamtoolsBasedDbi.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Formats/src/tasks/BgzipTask.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: I used `U2Region:split`, its a good point I didn't know about, but recursion is still needed - just look at the code. Recursion is needed in case we split bam file and need 2 iterations to unite it - if the first split still produce a large amount of files, which could cause `EMFILE`

DmitriiSukhomlinov:
> This is a new complex/useful functionality and we should not lose it by chance in the future.
> 
> Could you please add a small XML or GUI test for it with a manually created set of small BAM/SAMs that will work fast?

The smallest file this problem is reproducible at is about 12.5 Gb - the cause of this bug is a file size and there is no way to make it smaller. I need time to compile this file and check how much time is needed to process it -but, I'm pretty sure, that at least 2-3 hours.

yalgaer:
> The smallest file this problem is reproducible at is about 12.5 Gb - the cause of this bug is a file size and there is no way to make it smaller. I need time to compile this file and check how much time is needed to process it -but, I'm pretty sure, that at least 2-3 hours.

The patch contains a new and important function (recursiveBamMergeCore) that has no tests. I expect that it is possible to create a test that will use the function you created and run in 2-3 seconds. This way we won't have an untested code in our codebase and the test will catch any unsafe changes in the future.

src/corelibs/U2Formats/src/BAMUtils.cpp
yalgaer: Thanks for using U2Region::split it made the code simpler and now I can see the logic of the method clearer

Question:
Once you did `auto mergeSplit = U2Region::split({ 0, size }, MAX_FILES_OPENED);`
the `newFilesToMerge` size is < MAX_FILES_OPENED and you can call `localBamMergeCore` instead of `recursiveBamMergeCore` right?

DmitriiSukhomlinov: Yes, you are right, it will work. I added the recursion here because of the following problem: if we have 1'000 files to merge, we split them to 10 bunches 100 files in each bunch. In the result, we have 10 files (each file is a result of merging of each bunch), then we merge these 10 files and everything is OK, it works.

But, if we have 100'000 files to merge, we split them into 1'000 bunches to merge (100 files in each bunch). So, in the result, we have 1'000 files (one file is the merge result from each bunch) and it's still to many to merge in one time, we need to split them again. This recursive algorithm prevent this scenario.

Actually, it may be redundantly, because the file which will be split to 100'000 should be **extremely** large and I've never met such a big data, but, just several weeks ago we've never met SAM file about 200Gb (which were sent us by Roman, who found this bug).

DmitriiSukhomlinov:
> > The smallest file this problem is reproducible at is about 12.5 Gb - the cause of this bug is a file size and there is no way to make it smaller. I need time to compile this file and check how much time is needed to process it -but, I'm pretty sure, that at least 2-3 hours.
> 
> The patch contains a new and important function (recursiveBamMergeCore) that has no tests. I expect that it is possible to create a test that will use the function you created and run in 2-3 seconds. This way we won't have an untested code in our codebase and the test will catch any unsafe changes in the future.

OK, I added a UNIT test

src/corelibs/U2Formats/src/BAMUtils.cpp
yalgaer: The line I mentioned
`int res = recursiveBamMergeCore(newOutFileName, newFilesToMerge)`
does not need the recursion call because `newFilesToMerge.size()` is always < MAX_FILES_OPENED, right?

BTW could you please add a comment which problem is being solved by the function. This functionality is not used by the original samtools right?

DmitriiSukhomlinov:
@yalgaer please approve, the only failed test has no relation to this issue, some random failure

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov: > BTW could you please add a comment which problem is being solved by the function. This functionality is not used by the original samtools right?

What function do you mean - `bamMergeCore` or `recursiveBamMergeCore`? Anyway, this functionality is not used by samtools, this functionality **uses** samtools.

yalgaer: I know what `bamMergeCore` does because I created it. I mean `recursiveBamMergeCore`. Could you please document the problem it solves and why the original samtools has no such problem?

I have 3 more questions about this function: 
1. Who cleans up the tmp files you created during recursion?  Could you please add a check to the test that no tmp files are left after the merge.
2. Why do you think that the first file location in a bucket is a good place for tmp files? 
3. About the name of the tmp merge file: please add a check that this name does not overlap with any of existing files - this name must be a unique name.

yalgaer (changes requested):
It looks like the recursive merge function created in this PR still have bugs

DmitriiSukhomlinov:
> It looks like the recursive merge function created in this PR still have bugs

Scenarios, please

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: > 1. Who cleans up the tmp files you created during recursion? Could you please add a check to the test that no tmp files are left after the merge.

 No one does because no one has been doing it before. All intermediate files, which are created during importing SAM/BAM to UGENEDB, are saved and I remember this behavior has had a place during the whole time I remember. I can change this behavior if you want, but please - not in this issue.

> 2. Why do you think that the first file location in a bucket is a good place for tmp files?

Don't get the point. What is "bucket"? What first file? What location?

> 3. About the name of the tmp merge file: please add a check that this name does not overlap with any of existing files - this name must be a unique name.

Let me do it in another issue, please. Current files are overwritten and this behavior has had a place during years.

DmitriiSukhomlinov:
Please, approve this PR - there are no known bugs here, there is a test which works good.

yalgaer:
> Please, approve this PR - there are no known bugs here, there is a test which works good.

I can't approve a PR that introduce bugs. Lets discuss each of the bugs separately in details. Maybe these are not new bugs (old bugs) in this case I will not ask you to fix them.

Bug with new files: 

Your PR adds a new function to merge N files into 1. 
The function creates intermediate files but never removes them. I consider it as a bug within the new function. There was no such behavior before. Please add a cleanup for all new files the function creates for its temporary needs.

yalgaer:
> Don't get the point. What is "bucket"? What first file? What location?

Bucket is a set of files processed at once as the result of the U2Region::split.
The tmp merge file dir is inherited from the first name in the bucket.
Question: why don't you use the TMP dir for tmp file?

src/corelibs/U2Formats/src/BAMUtils.cpp
yalgaer: Moving discussion into a separate thread: 1 per issue.

yalgaer:
> > About the name of the tmp merge file: please add a check that this name does not overlap with any of existing files - this name must be a unique name.
> Let me do it in another issue, please. Current files are overwritten and this behavior has had a place during years.

I'm not asking you to fix the old code that may (or may not) have this bug. I'm only asking not adding a new code with issues like this.
Please make your new code correct: check that the file name you generate does not overlap with any user file - so you overwrite it accidentally.

DmitriiSukhomlinov:
> > Please, approve this PR - there are no known bugs here, there is a test which works good.
> 
> I can't approve a PR that introduce bugs. Lets discuss each of the bugs separately in details. Maybe these are not new bugs (old bugs) in this case I will not ask you to fix them.
> 
> Bug with new files:
> 
> Your PR adds a new function to merge N files into 1. The function creates intermediate files but never removes them. I consider it as a bug within the new function. There was no such behavior before. Please add a cleanup for all new files the function creates for its temporary needs.

OK

DmitriiSukhomlinov:
> > Don't get the point. What is "bucket"? What first file? What location?
> 
> Bucket is a set of files processed at once as the result of the U2Region::split. The tmp merge file dir is inherited from the first name in the bucket. Question: why don't you use the TMP dir for tmp file?

When we import BAM/SAM file to UGENEDB, there are a lot of intermediate files are created - and them all are saved to the directory with the original file, not to the temporary directory. The reason I save these new intermediate files to the original file directory is that it would be better to have them all in the same directory, not to save part of them with the original file and another part to the temporary directory. 

The reason of keeping these files to the original file I do not know (it has been created far before me), but I can anticipate. The main problem here is the size of files. On Windows (80% UGENE users have Windows) the disk space is usually separated to two disks - C (System drive, the small one for OS and some other important stuff, which should be located near operating system) and D (anything else drive, the large one). The default temporary folder on Windows is located on C. If we save all this intermediate files to the Temporary folder (even if we change UGENE behavior and start removing them), they could fill system drive extremally (during the calculation process), which could cause a lot of problems for OS - it may start freeze, all work may slow down, etc. (it's recommended to have at least 50 Gb free space on system drive on Windows). I know, that we can change Temporary folder location, but, let be honest, very small amount of our users know that too.

The example. Roman (who sent us this 200 Gb file) had two drives - C (250 GB free space) and D (2+ Tb free space). His large file requires about 500 Gb free disk space just to be imported to UGENEDB. That means, that in all these temporary files would be stored to the C drive, this operation wouldn't be completed at all.

The result - it's better not to store to system drive files, which may have unpredictable large size.

DmitriiSukhomlinov:
> > > About the name of the tmp merge file: please add a check that this name does not overlap with any of existing files - this name must be a unique name.
> > > Let me do it in another issue, please. Current files are overwritten and this behavior has had a place during years.
> 
> I'm not asking you to fix the old code that may (or may not) have this bug. I'm only asking not adding a new code with issues like this. Please make your new code correct: check that the file name you generate does not overlap with any user file - so you overwrite it accidentally.

OK

src/plugins/api_tests/src/core/util/BAMUtilsUnitTests.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <U2Formats/BAMUtils.h>
+
+namespace U2 {
+
+IMPLEMENT_TEST(BAMUtilsUnitTests, bamMergeCore) {
+    TestRunnerSettings* trs = AppContext::getAppSettings()->getTestRunnerSettings();
+    QString inputFilesDir = trs->getVar("COMMON_DATA_DIR") + "/regression/7862/";
+
+    QStringList bamUrls;
+    for (int i = 0; i < 550; i++) {
```
yalgaer: On Linux the default ulimit for files is 1000, so the test does not check the issue: 550 will work even if the new function does nothing (delegates to the samtools one). Is it possible to increase the number to 1000? WIll it be slow (>10 seconds?)

src/plugins/api_tests/src/core/util/BAMUtilsUnitTests.cpp
```diff
@@ -0,0 +1,58 @@
...
+        bamUrls << inputFilesDir + QString::number(i + 1) + ".bam";
```
yalgaer: Do I understand correctly that you added 500+ identical files into SVN repo? If yes, please replace with 1 file and copy it N times in the test.

yalgaer:
> On Windows (80% UGENE users have Windows) the disk space is usually separated to two disks - C

I do not suggest using system TMP dir: in UGENE we have a dedicated configurable temp dir (see settings) used exactly for this by other tasks.

If you really care about the disk space overflow you can solve it correctly: check the available disk space at destination before the task is started: BAM/SAM size is predictable.

Even if you do not know about UGENE temp dir concept: why did you selected the source folder for the tmp files but not the destination one? When user selects a destination dir he cares about write permissions and space.

src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -606,27 +631,48 @@ static int bamMergeCore(const QString& outFileName, const QList<QString>& filesT
...
+    QStringList newOutFileNameList;
+    for (int i = 0; i < mergeSplit.size(); i++) {
+        const auto& currentRange = mergeSplit.at(i);
+         QList<QString> newFilesToMerge = filesToMerge.mid(currentRange.startPos, currentRange.endPos());
+        QString newOutFileName = newFilesToMerge.first();
+        do {
+            // Remove ".bam" from the end
+            newOutFileName = newOutFileName.left(newOutFileName.size() - 4);
+            newOutFileName = newOutFileName + "_" + QString::number(i) + ".bam";
+        } while (QFile(newOutFileName).exists());
```
yalgaer: nit: access to the FS is not free. A better & simpler 1 liner solution is possible here: append a randomly generated uuid() to the original file name.

src/corelibs/U2Formats/src/BAMUtils.cpp
```diff
@@ -606,27 +631,48 @@ static int bamMergeCore(const QString& outFileName, const QList<QString>& filesT
...
-void* BAMUtils::loadIndex(const QString& filePath) {
-    // See bam_index_load_local.
-    QString mode = "rb";
-    NP<FILE> file = BAMUtils::openFile(filePath + ".bai", mode);
-    if (file == nullptr && filePath.endsWith("bam")) {
-        file = BAMUtils::openFile(filePath.chopped(4) + ".bai", mode);
+int BAMUtils::bamMergeCore(const QString& outFileName, const QList<QString>& filesToMerge) {
+    coreLog.trace("bamMergeCore: " + filesToMerge.join(",") + " to " + outFileName);
+
+    QStringList intermediateFiles;
```
yalgaer: Intermediate files should be cleaned up right after they merged. In this case the disk space overhead will be <=1x (where x is the size of the result file which is approximately equal to the total size of all input files or a total size of the merged files per recursive iteration).

When you clean them up at the end the overhead is (N-1)x where N is a depth of the recursion. So if a user has 200Gb file you may end up with 10x overhead: 2Tb instead of 200Gb.

yalgaer (changes requested):

src/plugins/api_tests/src/core/util/BAMUtilsUnitTests.cpp
DmitriiSukhomlinov: Good idea, OK

src/plugins/api_tests/src/core/util/BAMUtilsUnitTests.cpp
DmitriiSukhomlinov: It still will be fast, done

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Formats/src/BAMUtils.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
> > On Windows (80% UGENE users have Windows) the disk space is usually separated to two disks - C
> 
> I do not suggest using system TMP dir: in UGENE we have a dedicated configurable temp dir (see settings) used exactly for this by other tasks.
> 
> If you really care about the disk space overflow you can solve it correctly: check the available disk space at destination before the task is started: BAM/SAM size is predictable.
> 
> Even if you do not know about UGENE temp dir concept: why did you selected the source folder for the tmp files but not the destination one? When user selects a destination dir he cares about write permissions and space.

You have described some complicated stuff I haven't completely understood. Let me say like this - I save this files to the same directory all other intermediate files are saved to (only UGENEDB is saved to the "destination directory"). I fixed all your comments, if you want me to fix something else - please, just tell me, what exactly you want.

yalgaer:
> You have described some complicated stuff I haven't completely understood. Let me say like this - I save this files to the same directory all other intermediate files are saved to (only UGENEDB is saved to the "destination directory"). I fixed all your comments, if you want me to fix something else - please, just tell me, what exactly you want.

Please save the intermediate files into the destination or UGENE temp directory, run all the tests and I will approve.

yalgaer (changes requested):
> You have described some complicated stuff I haven't completely understood. Let me say like this - I save this files to the same directory all other intermediate files are saved to (only UGENEDB is saved to the "destination directory"). I fixed all your comments, if you want me to fix something else - please, just tell me, what exactly you want.

Please save the intermediate files into the destination or UGENE temp directory, run all the tests and I will approve.

yalgaer (approved):

PR #1224, opened 2023-04-20
UGENE-7837 Add "commercial sources" to the "Restriction sites" dialog and make possible to filter by them

DmitriiSukhomlinov:


src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,148 @@
...
+    return checkedItems;
+}
+
+void ComboBoxWithCheckBoxes::setCheckedItems(const QStringList &items) {
+    QStandardItemModel *standartModel = qobject_cast<QStandardItemModel *>(model());
+    SAFE_POINT(standartModel != nullptr, L10N::nullPointerError("QStandardItemModel"), );
+
+    disconnect(standartModel, &QStandardItemModel::itemChanged, this, &ComboBoxWithCheckBoxes::sl_modelItemChanged);
+
+    for (int i = 0; i < count(); ++i) {
```
yalgaer: Here you should swap states only for the changed items.

Reason: external users of model() will receive extra 'unchecked'->'checked' update for the items that stays 'checked' with your current code.

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,148 @@
...
+        int index = findText(items.at(i));
+
+        if (index != -1) {
+            standartModel->item(index)->setData(Qt::Checked, Qt::CheckStateRole);
+        }
+    }
+
+    connect(standartModel, &QStandardItemModel::itemChanged, this, &ComboBoxWithCheckBoxes::sl_modelItemChanged);
+
+    collectCheckedItems();
```
yalgaer: this method is badly named. It does 'repaint' and emits a signal that state is changed.
Suggested name: `updateOnCheckedItemsChange()`

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,148 @@
...
+    QStyleOptionComboBox option;
+    initStyleOption(&option);
+    painter.drawComplexControl(QStyle::CC_ComboBox, option);
+
+    QRect textRect = rect().adjusted(displayRectDelta.left(), displayRectDelta.top(), displayRectDelta.right(), displayRectDelta.bottom());
+    painter.drawText(textRect, Qt::AlignVCenter, displayText);
+}
+
+void ComboBoxWithCheckBoxes::resizeEvent(QResizeEvent *) {
+    updateDisplayText();
```
yalgaer: q: send resize to the parent too?

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,148 @@
...
+        SAFE_POINT(currentItem != nullptr, L10N::nullPointerError("QStandardItem"), );
+
+        Qt::CheckState checkState = static_cast<Qt::CheckState>(currentItem->data(Qt::CheckStateRole).toInt());
+        if (checkState == Qt::Checked) {
+            checkedItems.push_back(currentItem->text());
+        }
+    }
+
+    updateDisplayText();
+    repaint();
```
yalgaer: q: Do you really need a repaint() ? Won't 'update' be enough?

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -217,22 +220,32 @@ QList<SEnzymeData> EnzymesIO::readBairochFile(const QString& url, IOAdapterFacto
...
             }
+        } else if (buff[0] == 'C' && buff[1] == 'R') {
+            auto suppliersAbbreviations = QByteArray(buff + 3, len - 3).trimmed();
+            for (const auto& abb : qAsConst(suppliersAbbreviations)) {
+                CHECK_CONTINUE(abb != '.');
+
+                currentData->suppliers << commercialSources.value(abb);
+            }
+            if (currentData->suppliers.isEmpty()) {
+                currentData->suppliers << tr("Not defined");
```
yalgaer: Sort the suppliers by name before showing them in the long list. Otherwise it is hard to navigate it. The sort can be done either here or closer to UI

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -133,25 +136,27 @@ EnzymeFileFormat EnzymesIO::detectFileFormat(const QString& url) {
...
+    int len = 0;
     bool lineOk = true;
-    while ((len = io->readUntil(buff, DocumentFormat::READ_BUFF_SIZE, LINE_BREAKS, IOAdapter::Term_Include, &lineOk)) > 0 && !ti.cancelFlag) {
+    QMap<char, QString> commercialSources;
+    while ((len = io->readUntil(buff, DocumentFormat::READ_BUFF_SIZE, LINE_BREAKS, IOAdapter::Term_Include, &lineOk)) > 0 && !os.isCanceled()) {
         line++;
         if (!lineOk) {
-            ti.setError(tr("Line is too long: %1").arg(line));
+            os.setError(tr("Line is too long: %1").arg(line));
             return res;
```
yalgaer, line 164: You can't use return here while 'io' is opened.

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.h
```diff
@@ -0,0 +1,62 @@
...
+#include <QComboBox>
+#include <QStandardItem>
+
+namespace U2 {
+
+//https://habr.com/ru/post/215289/
+class ComboBoxWithCheckBoxes : public QComboBox {
+    Q_OBJECT
+
+    Q_PROPERTY(QStringList checkedItems READ getCheckedItems WRITE setCheckedItems)
```
yalgaer: Any reason to have this line?

src/plugins/enzymes/src/EnzymesLoader.h
```diff
@@ -0,0 +1,51 @@
...
+private:
+    static void calculateSuppliers();
+
+    static QList<SEnzymeData> loadedEnzymes;
+    static QStringList loadedSuppliers;
+};
+
+
+
+}
```
yalgaer: add a new line

src/plugins/enzymes/src/EnzymesLoader.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <QFileInfo>
+
+namespace U2 {
+
+QList<SEnzymeData> EnzymesLoader::loadedEnzymes;
+QStringList EnzymesLoader::loadedSuppliers;
+
+void EnzymesLoader::setupEnzymes() {
+    QString lastUsedFile = AppContext::getSettings()->getValue(EnzymeSettings::DATA_FILE_KEY).toString();
```
yalgaer: You should also store a selected supplier list and reset it to all only when a new rebase file is selected.
Selecting/de-selecting 10+ suppliers on every view open may be frustrating

src/plugins/enzymes/src/EnzymesLoader.h
```diff
@@ -0,0 +1,51 @@
...
+
+class  EnzymesLoader {
+    Q_DECLARE_TR_FUNCTIONS(EnzymesLoader)
+public:
+    EnzymesLoader() = delete;
+
+    static void setupEnzymes();
+    static const QList<SEnzymeData>& loadEnzymes(const QString& url);
+
+    static const QList<SEnzymeData>& getLoadedEnzymes();
```
yalgaer: Document public methods you change.

src/plugins/enzymes/src/EnzymesPlugin.cpp
```diff
@@ -74,6 +75,7 @@ EnzymesPlugin::EnzymesPlugin()
     }
 
     EnzymesSelectorWidget::setupSettings();
+    EnzymesLoader::setupEnzymes();
```
yalgaer: You can't do it here: it affects UGENE startup time.

If it is fast: do it in the dialog or on the first access. 
If it is slow - there is an extra reason not doing it during startup.

If you do not like that it blocks the dialog: show a loading message before the dialog.

yalgaer (changes requested):

src/plugins/enzymes/src/EnzymesPlugin.cpp
DmitriiSukhomlinov: I didn't get you logic at all and I am not about to change it. As far as our restriction base grown up, its loading became a bit long - it freezes "Restriction Enzyme" dialog if it loads on the first opening. The best idea is to move it to UGENE loading - as, for example, we validate External Tools or load plugins.

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: Yes, `updateDisplayText()` just calculates new text to be shown, `repaint()` calls `paintEvent(QPaintEvent*)`, which redraws widget and applies this text

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: OK, I added it to `EnzymesLoader::calculateSuppliers()`

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: OK, deleter has been added

src/plugins/enzymes/src/EnzymesLoader.h
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.h
DmitriiSukhomlinov: This is some general realization "how it should be", in our case it's not needed, removed

src/plugins/enzymes/src/EnzymesLoader.cpp
DmitriiSukhomlinov: Don't get the point. And what is "new rebase file"?

src/plugins/enzymes/src/EnzymesLoader.h
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesPlugin.cpp
yalgaer: You can't compare enzymes file loading on startup & external tools directly. 
Here are the reasons:

1. I invested a lot of work reducing external tools validation time from 10+ seconds to 2-3 seconds and have more plans to reduce it even event more (make it sub-second). This work shows the importance of the problem: UGENE startup must be fast. Startup time is a very important metric.

2. Even current 2-3 seconds needed for ET validation is a 1 time job per version: only the initial run is slow. Your patch makes every run slow.

3. I'm glad you started to improve the restriction sites plugin, but the lazy (delayed) loading was not a miss in the original design but an intentional optimization.

Unfortunately I can't approve the patch (which has a real value) because the patch tries to  fix the unrelated and undiscussed part of the code in the incorrect way.

yalgaer (changes requested):
Please restore the lazy loading of the enzymes and remove it from the UGENE startup

src/plugins/GUITestBase/src/runnables/ugene/plugins/enzymes/GTComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,164 @@
...
+                case GTGlobals::UseKey:
+                case GTGlobals::UseKeyBoard:
+                {
+                    if (state != expectedState) {
+                        GTKeyboardDriver::keyClick(Qt::Key_Space);
+                        GTThread::waitForMainThread();
+                    }
+
+                    GTKeyboardDriver::keyClick(Qt::Key_Down);
+                    GTThread::waitForMainThread();
```
yalgaer: this line is not needed: `GTKeyboardDriver::keyClic` already waits for the main thread by default.

src/plugins/GUITestBase/src/runnables/ugene/plugins/enzymes/GTComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,164 @@
...
+    // Access to the internal comboBox->view() must be done from the main thread (the view is lazily instantiated).
+    class MainThreadAction : public CustomScenario {
+    public:
+        MainThreadAction(QComboBox* comboBox, const QList<int>& indexes, GTGlobals::UseMethod method)
+            : CustomScenario(), comboBox(comboBox), indexes(indexes), method(method) {
+        }
+
+        void run(HI::GUITestOpStatus& os) override {
+            GTWidget::click(os, comboBox);
+            GTThread::waitForMainThread();
```
yalgaer: Mouse click  already waits for the main thread.

src/plugins/GUITestBase/src/runnables/ugene/plugins/enzymes/GTComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+#define GT_METHOD_NAME "selectItemByIndex"
+void GTComboBoxWithCheckBoxes::selectItemByIndex(GUITestOpStatus& os, QComboBox* comboBox, const QList<int>& indexes, GTGlobals::UseMethod method) {
+    GT_CHECK(comboBox != nullptr, "QComboBox* == NULL");
+
+    // Access to the internal comboBox->view() must be done from the main thread (the view is lazily instantiated).
+    class MainThreadAction : public CustomScenario {
+    public:
+        MainThreadAction(QComboBox* comboBox, const QList<int>& indexes, GTGlobals::UseMethod method)
+            : CustomScenario(), comboBox(comboBox), indexes(indexes), method(method) {
```
yalgaer: nit: do not call the default super constructor explicitly: it is a meaningless code

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,155 @@
+/**
```
yalgaer: This ui component should go into U2Gui and available to all users

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
```diff
@@ -0,0 +1,155 @@
...
+    QList<int> checkedIndexes;
+    for (int i = 0; i < items.count(); ++i) {
+        checkedIndexes << findText(items.at(i));
+    }
+
+    for (int i = 0; i < count(); ++i) {
+        QStandardItem* currentItem = standartModel->item(i);
+        SAFE_POINT(currentItem != nullptr, L10N::nullPointerError("QStandardItem"), );
+
+        Qt::CheckState newState = Qt::Unchecked;
```
yalgaer: nit: can be a one-liner with ? operator

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
yalgaer: `repaint` forces an immediate repaint. My question was if the `update` (deferred repaint) is OK here

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -133,25 +136,30 @@ EnzymeFileFormat EnzymesIO::detectFileFormat(const QString& url) {
...
 }
 
-QList<SEnzymeData> EnzymesIO::readBairochFile(const QString& url, IOAdapterFactory* iof, TaskStateInfo& ti) {
+QList<SEnzymeData> EnzymesIO::readBairochFile(const QString& url, IOAdapterFactory* iof, U2OpStatus& os) {
     QList<SEnzymeData> res;
 
-    QScopedPointer<IOAdapter> io(iof->createIOAdapter());
+    QSharedPointer<IOAdapter> io(iof->createIOAdapter(), [](IOAdapter* ioAdapter) {
+        ioAdapter->close();
+        ioAdapter->deleteLater();
```
yalgaer: Could you please add a comment why the default `delete` behavior is not enough?

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -133,25 +136,30 @@ EnzymeFileFormat EnzymesIO::detectFileFormat(const QString& url) {
...
     SEnzymeData currentData(new EnzymeData());
     QByteArray buffArr(DocumentFormat::READ_BUFF_SIZE, 0);
     char* buff = buffArr.data();
     const QBitArray& LINE_BREAKS = TextUtils::LINE_BREAKS;
-    int line = 0, len = 0;
+    int line = 0;
+    int len = 0;
     bool lineOk = true;
-    while ((len = io->readUntil(buff, DocumentFormat::READ_BUFF_SIZE, LINE_BREAKS, IOAdapter::Term_Include, &lineOk)) > 0 && !ti.cancelFlag) {
+    QMap<char, QString> commercialSources;
```
yalgaer: nit: use QHash for keys with an identity (numbers)

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -166,30 +174,28 @@ QList<SEnzymeData> EnzymesIO::readBairochFile(const QString& url, IOAdapterFacto
             }
             continue;
         }
-        if (len < 4) {
-            continue;
-        }
-        if (buff[0] == 'I' && buff[1] == 'D') {
+        CHECK_CONTINUE(len >= 4);
+
+        if (buff[0] == 'C' && buff[1] == 'C' && buff[6] == '=') {
```
yalgaer: Access to `buff[6]` is unsafe here

src/plugins/enzymes/src/EnzymesIO.cpp
yalgaer: Is the `Not defined` still the first after the new sort algorithm? 'Not default' (or 'Public sources') is a not a supplier but a different domain and I would suggest having it the first item in the list like you had before.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -128,41 +131,65 @@ QList<SEnzymeData> EnzymesSelectorWidget::getSelectedEnzymes() {
...
+        if (!QFileInfo(url).exists()) {
+            os.setError(tr("File not exists: %1").arg(url));
         } else {
-            ioLog.error(ti.getError());
+            GTIMER(c1, t1, "FindEnzymesDialog::loadFile [EnzymesIO::readEnzymes]");
+            enzymes = EnzymesIO::readEnzymes(url, os);
         }
-        return;
+        if (os.hasError()) {
+            if (isVisible()) {
```
yalgaer, line 181: q: why the widget can be invisible here?

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -195,6 +222,10 @@ void EnzymesSelectorWidget::saveFile(const QString& url) {
     }
     if (QMessageBox::question(this, tr("New enzymes database has been saved."), tr("Do you want to work with new database?"), QMessageBox::Yes | QMessageBox::No) == QMessageBox::Yes) {
         loadFile(url);
+        if (!loadedEnzymes.isEmpty()) {
```
yalgaer: These 5 lines of code are duplicated (from loadFile to emit a signal). Move them into a dedicated function.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +635,57 @@ void FindEnzymesDialog::accept() {
...
+    const auto& loadedSuppliers = EnzymesSelectorWidget::getLoadedSuppliers();
+    cbSuppliers->clear();
+    for (const auto& supplier : qAsConst(loadedSuppliers)) {
+        cbSuppliers->addItem(supplier);
+    }
+    QString selStr = AppContext::getSettings()->getValue(EnzymeSettings::CHECKED_SUPPLIERS).toString();
+    if (selStr.isEmpty()) {
+        cbSuppliers->setCheckedItems(loadedSuppliers);
+    } else {
+        cbSuppliers->setCheckedItems(selStr.split(SUPPLIERS_LIST_SEPARATOR));
```
yalgaer: You should either avoid translation of default ("Public sources," "Not defined") supplier or save it separately. Otherwise if a user switches language the setting is lost.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +635,57 @@ void FindEnzymesDialog::accept() {
...
+    }
+    enzSel->setEnzymesList(visibleEnzymes);
+    AppContext::getSettings()->setValue(EnzymeSettings::CHECKED_SUPPLIERS, checkedSuppliers.join(SUPPLIERS_LIST_SEPARATOR));
+}
+
+void FindEnzymesDialog::sl_updateSuppliers() {
+    const auto& loadedSuppliers = EnzymesSelectorWidget::getLoadedSuppliers();
+    cbSuppliers->clear();
+    for (const auto& supplier : qAsConst(loadedSuppliers)) {
+        cbSuppliers->addItem(supplier);
```
yalgaer: Use `addItems` to add all suppliers in 1 line and have only 1 set of change signals as the result.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +635,57 @@ void FindEnzymesDialog::accept() {
...
+    }
+    QString selStr = AppContext::getSettings()->getValue(EnzymeSettings::CHECKED_SUPPLIERS).toString();
+    if (selStr.isEmpty()) {
```
yalgaer: nit: 5 lines can be replaces with 1 using ? operator

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +635,57 @@ void FindEnzymesDialog::accept() {
     QDialog::accept();
 }
 
+void FindEnzymesDialog::sl_setEnzymesBySuppliers(QStringList checkedSuppliers) {
```
yalgaer: nit: you do more than update enzymes in this method. It is better to rename it to something like `sl_handleSupplierSelectionChange`

yalgaer (changes requested):
Some comments are 'nit' but some are real bugs. Please update.

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: Yes

src/plugins/GUITestBase/src/runnables/ugene/plugins/enzymes/GTComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/runnables/ugene/plugins/enzymes/GTComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
yalgaer: Please change to `update` in this case.

src/plugins/enzymes/src/EnzymesLoader.cpp
yalgaer: rebase is a name of the database. "rebase file" is a file with enzymes in rebase format.

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesLoader.cpp
DmitriiSukhomlinov: I still don't get the point. Please, provide scenario: 1, 2, 3, Expected: blah-blah-blah Current: blah-blah-blah

src/plugins/enzymes/src/EnzymesPlugin.cpp
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/runnables/ugene/plugins/enzymes/GTComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesLoader.cpp
yalgaer: After my initial comment you already added a saving of selected suppliers. If you have no other questions you can consider this issue resolved.

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/ComboBoxWithCheckBoxes.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: Well, as far as I can see, this is as obvious as possible, I don't even know what comment I can have here, but I'll try to

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: I don't like `QHash`, for some mysterious reasons MSVS debugger can't show its values, but with `QMap` debugger works perfectly. They have no differences, which would be noticeable in our work.

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: That's a good one, OK

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: I dunno, it was [before me](https://github.com/ugeneunipro/ugene/pull/1224/files#diff-f9b76a626092e365fb3bca9cdee014eff336a326e00ae833ee77a3ab982ef542L151)

src/plugins/enzymes/src/EnzymesIO.cpp
yalgaer: Originally it was a regular delete by a shared pointer. You changed it to a non-trivial delete. Could you please explain the change? The best place to explain is to add a comment.

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: OK, I will check it later. Resolved.

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/EnzymesLoader.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: OK

yalgaer: Your comment is wrong: destructors of all IOAdapters close the resource automatically: you do not need to do it manually.
My original question was why do you use `deleteLater` instead of the simple `delete`?

yalgaer (changes requested):
Please check the thread about io-adapter delete call: the comment & new code you wrote is wrong (not needed)

q: are the problems with "Not defined" supplier sorting (mus be first in the list) & storing in settings in localized form fixed and ready for testing?

DmitriiSukhomlinov:
>Please check the thread about io-adapter delete call: the comment & new code you wrote is wrong (not needed)

See comment above

> are the problems with "Not defined" supplier sorting (mus be first in the list) & storing in settings in localized form fixed and ready for testing?

Yes, I fixed it

src/plugins/enzymes/src/EnzymesIO.cpp
yalgaer: ok

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +640,55 @@ void FindEnzymesDialog::accept() {
...
+            CHECK_CONTINUE(checkedSuppliers.contains(supplier));
+
+            visibleEnzymes.append(enzyme);
+            break;
+
+        }
+    }
+    enzSel->setEnzymesList(visibleEnzymes);
+    static const QString notDefinedTr = EnzymesIO::tr(EnzymesIO::NOT_DEFINED_SIGN);
+    auto value = checkedSuppliers.join(SUPPLIERS_LIST_SEPARATOR).replace(notDefinedTr, EnzymesIO::NOT_DEFINED_SIGN);
```
yalgaer: This way you will replace any supplier with "Not defined" (a valid English phrase) inside. This is a bug. Make the replace before the join using a complete name match.

yalgaer (changes requested):
I see 2 bugs left:
1) Unsafe replace in the string of the joined suppliers (see my comment in-place).

2) Please remove the Deleter you added to the shared pointer of ioAdapter
All IOAdapters already cleanup (close) owned resources when deleted. + nit: QScopedPointer is more semantically correct in the context of this function.

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
```diff
@@ -52,12 +52,14 @@ class DNAAlphabet;
 
 #define ENZYME_CUT_UNKNOWN 0x7FFFFF
 #define ENZYME_LIST_SEPARATOR ","
+#define SUPPLIERS_LIST_SEPARATOR ";"
```
yalgaer: q: can `;` be a part of a supplier name? If yes, change the separator to `\n`

DmitriiSukhomlinov:
> 1. Unsafe replace in the string of the joined suppliers (see my comment in-place).

Fixed

> 2. Please remove the Deleter you added to the shared pointer of ioAdapter
>    All IOAdapters already cleanup (close) owned resources when deleted. + nit: QScopedPointer is more semantically correct in the context of this function.

OK, I took the original code back, but I don't quite get what the original comment was about

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +640,58 @@ void FindEnzymesDialog::accept() {
...
+    AppContext::getSettings()->setValue(EnzymeSettings::CHECKED_SUPPLIERS, value);
+}
+
+void FindEnzymesDialog::sl_updateSuppliers() {
+    const auto& loadedSuppliers = EnzymesSelectorWidget::getLoadedSuppliers();
+    cbSuppliers->clear();
+    cbSuppliers->addItems(loadedSuppliers);
+    QString selStr = AppContext::getSettings()->getValue(EnzymeSettings::CHECKED_SUPPLIERS).toString();
+    static const QString notDefinedTr = EnzymesIO::tr(EnzymesIO::NOT_DEFINED_SIGN);
+    selStr = selStr.replace(EnzymesIO::NOT_DEFINED_SIGN, notDefinedTr);
```
yalgaer, line 670: Replace the translated value after the split.
This is the last improvement needed. Once test pass I will approve.

yalgaer (changes requested):
Replace the translated value after the split (line 670).
This is the last improvement needed. Once test pass I will approve.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -589,6 +640,58 @@ void FindEnzymesDialog::accept() {
...
+}
+
+void FindEnzymesDialog::sl_updateSuppliers() {
+    const auto& loadedSuppliers = EnzymesSelectorWidget::getLoadedSuppliers();
+    cbSuppliers->clear();
+    cbSuppliers->addItems(loadedSuppliers);
+    QString selStr = AppContext::getSettings()->getValue(EnzymeSettings::CHECKED_SUPPLIERS).toString();
+    static const QString notDefinedTr = EnzymesIO::tr(EnzymesIO::NOT_DEFINED_SIGN);
+    cbSuppliers->setCheckedItems(selStr.isEmpty() ? loadedSuppliers : selStr.split(SUPPLIERS_LIST_SEPARATOR));
+    selStr = selStr.replace(EnzymesIO::NOT_DEFINED_SIGN, notDefinedTr);
```
yalgaer: This code looks wrong (a typo?)
Here what you should do here:
1) Get list of the original suppliers from (loaded suppliers)
2) Replace the "NOT_DEFINED" with a translated value in the list. Set it to the UI.
3) Get CHECKED_SUPPLIERS from the settings: these are untranslated values. Split them into tokens. Replace the NOT_DEFINED token with a translated value
4) Call setCheckedItem().

yalgaer (changes requested):
See comment about "not defined" list & selStr

yalgaer:
Looks good. Will approve after tests are finished.

yalgaer (approved):

PR #1265, opened 2023-05-18
UGENE-7884 Translate Option panel for Statistics page (Russian locale)

smlnkcv:


yalgaer (changes requested):
With "U2::ClassName" prefx the QT_TR_NOOP solution introduced by Dmitry earlier this year went too far as a complex and non-standard way of how translations are used in UGENE.

The correct solution is the following: 

replace all translations like
```
static const char* CAPTION_SEQ_REGION_LENGTH = QT_TRANSLATE_NOOP("U2::SequenceInfo", "Length");
```
 with a private field initialized in the  SequenceInfo constructor
```
sequenceRegionLengthCaption = tr("Length");
```

and use this field everywhere in SequenceInfo as is.

The requested work can be done in 10 minutes.

yalgaer (dismissed):

yalgaer:
Please run at least Linux GUI tests before merging.
As a merging point always use master. I will cherry pick into the release manually.

yalgaer (approved):

yalgaer (approved):

PR #1266, opened 2023-05-18
UGENE-7885 Out of memory when cut or copy alignment

DmitriiSukhomlinov:


yalgaer (changes requested):
The solution is not correct: 

1) It keeps the data allocated inside try-catch leaked (you do not even know the place where OOM error happens)
2) It is not aligned with OOM handling in other parts of UGENE.

The best solution here is not to catch but  prevent OOM error completely. We need to catch only when the wrapped code is out of control.

A correct solution is to use a constant: see MAX_SAFE_COPY_TO_CLIPBOARD_SIZE and fail before copying.

Once you use it I think it won't be possible to get this error anymore.
If the error is still reproducible with MAX_SAFE_COPY_TO_CLIPBOARD_SIZE I'm OK to keep the catch or consider lowering MAX_SAFE_COPY_TO_CLIPBOARD_SIZE constant.

About the return value: create a new method and pass `os` as a parameter. Call it from sl_copySelection() (slot) and handle the error.

DmitriiSukhomlinov:
@yalgaer all tests are passed except the one which fail on all branches (no relation to this issue). Please, check and approve

yalgaer (approved):
Path is good. 
To speed up approvals in case of failed tests please add a comment that you checked the failed tests and the failures are not related to the patch.

PR #1271, opened 2023-05-26
UGENE-7835 Load some information about sequence when it's loaded from DB

DmitriiSukhomlinov:
Read issue comments for details


src/corelibs/U2Core/src/tasks/LoadRemoteDocumentTask.cpp
```diff
@@ -209,6 +233,21 @@ void LoadRemoteDocumentTask::prepare() {
...
+    CHECK(!hasError(), tr("Failed to download %1 from %2. Error: %3").arg(accNumber).arg(dbName).arg(getError()));
+    CHECK(!isCanceled(), {});
+
+    QString res = tr("Document was successfully downloaded: [%1, %2] -> <a href='%3'>%3</a>").arg(dbName).arg(accNumber).arg(fullPath);
+    QString pageLink = RemoteDBRegistry::PAGE_LINKS.value(dbName);
+    SAFE_POINT(!pageLink.isEmpty(), QString("No database found: %1").arg(dbName), res);
+
+    pageLink = pageLink.arg(accNumber);
+    res += "<br>";
+    res += tr("The database page link: <a href='%1'>%1</a>").arg(pageLink);
```
yalgaer: Nit: "the database" looks bad (in english) here, because there is no context of the database in the notification text
Change to "External database link: %1"...

src/corelibs/U2Core/src/tasks/LoadRemoteDocumentTask.h
```diff
@@ -54,17 +55,19 @@ class U2CORE_EXPORT RemoteDBRegistry {
     QMap<QString, QString> httpDBs;
     QMap<QString, QString> hints;
     QMap<QString, QString> aliases;
-    QMap<QString, QString> externalLinks;
 
 public:
+    // Links to the database (to open in browser) by the database name.
+    static const QMap<QString, QString> EXTERNAL_LINKS;
+    // Links to the database (to open in browser) to open a sequence page by ID.
```
yalgaer: "by ID" -> "by database name"

src/corelibs/U2Core/src/tasks/LoadRemoteDocumentTask.h
```diff
@@ -42,6 +42,7 @@ namespace U2 {
 #define FASTA_FORMAT "fasta"
 #define GENBANK_WITH_PARTS "gbwithparts"
 #define FORCE_DOWNLOAD_SEQUENCE_HINT "gbwithparts"
+#define SHOW_REPORT_HINT "show_report_hint"
```
yalgaer: You do not need this hint on the "LoadRemoteDocument" task level
Reasons:
1) LoadRemoteDocument does not know about UI reports. So it can't show anything.
2) All this flag does it sets "true" flag to setReportingEnabled/setReportingSupported

So instead of the new entity "show_report_hint"  do the following
1) Implement generateReport method (as you did)
2) Set reporting supported to true by default:  setReportingSupported(true);
3) Enable reporting by the parent UI task when needed:  setReportingEnabled(true);

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
```diff
@@ -1607,9 +1609,14 @@ void AnnotationsTreeView::sl_itemClicked(QTreeWidgetItem* i, int column) {
     if (lastMB != Qt::LeftButton || item == nullptr || !item->isColumnLinked(column)) {
         return;
     }
+
     QString fileUrl = item->getFileUrl(column);
     if (!fileUrl.isEmpty()) {
-        Task* task = new LoadRemoteDocumentAndAddToProjectTask(fileUrl);
```
yalgaer, line 1612: There was a contract in the AnnotationsTreeView: if an item contains an `fileUrl` it points to the downloadable resource.
It is does not matter where this URL comes from: a DbXRef subsystem or some other source.

With this change you break this abstraction and force the URL to come from DbXref only.
The original behavior should be preserved: if you have an URL the url should be downloaded and opened as is.

I see that in this case you may loose the relation of the URL to the DB and do not know how to generate the report (the goal of the task).

This can be done differently: build extra information about the URL: 
1) either by parsing & matching the URL inside of the report generator 
or
2) by passing additional report metadata to the loader task

src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/GetSequenceByIdFiller.h
```diff
@@ -0,0 +1,40 @@
...
+#include "utils/GTUtilsDialog.h"
+
+namespace U2 {
+using namespace HI;
+
+class GetSequenceByIdFiller : public Filler {
+public:
+    GetSequenceByIdFiller(HI::GUITestOpStatus& _os, const QString& saveTo, bool addToProject);
+
+    void commonScenario();
```
yalgaer: nit: add override

src/corelibs/U2Gui/src/util/DownloadRemoteFileDialog.cpp
```diff
@@ -207,21 +208,17 @@ void DownloadRemoteFileDialog::accept() {
...
         }
     }
-    bool hasLoadOnlyDocuments = false;
-    foreach (const QString& resId, resIds) {
-        LoadRemoteDocumentMode mode = LoadRemoteDocumentMode_LoadOnly;
-        if (addToProject) {
-            mode = taskCount < OpenViewTask::MAX_DOC_NUMBER_TO_OPEN_VIEWS ? LoadRemoteDocumentMode_OpenView : LoadRemoteDocumentMode_AddToProject;
+    for (int i = 0; i < resIds.size(); i++) {
+        if (!addToProject) {
+            hints.insert(SHOW_REPORT_HINT, true);
```
yalgaer: This should be a single report for the whole op, but not a spam of potentially 1000 notifications per each downloaded document.

yalgaer (changes requested):
The major issues

1) DownloadRemoteFileDialog task must produce a single report, not a N reports (N may be large).

2)  AVItem that keeps URL should work as before: the URL in AVItem is abstract and may come not from the external DB registry

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
DmitriiSukhomlinov: > There was a contract in the AnnotationsTreeView: if an item contains an fileUrl it points to the downloadable resource.

Could you, please, provide an exact scenario of this. There is no GUI tests for this scenario, also I tried to edit qualifiers by my own and failed in reproducing, Without scenario I can't get what is wrong about it. If you provide a scenario, I will fix and and will add a GUI test.

src/corelibs/U2Core/src/tasks/LoadRemoteDocumentTask.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/tasks/LoadRemoteDocumentTask.h
DmitriiSukhomlinov: OK

src/corelibs/U2Core/src/tasks/LoadRemoteDocumentTask.h
DmitriiSukhomlinov: OK

src/plugins/GUITestBase/src/runnables/ugene/plugins/dna_export/GetSequenceByIdFiller.h
DmitriiSukhomlinov: OK

src/corelibs/U2Gui/src/util/DownloadRemoteFileDialog.cpp
DmitriiSukhomlinov: Yes, of course, it's how it works now!

DmitriiSukhomlinov:
> DownloadRemoteFileDialog task must produce a single report, not a N reports (N may be large).

This is how it works now.

> AVItem that keeps URL should work as before: the URL in AVItem is abstract and may come not from the external DB registry

Please, provide a scenario, I couldn't handle to reproduce it. The scenario required at least for me to write a GUI test to prevent accidentally removing this feature in the future.

src/corelibs/U2View/src/ov_sequence/AnnotationsTreeView.cpp
yalgaer: AVItem interface provides the following API:  

`QString getFileUrl(int col) const;`

The source of the URL is intentionally abstracted. It can be either dbXref or some other URL. This was done intentionally in UGENE as a point of a potential extension.

The one who works with this URL should deal with it as with the abstract URL. 

If you want to restore the original DB Xref information from this URL you should handle it explicitly in some way (like checking the qualifier name).

Moving this discussion to the bottom of the issue.

src/corelibs/U2Gui/src/util/DownloadRemoteFileDialog.cpp
yalgaer: Ok, I see MultiTask handles it.

yalgaer (changes requested):
Only 1 issue left: please keep the fileUrl (link) abstracted. 
If you need to extract extra information about the URL use qualifier name (for example).

Reason for this: UGENE may have different URLs in AVItem that are based on the qualifier semantic. 
The reason that we support only DbXRef today is a miss of functionality that is worth to be implemented.
Example of other qualifiers that may be rendered like URLs:

```
allele (external link may be built on top of the value)
anticodon (inner link, pos:join(5,495..496),aa:Leu,seq:taa)
bio_material (external link [<institution-code>:[<collection-code>:]]<material_id>)
citation (external link may be opened for the citation)
bio_material (external link may be built for the material <institution-code>:[<collection-code>:]]<material_id>)
bound_moiety (external link may be built for a molecule, for example to PDB)
cell_type (external link may be opened, like Wikipedia or inner docs)
chromosome (external link to the whole chromosome)
....

read more: https://www.insdc.org/submitting-standards/feature-table
```

Example of a code where we mark a column as 'linked' and assign fileUrl to it:
```
bool AVItem::processLinks(const QString& qName, const QString& qValue, int col) {
...
```

Below is a description of the complete solution:
1) Make `LoadRemoteDocumentTask` work only with a raw URL.
2) Extend it (by inheritance or composition) to support URL building from DbXRef & custom report generation.
3) In the place where you need to run `LoadRemoteDocumentTask` based on DbXRef use the extended version from 2)

It is not necessary to implement the complete solution in your PR, the description above should demonstrate you the final state and all intermediate patches (like yours) should not go into wrong direction. In your patch you removed the code that works with abstract URLs in `LoadRemoteDocumentTask` and this is a wrong way.

The minimal solution for you can be as simple as this:
1) Inside of the sl_itemClicked() check if the clicked item is a dbxref qualifier.
2) If it is use a different constructor for the LoadRemoteDocumentTask.

The only expected modification in the `LoadRemoteDocumentTask` should be a `generateReport` method: it should provide a report in case if db & accession are provided or something generic if only an URL is provided.

Once you submit the minimal changes I can join and implement the complete solution (split LoadRemoteDocumentTask into the URL & DbXRef based versions) in a separate PR.

DmitriiSukhomlinov:
Done

yalgaer:
Will approve after the tests

yalgaer (approved):

PR #1276, opened 2023-06-05
UGENE-7887 Show full enzyme as a tooltip of the "sequence" column

DmitriiSukhomlinov:


src/plugins/GUITestBase/src/tests/common_scenarios/sequence_view/GTTestsSequenceView.cpp
```diff
@@ -2367,6 +2368,41 @@ GUI_TEST_CLASS_DEFINITION(test_0079_2) {
...
+            GTUtilsDialog::waitForDialog(os, new GTFileDialogUtils(os, testDir + "_common_data/enzymes/all_possible_tooltips.bairoch"));
+            GTWidget::click(os, GTWidget::findWidget(os, "enzymesFileButton", dialog));
+
+            auto tree = GTWidget::findTreeWidget(os, "tree", dialog);
+            for (int i = 0; i < 22; i++) {
+                auto name = QString::number(i);
+                if (name.size() == 1) {
+                    name.insert(0, "0");
+                }
+                 name.insert(0, "A");
```
yalgaer: nit: reformat the changes

src/plugins/enzymes/src/FindEnzymesDialog.h
```diff
@@ -134,6 +134,14 @@ class EnzymeTreeItem : public QTreeWidgetItem {
     EnzymeTreeItem(const SEnzymeData& ed);
     SEnzymeData enzyme;
     bool operator<(const QTreeWidgetItem& other) const;
+
+private:
+    enum class Ns {
```
yalgaer: nit: move private stuff into C++ if possible: this way you won't slow down a recompilation in case it is changed.

yalgaer: Document the name if values. Or use a good enum name that won't require documentation.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -771,6 +783,211 @@ bool EnzymeTreeItem::operator<(const QTreeWidgetItem& other) const {
...
+        return calculateNShiftRes;
+    };
+    // generates tooltop elements, which are located outside of enzyme
+    // these elements will be joined in the end
+    // returns list of elements
+    auto generateTooltipElements = [](int out, int in, bool forward, Ns type, bool otherHasLeftOut, bool otherHasLeftIn) -> QStringList {
+        auto outPart = [out, in, forward, type, otherHasLeftOut]() {
+            QStringList generateTooltipElementsResult;
+            if (out != 0) {
+                switch (type) {
```
yalgaer, line 846: nit: reformat using the project code style.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -771,6 +783,211 @@ bool EnzymeTreeItem::operator<(const QTreeWidgetItem& other) const {
...
+            calculateNShiftRes.first = forward ? Ns::Right : Ns::Left;
+            calculateNShiftRes.second = cut - enzymeSize;
+        }
+        return calculateNShiftRes;
+    };
+    // generates tooltop elements, which are located outside of enzyme
+    // these elements will be joined in the end
+    // returns list of elements
+    auto generateTooltipElements = [](int out, int in, bool forward, Ns type, bool otherHasLeftOut, bool otherHasLeftIn) -> QStringList {
+        auto outPart = [out, in, forward, type, otherHasLeftOut]() {
```
yalgaer: nit: document inner functions. Start function names with verbs

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -771,6 +783,211 @@ bool EnzymeTreeItem::operator<(const QTreeWidgetItem& other) const {
...
+        }
+        return result;
+    };
+    QString forwardMainPart = generateMainPart(enzyme->seq, enzyme->cutDirect, true);
+    QString reverseMainPart = generateMainPart(seqComplement, enzyme->cutComplement, false);
+
+    // Join parts to a single tooltip
+    auto generateTooltip = [](Ns type, const QStringList& elements, const QString& mainPart) -> QString {
+        QString result;
+        auto appendElements = [&elements]() -> QString {
```
yalgaer: nit: join will sound better, append usually mean you append to the current (some) object in the context.

Btw: do I understand correctly that you compute the joined value in any case? In this case why don't you just compute it as variable and append later to the correct part: why all this complexity with the inner function - the code can be simpler and more straightforward (readable)

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -771,6 +783,211 @@ bool EnzymeTreeItem::operator<(const QTreeWidgetItem& other) const {
...
+            QString appendElementsResult;
+            for (int i = 0; i < elements.size(); i++) {
+                const auto el = elements[i];
```
yalgaer: nit: add &, use non-indexed for loop (looks like you do not need `i`

yalgaer (commented):
Looks good, minor style issues only.
Please update, run the tests and once the tests are green I will approve

src/plugins/GUITestBase/src/tests/common_scenarios/sequence_view/GTTestsSequenceView.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/FindEnzymesDialog.h
DmitriiSukhomlinov: OK

DmitriiSukhomlinov: OK

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp

> Btw

I'm not sure how to answer this question, but he main point here that I have thought that this issue would be much easier that it's turned out. The main complexity here is that the direct enzyme depends on the reverse-complement (and vice-versa), so it is very complicated to calculate, for example, the beginning part of the direct enzyme until you know how the whole reverse enzyme looks like. This is why I need so mana pre-calculation functions and join all calculated stuff in the very end. 

You may check the enzyme file which I use in GUI test - there are a lot of different cases which calculates differently, it was very complicated to handle them all.

src/plugins/enzymes/src/FindEnzymesDialog.cpp

yalgaer (commented):
Looks good. Will approve after the tests

yalgaer (approved):

PR #1279, opened 2023-06-07
UGENE-7888 Make the Suppliers tooltip separated by a new line

DmitriiSukhomlinov:
Very minor, so no GUI test. I will add if needed

PR #1280, opened 2023-06-07
UGENE-7886 Remove Accession and Type columns

DmitriiSukhomlinov:


DmitriiSukhomlinov:
Very minor, so no GUI test. I will add if needed

yalgaer:
Why do you think the columns are not needed? There will be no other way except manually opening gz file to see these fields for users.

Types are important and even shown in a long textual form by the rebase search (we show a shortened code name only)

Accessions are the best way to access the exact enzyme version in the database using direct link, like http://rebase.neb.com/rebase/enz/42121.html

May be a better patch would be extending the width of the dialog (up to 1200-1300px) ?
For the accession column: I would recommend making them links (unless it make UI slow).

DmitriiSukhomlinov:
> Why do you think the columns are not needed? There will be no other way except manually opening gz file to see these fields for users.

I haven't seen this values in other tools (as SnapGene) and I can't even explain what they mean. "Accession" is pretty understandable, but the meaning of "Type" is vague. I cannot explain it when someone asks me about it (like in seminars has happened).

> Types are important and even shown in a long textual form by the rebase search (we show a shortened code name only)

It is not clear, what they mean. Need to add some kind of encoding. Also, rebase search do not have this parameter, "Organism type" is not related to this, this is called "Enzyme type", different thing.

> Accessions are the best way to access the exact enzyme version in the database using direct link, like http://rebase.neb.com/rebase/enz/42121.html

The same thing works with the enzyme name - http://rebase.neb.com/rebase/enz/AbaHI.html. We may add external link using its name.  
This is not critical at all, we may keep this parameter (in this case, I can at least suggest to remove the "RB" prefix, I suppose, it refers us to the "REBASE" word, but not used anywhere, just the number has meaning).

> May be a better patch would be extending the width of the dialog (up to 1200-1300px) ?

Not more than 1024, a lot of monitors has maximum possible resolution 1024x768. In this case, I would like to add a new text area and show the selected enzyme info (because I think, it would be not really convenient to look at some information - like at enzyme's cuts, which I've done in [UGENE-7887](https://ugene.dev/tracker/browse/UGENE-7887) - on tooltips only). It will be the same "Selected enzymes" text area on the left and new the "Selected enzyme info" area on the right:
image

> For the accession column: I would recommend making them links (unless it make UI slow).

Good idea.

So, the result:
1. Take the "Accession" column back, remove "RB" in the beginning of each accession, make each accession clickable external link.
2. Take the "Type" column back, rename it with "Enzyme type" (to avoid confusion with the "Organism type"), leave the same short value in the column, but add meaning as tooltip and to the "Selected enzyme info" area (see point 4).
3. Increase the width of the dialog, but not too much (I will test the best width for each column).
4. Add new the "Selected enzyme info" area, add there information about the selected enzyme.

@yalgaer agreed?

yalgaer:
> I haven't seen this values in other tools (as SnapGene) and I can't even explain what they mean. "Accession" is pretty understandable, but the meaning of "Type" is vague. I cannot explain it when someone asks me about it (like in seminars has happened).

"Type" is the first line on every enzyme page.
For a better documentation/contextual help you can add a tooltip that will build a complete type name.
image


I would not rename it to "Enzyme Type" because 1) This will cause negative UX effect: the column will be wider than because of the name, not data. 2) In clear from the context: if every line is an enzyme, the "Type" is a type of the enzyme unless other value is specified (like "Organism type")

yalgaer:
> > Accessions are the best way to access the exact enzyme version in the database using direct link, like http://rebase.neb.com/rebase/enz/42121.html
> 
> The same thing works with the enzyme name - http://rebase.neb.com/rebase/enz/AbaHI.html. We may add external link using its name. This is not critical at all, we may keep this parameter (in this case, I can at least suggest to remove the "RB" prefix, I suppose, it refers us to the "REBASE" word, but not used anywhere, just the number has meaning).

I would show exactly the value like it is written in the file. So if a user opens REBASE as a regular text file no extra translation is needed.

The number in a link is better than the Enzyme name because 1) the number is not affected by versioning (updating of enzyme details) 2) the number is not affected by the visual name/possible renaming

yalgaer:
> > May be a better patch would be extending the width of the dialog (up to 1200-1300px) ?
> 
> Not more than 1024, a lot of monitors has maximum possible resolution 1024x768. In this case, I would like to add a new text area and show the selected enzyme info (because I think, it would be not really convenient to look at some information - like at enzyme's cuts, which I've done in [UGENE-7887](https://ugene.dev/tracker/browse/UGENE-7887) - on tooltips only). It will be the same "Selected enzymes" text area on the left and new the "Selected enzyme info" area on the right: image

Ok, lets make it 1024 or even less.
I would not rely a lot on tooltips because they are not accessible in general. The rule for the tooltip can be: a tooltip should contain some very basic and generic help method, but not the data a user really needed. The data should be shown directly in a more printable/accessible way.

yalgaer:
> So, the result:
> 
>     1. Take the "Accession" column back, remove "RB" in the beginning of each accession, make each accession clickable external link.

OK, but keep the AC with no modification.

> 
>     2. Take the "Type" column back, rename it with "Enzyme type" (to avoid confusion with the "Organism type"), leave the same short value in the column, but add meaning as tooltip and to the "Selected enzyme info" area (see point 4).

Do not rename the column. There is no confusion. Add a tooltip for those who have the confusion.

>     3. Increase the width of the dialog, but not too much (I will test the best width for each column).
Ok.

You can add an extra improvements here: 
1) Add a search/filter field
2) Allow viewing the list flattened. This way a global sorting will be possible.

DmitriiSukhomlinov:
> Make each accession clickable external link

I haven't done it because it's a bit more complicated, that I anticipated. We use `QTreeWidget` as a tree view for enzymes. This widget isn't very flexible and it's not possible to add a hyperlink to its column in a native way. It could be done by adding some widget as column element (like `QLabel` or `QTextEdit`), but we have about 16 thousand enzymes, it makes everything extremally freezing (of course, I have tested). The correct way is to replace `QTreeWidget` with `QTreeView` with some custom model. I think that, al least, there is no need to write the whole model from scratch, [QStandartItemModel](https://doc.qt.io/qt-6/qstandarditemmodel.html) could be used. Instead of hyperlinking the accession number I have added a hyperlink to the new description field. 

> Type - leave the same short value in the column, but add meaning as tooltip and to the "Selected enzyme info" area (see point 4).

Done.

> Increase the width of the dialog, but not too much (I will test the best width for each column).

As far as I'm concerned, 100 x 600 is good enough.

> The rule for the tooltip can be: a tooltip should contain some very basic and generic help method, but not the data a user really needed. The data should be shown directly in a more printable/accessible way.

I have created a separate view, which shows the enzyme information when this enzyme is shown.

> Add a search/filter field

It is already in the dialog, isn't it?

> Allow viewing the list flattened. This way a global sorting will be possible.

Don't get the point.

DmitriiSukhomlinov:
@yalgaer please, review again

DmitriiSukhomlinov:
Linux GUI test 4209 passes locally and has no relation lo this feature

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -783,6 +792,22 @@ bool EnzymeTreeItem::operator<(const QTreeWidgetItem& other) const {
     return text(col) < ei.text(col);
 }
 
+QString EnzymeTreeItem::getEnzymeInfo() const {
+    QString result;
+    auto accession = text(Column::Accession);
+    accession = accession.right(accession.size() - 2);
```
yalgaer: Use accession in the link. Otherwise it is an unused variable.
Check that accession is RB (rebase) before opening the link.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -753,21 +763,20 @@ static const QString TOOLTIP_SPACE = "<sub>&nbsp;</sub>";
...
-    setText(1, enzyme->accession);
-    setData(1, Qt::ToolTipRole, tt);
-    setText(2, enzyme->type);
-    setData(2, Qt::ToolTipRole, tt);
-    setText(3, enzyme->seq);
-    setData(3, Qt::ToolTipRole, tt);
-    setText(4, enzyme->organizm);
-    setData(4, Qt::ToolTipRole, enzyme->organizm);
+    setText(Column::Id, enzyme->id);
+    setData(Column::Id, Qt::ToolTipRole, enzyme->id);
```
yalgaer: Is there any reason to duplicate item text in the tooltip? The tooltip is useless in this case. May be show a unified tooltip for the whole row with complete details?

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -1021,6 +1046,38 @@ QString EnzymeTreeItem::generateEnzymeTooltip() const {
     return TOOLTIP_TAG.arg(forwardTooltip).arg(reverseTooltip);
 }
 
+QString EnzymeTreeItem::getTypeInfo() const {
+    auto type = text(Column::Type);
+    QString result;
+    static constexpr const char* UNEXPECTED_TYPE_ERROR = QT_TR_NOOP("Unexpected enzyme type: %1");
```
yalgaer: You can just have a simple and more readable:
 `QString unexpectedError = tr("Unexpected enzyme type: %1")`
line instead of all that magic with `constexpr` and `QT_TR_NOOP`

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -1021,6 +1046,38 @@ QString EnzymeTreeItem::generateEnzymeTooltip() const {
...
+
+    } else if (type.size() == 2) {
+        if (type == "IE") {
+            result = tr("An intron-encoded (homing) endonuclease");
+        } else if (type.startsWith("R")) {
+            result = tr("The restriction enzyme of the %1 type").arg(type.back());
+        } else if (type.startsWith("M")) {
+            result = tr("The methylase of the %1 type").arg(type.back());
+        } else {
+            coreLog.error(tr(UNEXPECTED_TYPE_ERROR).arg(type));
```
yalgaer: This is not an error at all. 
I would even avoid logging it into trace because it may create a lot of spam in the log. 
User already sees the type/absence of the type in the text view.

yalgaer (changes requested):
Looks good. Minor issues only. Please fix and I will approve

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: I just thought that it's convenient. I think, it's better to remove it.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: `constexpr` isn't magic, just a modern C++ feature. `QT_TR_NOOP` helps translate static variables, makes this translation "delayed" until it should be done in runtime. 

You code creates variable in runtime - unneeded computations. Probably, compiler will optimize it, but you can't be sure.

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: Thank you for the explanation about C++ constructs but my comment was about readability.

Please use the following standard code for translations in UGENE:
```
QString unexpectedEnzymeTypeMessage = tr("Unexpected enzyme type: %1");
```

instead of the non standard `static constexpr const char* & QT_TR_NOOP`. 

A non-standard code stands out when you read/review the code and here there is no reason to use a non-standard solution

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -1021,6 +1042,38 @@ QString EnzymeTreeItem::generateEnzymeTooltip() const {
...
+    } else if (type.size() == 3) {
+        if (type.startsWith("R") && type.endsWith("*")) {
+            result = tr("The restriction enzyme of the %1 type,<br>but only recognizes the sequence when it is methylated").arg(type.at(1));
+        } else if (type.startsWith("RM")) {
+            result = tr("The enzyme of the %1 type, which acts as both -<br>a restriction enzyme and a methylase").arg(type.back());
+        } else {
+            coreLog.details(tr(UNEXPECTED_TYPE_ERROR).arg(type));
+        }
+    } else {
+        coreLog.details(tr(UNEXPECTED_TYPE_ERROR).arg(type));
```
yalgaer: Remove logging about our parsing issues. 
If you think that our parsing failure is important to user just show this in the view.

yalgaer:
Here what I see in Rebase Web interface:
image


And here what I see in UGENE:
image

Note that we show `BTGATAG` and Rebase shows 
image

I think we should repeat the Rebase behavior and always show 3/5 ends for the sequences to avoid confusion or multiple interpretation of the same data

yalgaer:
When I run the dialog in Linux with default settings the info field is not big enough to include basic enzyme details with no extra scrolling:
image

Please increase the field size

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -1021,6 +1042,38 @@ QString EnzymeTreeItem::generateEnzymeTooltip() const {
...
+        } else if (type.startsWith("R")) {
+            result = tr("The restriction enzyme of the %1 type").arg(type.back());
+        } else if (type.startsWith("M")) {
+            result = tr("The methylase of the %1 type").arg(type.back());
+        } else {
+            coreLog.details(tr(UNEXPECTED_TYPE_ERROR).arg(type));
+        }
+    } else if (type.size() == 3) {
+        if (type.startsWith("R") && type.endsWith("*")) {
+            result = tr("The restriction enzyme of the %1 type,<br>but only recognizes the sequence when it is methylated").arg(type.at(1));
```
yalgaer: "of the 2 type" sounds bad in English and not correct from the Rebase terminology
Use: "Type II"

yalgaer:
For Type II enzymes Rebase page also shows `subtype` info
image

Can we show this info too?

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp

DmitriiSukhomlinov:
> Here what I see in Rebase Web interface: image
> 
> And here what I see in UGENE: image
> 

> 


> Note that we show `BTGATAG` and Rebase shows image

I already noticed this mistake and fixed [here](https://github.com/ugeneunipro/ugene/pull/1279/files#r1224103217).

> I think we should repeat the Rebase behavior and always show 3/5 ends for the sequences to avoid confusion or multiple interpretation of the same data

OK

src/plugins/enzymes/src/FindEnzymesDialog.cpp

DmitriiSukhomlinov:
> For Type II enzymes Rebase page also shows `subtype` info image
> 
> Can we show this info too?

The bairoch file does not have this information.

DmitriiSukhomlinov:
> When I run the dialog in Linux with default settings the info field is not big enough to include basic enzyme details with no extra scrolling: image
> 
> Please increase the field size

It varieties on different OSs, I increased size (but on the computer I tested it size was not like at your picture)

DmitriiSukhomlinov:
This PR and UGENE-7888 started conflicting each other, so I added UGENE-7888 here.

yalgaer:
The patch looks good. I'm ready to approve. Please check why XML tests failed or just re-run.

yalgaer (approved):

PR #1290, opened 2023-06-13
UGENE-7872 PDB parser should inform about wrong indexes

rasputinkirill:


src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -190,6 +190,7 @@ void PDBFormat::PDBParser::parseBioStruct3D(BioStruct3D& biostruct, U2OpStatus&
     char* buf = readBuff.data();
     qint64 len = 0;
     bool firstCompndLine = true;
+    QList<int> idSequence;
```
yalgaer: nit: rename to `sequenceIds` (a collection of sequence ids)

src/corelibs/U2Formats/src/PDBFormat.h
```diff
@@ -91,7 +91,7 @@ class U2FORMATS_EXPORT PDBFormat : public TextDocumentFormatDeprecated {
         void parseSecondaryStructure(BioStruct3D& biostruct, U2OpStatus& ti);
         void parseHet(BioStruct3D& biostruct, U2OpStatus& ti);
         void parseAtomConnections(BioStruct3D& biostruct, U2OpStatus& ti);
-        void parseAtom(BioStruct3D& biostruct, U2OpStatus& ti);
+        void parseAtom(BioStruct3D& biostruct, U2OpStatus& ti, QList<int>& idSequence);
```
yalgaer: Add documentation what the last parameter is for. Rename to `sequenceIds` too

src/corelibs/U2Formats/src/PDBFormat.h
```diff
@@ -104,7 +104,9 @@ class U2FORMATS_EXPORT PDBFormat : public TextDocumentFormatDeprecated {
 
         // Returns the end-of-name index for `specification`. Changes `readingMoleculeName` if needed.
         int returnEndOfNameIndexAndUpdateParserState(const QString& specification);
-
+        
+        // Check if IDs in list follows natural order
+        void checkIdsOrder(QList<int>& idSequence, U2OpStatus& ti);
```
yalgaer: nit: 
1. Check -> Checks
2. Add details why `sequenceIds` is modified inside of the method.
Move the method into cpp file - there is not need to have it in the public interface: make it a static private cpp function

src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -265,6 +268,17 @@ void PDBFormat::PDBParser::parseBioStruct3D(BioStruct3D& biostruct, U2OpStatus&
     updateSecStructChainIndexes(biostruct);
 }
 
+void PDBFormat::PDBParser::checkIdsOrder(QList<int>& idSequence, U2OpStatus& ti) {
+    QList<int> sortedIdSequence(idSequence);
+    std::sort(sortedIdSequence.begin(), sortedIdSequence.end());
+    if (sortedIdSequence != idSequence) {
+        ti.addWarning(PDBFormat::tr("The atoms in the file are not located according to the order of their identifiers."
```
yalgaer: Is it OK to open such PDB files? Why don't we fail and mark the document as broken? Are there public sources (databases) that provide such broken files and what do we expect from user after showing the message in the log?

rasputinkirill: There are no such files among ~30 PDBs in our repositories.
According to PDB format documentation http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM this field described as ''serial number", with no other description. As i think value supposed to follow natural order starting with 1.
According to our code this id used as key in map and it should not cause crash - we do not rely on their order, we wiil construct molecule in any case.
I think its not a criminal to open such files and show what we got, especially user already warned in log about wrong order in file.
It is up to you to decide which solution to apply.

src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -185,11 +185,24 @@ PDBFormat::PDBParser::PDBParser(IOAdapter* _io)
     flagAtomRecordPresent = false;
 }
 
+// Checks if IDs in list follows natural order, keep in mind - list will be cleared after
+static void checkIdsOrder(QList<int>& sequenceIds, U2OpStatus& ti) {
+    QList<int> sortedIdSequence(sequenceIds);
+    std::sort(sortedIdSequence.begin(), sortedIdSequence.end());
+    if (sortedIdSequence != sequenceIds) {
+        ti.addWarning(QObject::tr("The atoms in the file are not located according to the order of their identifiers."
```
yalgaer: Use PDBFormat class scope for tr()

yalgaer: Could you please move list cleanup out of the method and clean it outside?

src/corelibs/U2Formats/src/PDBFormat.cpp
```diff
@@ -185,11 +185,24 @@ PDBFormat::PDBParser::PDBParser(IOAdapter* _io)
...
+    }
+    sequenceIds.clear();
+}
+
 void PDBFormat::PDBParser::parseBioStruct3D(BioStruct3D& biostruct, U2OpStatus& ti) {
     QByteArray readBuff(DocumentFormat::READ_BUFF_SIZE + 1, 0);
     char* buf = readBuff.data();
     qint64 len = 0;
     bool firstCompndLine = true;
+    QList<int> idSequence;
```
yalgaer: Rename to `sequenceIds` to match other places

src/corelibs/U2Formats/src/PDBFormat.cpp
yalgaer: Do you think the user will understand the details in the message in order to make a decision?

If the wrong order can break anything in UGENE algorithms/rendering: we should not allow opening this file or reading the broken models at all.

If the wrong order is automatically recovered by UGENE the warning message should announce that "UGENE fixed the order of atoms..." and open the file.

Here is a source of hundreds of PDBs (if you need)
```
wget -m -nH --cut-dirs=6 -np --reject 'index.*' https://files.wwpdb.org/pub/pdb/data/structures/models/current/pdb/
```

yalgaer (commented):
Looks good, will approve after tests

yalgaer (approved):

PR #1306, opened 2023-06-22
UGENE-7890 Add a filter by site length

DmitriiSukhomlinov:
Pay attention, that "N"s are not included into recognition sites (I added some colorization to the enzyme info field")


src/plugins/enzymes/src/FindEnzymesDialog.ui
```diff
@@ -113,38 +113,137 @@
     </widget>
    </item>
    <item>
-    <widget class="QGroupBox" name="groupBox_2">
+    <widget class="QGroupBox" name="groupBox_3">
      <property name="title">
-      <string>Show only selected suppliers:</string>
+      <string>Show/hide enzymes:</string>
```
yalgaer, line 118: nit: Use "Filter" label

src/plugins/enzymes/src/FindEnzymesDialog.ui
yalgaer, line 1: The suggested layout has several flows:
image

1) It consumes 3 lines of vertical space and the vertical space for this dialog already becomes too big
2) It uses a composition of 2 states: min/max selector + short enzymes
3) It has long labels. 

Please update the layout to a simple 1-liner:
image

To avoid 2 states make the 'min' selection starting from  '1'.
The default state (when checkbox is enabled) should be some reasonable filter (like 4+, up to you)

yalgaer (changes requested):
See comment about the dialog layout

src/plugins/enzymes/src/FindEnzymesDialog.ui
DmitriiSukhomlinov: I tried to use the word "Show" when we are talking about visibility of enzymes, which you are capable to choose. "Filter" is already in use in parameters, which describes the number of enzymes, which will be found (during the searching process). I thought, that it allows users to slit these two types of parameters:
image

I could change it, if you want.

src/plugins/enzymes/src/FindEnzymesDialog.ui
DmitriiSukhomlinov: The first thing is that I agree, that the current button design could be not very "beautiful" because of the huge amount of blank space. I can fix it.

The second thing - I do not like the interface you suggested, because:

1. It still has a huge amount of blank space:
image
This comment may be just skipped, because we have not enough widgets to fill the whole line (which is pretty large - about 700 pixels width). 
2. Usefulness of the "Length of recognition sequence" checkbox is doubtful. As far as I understood, the idea of this checkbox is to enable/disable this feature - so if this checkbox is checked, this kind of filtration works, if it is not checked - it doesn't and all enzymes are shown. From the other hand, if we set "Minimum" and "Maximum" values to "1" and "9+", all enzymes will be shown too. That means, that disabling "Length of recognition sequence" and setting "min-max" to "1-9+" has the same effect. I do not like this kind of _duality_. I assume, that this checkbox is excess here.
3. Checkbox "Short enzymes". First I need to say, that the whole interface wasn't invented by me - I "borrowed" it from SnapGene:
photo_2023-06-23_16-41-58
SnapGene and Genious do not have enzymes shorter that 4 bases. This forced me to think, that this kind of enzymes could be pretty useless and, sometimes even harmful (if you open long sequence and add 1-base enzyme it may freeze an application), so I decided that it is not required to have them separated by bases. 

So what I suggest:

1. Remake GUI as at your picture.
2. No additional "Length of recognition sequence" checkbox.
3. I would rather leave the "Show short" checkbox (but this may be because I do not wat to rewrite all connected code places),  but can replace it with 1, 2 and 3 values if you want.


@yalgaer agree?

DmitriiSukhomlinov:
Didn't fix GUI tests cuz I'm waiting for decision

src/plugins/enzymes/src/FindEnzymesDialog.ui
yalgaer: > 1. It still has a huge amount of blank space:

Agree, but I do not know how to fill that space. Having a spacer between the label & combo also does not look like a good solution because will separate logically related label & controls.

> 2. Usefulness of the "Length of recognition sequence" checkbox is doubtful. 

It depends how the whole control is designed. Here is how I think:

By default the dialog shows all enzymes (no filtering). With a checkbox disabled a user has a clear signal: the settings in this part of the filter has no effect.

If you do not have the checkbox the default state of the filter still should be 1...9+ (everything included). 
The filter will be 'always enabled' - a user must always pay an attention to it. 

I'm OK to remove the checkbox, but the dialog state will be more complex without the checkbox (more active controls, more enabled filtering states).  

Both with/without the checkbox the default filtering state should stay to 'all enzymes are shown'.

> 3. Checkbox "Short enzymes". First I need to say, that the whole interface wasn't invented by me - I "borrowed" it from SnapGene

Lets consider your option when the filter contains values from 4...9+ and we have a checkbox to show/hide short enzymes (the checkbox may be placed into the empty space discussed at the step 1) .

In this case we will have combination of 2 controls that do the same: control length of visible enzymes. 
And the default state for these 2 controls still should be: 100% of enzymes are shown.

This way the range selector will show 4...9+ and the checkbox will be checked by default. 

What I do not like about this solution is that a user will clearly see the range 4...9+ but that won't be the full truth: a user will have to adjust this range with the checkbox state: make a mental exercise. This can be a source of errors for the user.

All the above works as a proof to me that the solution I suggested is the least ambiguous one:
- By default the filter is turned off (a checkbox is disabled) - there is no reason even to look inside of the filter.
- Once enabled (a user wants us to filter) 
   - The default filtering range is 4...9+ (bad short enzymes are excluded)
   - A user see simple min & max values and does not need to combine state of multiple controls to understand the final logic of the filter.
 - Count of enabled controls is minimal. All controls fit into a single line and even have more space for the future extension.

Result: I tried to follow your suggestion and analyze the UX. I can't agree we need any additional complexity here beyond a simple range selector. I'm OK using/not-using the filter enabling checkbox. I'm not OK to add 'short enzymes' checkbox since it makes the filter state more ambiguous.

yalgaer:
UI looks good, please fix some minor issues (in other PRs)
image

1) Add some vertical spacing between suppliers & length filters (~5px)
2) Align "Region" with other filters horizontally.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -602,6 +618,37 @@ void FindEnzymesDialog::sl_onSelectionModified(int total, int nChecked) {
...
+        for (const auto& supplier : qAsConst(enzyme->suppliers)) {
+            CHECK_CONTINUE(checkedSuppliers.contains(supplier));
+
+            okSupplier = true;
+            break;
+        }
+
+        int recognitionSequenceLength = 0;
+        for (const auto& ch : qAsConst(enzyme->seq)) {
+            CHECK_CONTINUE(ch != 'N');
```
yalgaer, line 639: nit: you can add to the model of the enzyme: SEnzymeData

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -602,6 +618,37 @@ void FindEnzymesDialog::sl_onSelectionModified(int total, int nChecked) {
...
 
+void FindEnzymesDialog::sl_updateVisibleEnzymes() {
+    QStringList checkedSuppliers = cbSuppliers->getCheckedItems();
+    int min = cbMinLength->itemData(cbMinLength->currentIndex()).toInt();
+    int max = cbMaxLength->itemData(cbMaxLength->currentIndex()).toInt();
+    U2Region region(min, max - min + 1);
+    const auto enzymes = EnzymesSelectorWidget::getLoadedEnzymes();
+    QList<SEnzymeData> visibleEnzymes;
+    for (const auto& enzyme : qAsConst(enzymes)) {
+        bool okSupplier = false;
```
yalgaer, line 629: nit: replace with a 1-liner `std::` method

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -602,6 +618,37 @@ void FindEnzymesDialog::sl_onSelectionModified(int total, int nChecked) {
...
+            break;
+        }
+
+        int recognitionSequenceLength = 0;
+        for (const auto& ch : qAsConst(enzyme->seq)) {
+            CHECK_CONTINUE(ch != 'N');
+
+            recognitionSequenceLength++;
+        }
+        bool okRSLength = region.intersects(U2Region(recognitionSequenceLength, 1));
```
yalgaer, line 643: nit: use U2Region::contains(point)

yalgaer (commented):
Looks good, will approve after tests.

DmitriiSukhomlinov:
@yalgaer please, approve. All tests are passed (except some random in Windows GUI). I'll rework all nit comments in [UGENE-7890](https://ugene.dev/tracker/browse/UGENE-7889?focusedCommentId=32250&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-32250) to avoid tests re-running

yalgaer (approved):

PR #1314, opened 2023-06-27
UGENE-7889 Add enzyme's overhangs and the corresponding filter

DmitriiSukhomlinov:
+ comments [from here](https://github.com/ugeneunipro/ugene/pull/1306)


src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
```diff
@@ -71,19 +73,40 @@ class U2ALGORITHM_EXPORT EnzymeSettings {
...
     QString organizm;
     QStringList suppliers; // commercial sources of the current enzyme
-    const DNAAlphabet* alphabet;
+    const DNAAlphabet* alphabet = nullptr;
+
+    // Cuts (cleavages) on direct and reverse-complementary strands
+    // form overhang (located between these two cuts).
+    // These overhangs could have following types:
+    enum OverhangType {
+        // Enzyme has no cuts and, consequently, no overhangs
```
yalgaer: Nit: add dots to the end of each comment for consistency.

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -72,25 +72,51 @@ QList<SEnzymeData> EnzymesIO::readEnzymes(const QString& url, U2OpStatus& os) {
...
+                    auto second = seqSize - d->cutComplement;
+                    auto overhang = d->seq.mid(qMin(first, second), qAbs(first - second));
+                    auto overhangAlphabet = U2AlphabetUtils::findBestAlphabet(overhang);
+                    if (overhangAlphabet->getId() == BaseDNAAlphabetIds::NUCL_DNA_DEFAULT() &&
+                        !overhang.contains(EnzymeData::UNDEFINED_BASE)) {
+                        d->overhangTypes |= EnzymeData::OverhangType::NondegenerateSticky;
+                    }
+                    d->overhangTypes |= EnzymeData::OverhangType::Sticky;
+                } else {
+                    d->overhangTypes |= EnzymeData::OverhangType::Sticky;
```
yalgaer, line 104: nit: assign it in a single place. 
Replace 
```
 else if ((0 <= d->cutDirect && d->cutDirect < seqSize)...
```
with 
```
else {
  d->overhangTypes |= EnzymeData::OverhangType::Sticky;
  if ((0 <= d->cutDirect && d->cutDirect < seqSize)...
```

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -72,25 +72,51 @@ QList<SEnzymeData> EnzymesIO::readEnzymes(const QString& url, U2OpStatus& os) {
...
+                                    .arg(d->id)
+                                    .arg(d->alphabet->getId())
+                                    .arg(QString(d->seq)));
+            }
+            // calculate enzyme overhang types
+            auto seqSize = d->seq.size();
+            if (d->cutDirect == ENZYME_CUT_UNKNOWN) {
+                d->overhangTypes |= EnzymeData::OverhangType::NoOverhang;
+            } else {
+                if (qFuzzyCompare((double)seqSize / 2, d->cutDirect) && d->cutDirect == d->cutComplement) {
```
yalgaer: Why qFuzzyCompare with int? Can you just cast the int into double to avoid a warning?

nit:put the result of the comparison into a named variable - the algorithm will be more readable.

src/plugins/enzymes/src/EnzymesIO.cpp
```diff
@@ -72,25 +72,51 @@ QList<SEnzymeData> EnzymesIO::readEnzymes(const QString& url, U2OpStatus& os) {
...
+                    d->overhangTypes |= EnzymeData::OverhangType::Blunt;
+                } else if ((0 <= d->cutDirect && d->cutDirect < seqSize) &&
+                    (0 <= d->cutComplement && d->cutComplement < seqSize)) {
+                    auto first = d->cutDirect;
+                    auto second = seqSize - d->cutComplement;
+                    auto overhang = d->seq.mid(qMin(first, second), qAbs(first - second));
+                    auto overhangAlphabet = U2AlphabetUtils::findBestAlphabet(overhang);
```
yalgaer: nit: Do not search for the alphabet again if the alphabet of the whole sequence is already NUCL_DNA_DEFAULT

yalgaer (commented):
Looks good. Minor style issues only.
Will approve after tests

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: > Why qFuzzyCompare with int? Can you just cast the int into double to avoid a warning?

`int` explicitly casts to "double", isn't it? And I do not get the idea about warning - I do not have any warnings on WIndows.

> nit: put the result of the comparison into a named variable - the algorithm will be more readable.
OK

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: It could be `BaseDNAAlphabetIds::NUCL_DNA_DEFAULT()` or `BaseDNAAlphabetIds::NUCL_DNA_EXTENDED()`. The alphabet of overhang could be different from the whole enzyme alphabet.

src/plugins/enzymes/src/EnzymesIO.cpp
yalgaer: Skip the question about compiler warnings - the reason I asked is a non-standard code to deal with integers using fuzzy-compare for doubles.

Looks like the current algorithm has a problem. Consider the following scenario:

1. Direct cut is in the middle of the sequence: a, b, c, d, e 
2. length = 5, directCutIndex = 2 (c). 
3. Your algorithm: length / 2 = 2.5, so directCutInTheMiddleOfSequence is false.

Correct check (no fuzzy & doubles): 
`bool directCutInTheMiddleOfSequence  = (length % 2 == 1) && ((length - 1) / 2) == directCut;`

wdyt?

src/plugins/enzymes/src/EnzymesIO.cpp
yalgaer: I mean if the alphabet of the whole sequence is NUCL_DNA_DEFAULT any sub-sequence also has NUCL_DNA_DEFAULT - no need to run alphabet detection for the sub-sequence again in this case.

DmitriiSukhomlinov:
@yalgaer please, approve, all tests passed on a previous iteration (except some random one on Windows), I rebased a conflict (which was as minor as possible):

image

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: OK

yalgaer (changes requested):
Did you see my question about the potential bug in this line here https://github.com/ugeneunipro/ugene/pull/1314#discussion_r1245485748 ?
```
bool directCutInTheMiddleOfSequence = qFuzzyCompare((double)seqSize / 2, d->cutDirect);
```
This is the only issue left.

src/plugins/enzymes/src/EnzymesIO.cpp
DmitriiSukhomlinov: A bit incorrect statement, but OK, fixed.

yalgaer: I see so the directCutInTheMiddleOfSequence should point to a position between 2 bases, but not on the middle base itself.

yalgaer (approved):

PR #1338, opened 2023-07-12
UGENE-7897 Show cut sites when enzyme located on junction point

DmitriiSukhomlinov:
No GUI test because it could be checked by looking only

### DIFF
--- a/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
+++ b/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
@@ -460,7 +460,7 @@ qint64 SequenceViewAnnotatedRenderer::correctCutPos(qint64 pos) const {
         } else if (result > sequenceLength) {
             result = result - sequenceLength;
         }
-    } else if (result <= 0 || result >= sequenceLength) {
+    } else if (result < 0 || result > sequenceLength) {
         // out of sequence, do not draw
         result = INCORRECT_CUT_POSITION;
     }
--- a/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
+++ b/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
@@ -196,8 +196,8 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
      * Calculates restriction cuts positions depends on visible area and calls drawCutSite.
      *
      * \param reg Region of the annotation, which sites should be drawn.
-     * \param drawDirect Draw direct cut if true and do not draw if false.
-     * \param drawComplement Draw complement cut if true do not draw if false.
+     * \param drawDirect Draw direct cut if true and do not draw if false. Default value is true.
+     * \param drawComplement Draw complement cut if true do not draw if false. Default value is true.
      */
     auto drawCutSiteLambda = [this, &p, &visibleRange, &a, &as, &availableHeight, &canvasSize, selected, &aData](const U2Region& reg, bool drawDirect = true, bool drawComplement = true) {
         if (reg.intersects(visibleRange)) {
@@ -230,14 +230,12 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
             SAFE_POINT(merged.size() == 1, "Unexpected merged enzyme region", );
 
             auto regionPair = merged.front();
-            // draw cuts separatelly,
+            // draw cuts separately,
             // because they are located one on the 3' end of the direct sequence,
             // and the other one on the 3' end of reverse-complementary sequence
             drawCutSiteLambda(regionPair.first, true, false);
             drawCutSiteLambda(regionPair.second, false, true);
-        } else {
-            SAFE_POINT(location.size() == 1, "Unexpected enzyme region", );
-
+        } else if (location.size() == 1) {
             drawCutSiteLambda(location.first());
         }
     }
--- a/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
+++ b/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
@@ -192,35 +192,6 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
         }
     }
     // draw restriction sites
-    /**
-     * Calculates restriction cuts positions depends on visible area and calls drawCutSite.
-     *
-     * \param reg Region of the annotation, which sites should be drawn.
-     * \param drawDirect Draw direct cut if true and do not draw if false. Default value is true.
-     * \param drawComplement Draw complement cut if true do not draw if false. Default value is true.
-     */
-    auto drawCutSiteLambda = [this, &p, &visibleRange, &a, &as, &availableHeight, &canvasSize, selected, &aData](const U2Region& reg, bool drawDirect = true, bool drawComplement = true) {
-        if (reg.intersects(visibleRange)) {
-
-            const U2Region visibleLocation = reg.intersect(visibleRange);
-            const U2Region y = getAnnotationYRange(a, 0, as, availableHeight);
-            CHECK(y.startPos >= 0, );
-
-            const int x1 = posToXCoord(visibleLocation.startPos, canvasSize, visibleRange);
-            const int x2 = posToXCoord(visibleLocation.endPos(), canvasSize, visibleRange);
-
-            const int rw = qMax(selected ? MIN_SELECTED_ANNOTATION_WIDTH : MIN_ANNOTATION_WIDTH, x2 - x1);
-            SAFE_POINT(rw > 0, "Negative length of annotationYRange", );
-
-            const QRect annotationRect(x1, y.startPos, rw, y.length);
-            if (annotationRect.width() > MIN_WIDTH_TO_DRAW_EXTRA_FEATURES) {
-                drawCutSite(p, aData, reg, annotationRect, as->color, canvasSize, visibleRange, drawDirect, drawComplement);
-            }
-        } else {
-            drawCutSite(p, aData, reg, QRect(), as->color, canvasSize, visibleRange);
-        }
-    };
-
     if (isRestrictionSite && displaySettings.displayCutSites) {
         p.setPen(borderPen);
         const bool isCircular = ctx->getSequenceObject()->isCircular();
@@ -233,10 +204,44 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
             // draw cuts separately,
             // because they are located one on the 3' end of the direct sequence,
             // and the other one on the 3' end of reverse-complementary sequence
-            drawCutSiteLambda(regionPair.first, true, false);
-            drawCutSiteLambda(regionPair.second, false, true);
+            const auto& qual =  aData->findFirstQualifierValue(GBFeatureUtils::QUALIFIER_CUT);
+            bool isDirectStrand = aData->getStrand().isDirect();
+            {
+                auto cutPosDirect = getCutPosition(regionPair.first, qual, isDirectStrand, true);
+                if (cutPosDirect != INCORRECT_CUT_POSITION) {
+                    auto annRect = getAnnotationRect(regionPair.first, canvasSize, visibleRange, selected, a, as);
+                    drawCutSite(p, aData, cutPosDirect, annRect, as->color, canvasSize, visibleRange, true);
+                }
+            }
+            {
+                auto cutPosComplement = getCutPosition(regionPair.second, qual, isDirectStrand, false);
+                if (cutPosComplement != INCORRECT_CUT_POSITION) {
+                    auto annRect = getAnnotationRect(regionPair.second, canvasSize, visibleRange, selected, a, as);
+                    drawCutSite(p, aData, cutPosComplement, annRect, as->color, canvasSize, visibleRange, false);
+                }
+            }
         } else if (location.size() == 1) {
-            drawCutSiteLambda(location.first());
+            const auto& reg = location.first();
+            const auto& qual = aData->findFirstQualifierValue(GBFeatureUtils::QUALIFIER_CUT);
+            QRect annRect;
+            bool isDirectStrand = aData->getStrand().isDirect();\
+            {
+                auto cutPosDirect = getCutPosition(reg, qual, isDirectStrand, true);
+                if (cutPosDirect != INCORRECT_CUT_POSITION) {
+                    annRect = getAnnotationRect(reg, canvasSize, visibleRange, selected, a, as);
+                    drawCutSite(p, aData, cutPosDirect, annRect, as->color, canvasSize, visibleRange, true);
+                }
+            }
+            {
+                auto cutPosComplement = getCutPosition(reg, qual, isDirectStrand, false);
+                if (cutPosComplement != INCORRECT_CUT_POSITION) {
+                    if (annRect.isEmpty()) {
+                        annRect = getAnnotationRect(reg, canvasSize, visibleRange, selected, a, as);
+                    }
+                    drawCutSite(p, aData, cutPosComplement, annRect, as->color, canvasSize, visibleRange, false);
+                }
+            }
+
         }
     }
 }
@@ -268,6 +273,52 @@ void SequenceViewAnnotatedRenderer::drawBoundedText(QPainter& p, const QRect& r,
         return;
     }
     p.drawText(r, Qt::TextSingleLine | Qt::AlignCenter, text.left(prefixLen));
+}
+
+int SequenceViewAnnotatedRenderer::getCutPosition(const U2Region& r, const QString& cutQual, bool isAnnotationDirect, bool isCutDirect) const {
+    bool hasDirect = false;
+    bool hasComplement = false;
+    int cutDirect = 0;
+    int cutComplement = 0;
+    if (!cutQual.isEmpty()) {
+        int complSplit = cutQual.indexOf('/');
+        if (-1 != complSplit) {
+            cutDirect = cutQual.left(complSplit).toInt(&hasDirect);
+            cutComplement = cutQual.mid(qMin(cutQual.length(), complSplit + 1)).toInt(&hasComplement);
+        } else {
+            cutDirect = cutQual.toInt(&hasDirect);
+            cutComplement = cutQual.toInt(&hasComplement);
+        }
+    }
+
+    int cutPosition = INCORRECT_CUT_POSITION;
+    if (isCutDirect && hasDirect) {
+        cutPosition = correctCutPos(isAnnotationDirect ? r.startPos + cutDirect : r.startPos + cutComplement);
+    } else if (!isCutDirect && hasComplement) {
+        cutPosition = correctCutPos(isAnnotationDirect ? r.endPos() - cutComplement : r.endPos() - cutDirect);
+    }
+
+    return cutPosition;
+}
+
+QRect SequenceViewAnnotatedRenderer::getAnnotationRect(const U2Region& reg, const QSize& canvasSize, const U2Region& visibleRange, bool selected, Annotation* a, const AnnotationSettings* as) const {
+    QRect res;
+    if (reg.intersects(visibleRange)) {
+
+        const U2Region visibleLocation = reg.intersect(visibleRange);
+        const U2Region y = getAnnotationYRange(a, 0, as, canvasSize.height());
+        CHECK(y.startPos >= 0, res);
+
+        const int x1 = posToXCoord(visibleLocation.startPos, canvasSize, visibleRange);
+        const int x2 = posToXCoord(visibleLocation.endPos(), canvasSize, visibleRange);
+
+        const int rw = qMax(selected ? MIN_SELECTED_ANNOTATION_WIDTH : MIN_ANNOTATION_WIDTH, x2 - x1);
+        SAFE_POINT(rw > 0, "Negative length of annotationYRange", res);
+
+        res = QRect(x1, y.startPos, rw, y.length);
+    }
+
+    return res;
 }
 
 void SequenceViewAnnotatedRenderer::drawAnnotationConnections(QPainter& p, Annotation* a, const AnnotationSettings* as, const AnnotationDisplaySettings& displaySettings, const QSize& canvasSize, const U2Region& visibleRange) {
@@ -327,35 +378,7 @@ void SequenceViewAnnotatedRenderer::drawAnnotationConnections(QPainter& p, Annot
     }
 }
 
-void SequenceViewAnnotatedRenderer::drawCutSite(QPainter& p, const SharedAnnotationData& aData, const U2Region& r, const QRect& annotationRect, const QColor& color,
-                                                const QSize& canvasSize, const U2Region& visibleRange, bool drawDirect, bool drawComplement) {
-    SAFE_POINT(drawDirect || drawComplement, "At least direct or complement site should be drawn", );
-
-    const QString cutStr = aData->findFirstQualifierValue(GBFeatureUtils::QUALIFIER_CUT);
-    bool hasD = false;
-    bool hasC = false;
-    int cutD = 0;
-    int cutC = 0;
-    if (!cutStr.isEmpty()) {
-        int complSplit = cutStr.indexOf('/');
-        if (-1 != complSplit) {
-            if (drawDirect) {
-                cutD = cutStr.left(complSplit).toInt(&hasD);
-            }
-            if (drawComplement) {
-                cutC = cutStr.mid(qMin(cutStr.length(), complSplit + 1))
-                    .toInt(&hasC);
-            }
-        } else {
-            if (drawDirect) {
-                cutD = cutStr.toInt(&hasD);
-            }
-            if (drawComplement) {
-                cutC = cutStr.toInt(&hasC);
-            }
-        }
-    }
-
+void SequenceViewAnnotatedRenderer::drawCutSite(QPainter& p, const SharedAnnotationData& aData, int cutPos, const QRect& annotationRect, const QColor& color, const QSize& canvasSize, const U2Region& visibleRange, bool isDirectCut) {
     bool isDirectStrand = aData->getStrand().isDirect();
     int availableHeight = canvasSize.height();
     U2Region cutSiteY = getCutSiteYRange(U2Strand(isDirectStrand ? U2Strand::Complementary : U2Strand::Direct), availableHeight);
@@ -365,15 +388,13 @@ void SequenceViewAnnotatedRenderer::drawCutSite(QPainter& p, const SharedAnnotat
 
     CutSiteDrawData toInsert;
     toInsert.color = color;
-    if (hasD) {
+    toInsert.pos = cutPos;
+    if (isDirectCut) {
         toInsert.direct = true;
-        toInsert.pos = correctCutPos(isDirectStrand ? r.startPos + cutD : r.startPos + cutC);
         toInsert.r = isDirectStrand ? annotationRect : mirroredAnnotationRect;
         drawCutSite(p, toInsert, canvasSize, visibleRange);
-    }
-    if (hasC) {
+    } else {
         toInsert.direct = false;
-        toInsert.pos = correctCutPos(isDirectStrand ? r.endPos() - cutC : r.endPos() - cutD);
         toInsert.r = !isDirectStrand ? annotationRect : mirroredAnnotationRect;
         drawCutSite(p, toInsert, canvasSize, visibleRange);
     }
--- a/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
+++ b/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
@@ -142,7 +142,6 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
 
     const bool isRestrictionSite = a->getType() == U2FeatureTypes::RestrictionSite;
     QVector<U2Region> location = aData->getRegions();
-    bool simple = location.size() == 1;
     int availableHeight = canvasSize.height();
     for (int ri = 0, ln = location.size(); ri < ln; ri++) {
         const U2Region& r = location.at(ri);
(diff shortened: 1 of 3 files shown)


src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+     * Calculates restriction cuts positions depends on visible area and calls drawCutSite.
+     *
+     * \param reg Region of the annotation, which sites should be drawn.
+     * \param drawDirect Draw direct cut if true and do not draw if false.
+     * \param drawComplement Draw complement cut if true do not draw if false.
+     */
+    auto drawCutSiteLambda = [this, &p, &visibleRange, &a, &as, &availableHeight, &canvasSize, selected, &aData](const U2Region& reg, bool drawDirect = true, bool drawComplement = true) {
+        if (reg.intersects(visibleRange)) {
+
+            const U2Region visibleLocation = reg.intersect(visibleRange);
```
yalgaer, line 205: nit: non standard code style: we do not use consts on locals in UGENE (even your patch does not use them consistently)

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+        } else {
+            drawCutSite(p, aData, reg, QRect(), as->color, canvasSize, visibleRange);
+        }
+    };
+
+    if (isRestrictionSite && displaySettings.displayCutSites) {
+        p.setPen(borderPen);
+        const bool isCircular = ctx->getSequenceObject()->isCircular();
+        auto seqLength = ctx->getSequenceLength();
+        auto merged = U1AnnotationUtils::mergeAnnotatedRegionsAroundJunctionPoint(location, seqLength);
```
yalgaer: I think the function you wrote earlier (mergeAnnotatedRegionsAroundJunctionPoint) contains a bug. So you can't rely on it in the new code.
Check:
```cpp
for (int i = 0; i < regions.size(); i++) {
        const U2Region reg = regions[i];
        if (reg.endPos() != sequenceLength) {
            result.append(RegionsPair(reg, U2Region()));
            continue;
        }
        if ((i + 1) >= regions.size()) {
            result.append(RegionsPair(reg, U2Region()));
            break;
        }
```

how the ` if ((i + 1) >= regions.size())` can be true?

yalgaer: Move circular sequence related logic under `isCircular` check. No need  to calculate merged regions for non-circular sequences

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+        if (isCircular && U1AnnotationUtils::isAnnotationContainsJunctionPoint(merged)) {
+            SAFE_POINT(merged.size() == 1, "Unexpected merged enzyme region", );
+
+            auto regionPair = merged.front();
+            drawCutSiteLambda(regionPair.first, true, false);
```
yalgaer: Add a comment what is going on here.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+            drawCutSiteLambda(regionPair.second, false, true);
+        } else {
+            int length = 0;
```
yalgaer: remove unused var

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+
+            const int rw = qMax(selected ? MIN_SELECTED_ANNOTATION_WIDTH : MIN_ANNOTATION_WIDTH, x2 - x1);
+            SAFE_POINT(rw > 0, "Negative length of annotationYRange", );
+
+            const QRect annotationRect(x1, y.startPos, rw, y.length);
+            if (annotationRect.width() > MIN_WIDTH_TO_DRAW_EXTRA_FEATURES) {
+                drawCutSite(p, aData, reg, annotationRect, as->color, canvasSize, visibleRange, drawDirect, drawComplement);
+            }
+        } else {
+            drawCutSite(p, aData, reg, QRect(), as->color, canvasSize, visibleRange);
```
yalgaer: this call does not use `bool drawDirect` and `bool drawComplement` - it violates the lambda contract.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+     * \param reg Region of the annotation, which sites should be drawn.
+     * \param drawDirect Draw direct cut if true and do not draw if false.
+     * \param drawComplement Draw complement cut if true do not draw if false.
+     */
+    auto drawCutSiteLambda = [this, &p, &visibleRange, &a, &as, &availableHeight, &canvasSize, selected, &aData](const U2Region& reg, bool drawDirect = true, bool drawComplement = true) {
+        if (reg.intersects(visibleRange)) {
+
+            const U2Region visibleLocation = reg.intersect(visibleRange);
+            const U2Region y = getAnnotationYRange(a, 0, as, availableHeight);
+            CHECK(y.startPos >= 0, );
```
yalgaer: cut-site may be on the different y-level from the annotation region (outside of the annotation region). See AbaAI enzymes.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+
+            const int x1 = posToXCoord(visibleLocation.startPos, canvasSize, visibleRange);
+            const int x2 = posToXCoord(visibleLocation.endPos(), canvasSize, visibleRange);
+
+            const int rw = qMax(selected ? MIN_SELECTED_ANNOTATION_WIDTH : MIN_ANNOTATION_WIDTH, x2 - x1);
+            SAFE_POINT(rw > 0, "Negative length of annotationYRange", );
```
yalgaer: Here you measure annotation region. Cut-site may be outside of this region and you won't draw it in this case, right?

yalgaer (changes requested):

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: 1. Insert breakpoint inside the `if ((i + 1) >= regions.size())` condition (line 361).
2. Open “_common_data\cmdline\DNA_circular.gb”.
3. Create a new annotation "740..744" (use Genbank format).
Current: The breakpoint has been triggered.

I suppose, `>=` could be replaced with `==` - it won't change anything, but will be more accurate. Do I need to fix it?

DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: Do not get the point

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK, I got this, but I can't get the scenario, where something works wrong. Here you need to understand, that I didn't write this code by myself - I split some logic (drawing usual annotations and restriction sites) and re-used some code. E.g. see [here](https://github.com/ugeneunipro/ugene/pull/1338/files#diff-f378eee993cc84932724155c8f4dace9314d18d12c76c43d909129c04b233accL156) - this is pretty the same part. If you want, I can get rid of this splitting and, it this case, you will see it, but, as far as I think, code will be less accurate

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: Same, this line [from here](https://github.com/ugeneunipro/ugene/pull/1338/files#diff-f378eee993cc84932724155c8f4dace9314d18d12c76c43d909129c04b233accL160) and it does the same thing

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: Yes please replace it with a more readable code like 
```
bool isLastRegion = i == regions.size() - 1;
```

the `>=` sign here is misleading and looks like a bug: like somebody expects `i+1 > region.size()` to be true that can't ever happen.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,58 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+    if (isRestrictionSite && displaySettings.displayCutSites) {
+        p.setPen(borderPen);
+        const bool isCircular = ctx->getSequenceObject()->isCircular();
+        auto seqLength = ctx->getSequenceLength();
+        if (isCircular && U1AnnotationUtils::isAnnotationContainsJunctionPoint(a, seqLength)) {
+            auto merged = U1AnnotationUtils::mergeAnnotatedRegionsAroundJunctionPoint(location, seqLength);
+            SAFE_POINT(merged.size() == 1, "Unexpected merged enzyme region", );
+
+            auto regionPair = merged.front();
+            // draw cuts separatelly,
```
yalgaer: nit: typo in 'separately'

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: The `drawCutSiteLambda` accepts these params:
```
 * \param drawDirect Draw direct cut if true and do not draw if false.
 * \param drawComplement Draw complement cut if true do not draw if false.
```

and calls `drawCutSite` but does not pass these params to that function

Either pass the params to the `drawCutSite` or update the lambda documentation and give a clear details when these params are used and when are not. 
Right now it reads like the params are always being used.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: You are right: it can be a bug in the old code: in this case the bug should be addressed in a new PR. I not ask you to fix in this PR. But if you add a new code with a copied bug this is not good. The new code should not have the bug.

The biggest problem in your patch is the following: 
1) You separated cut-site drawing into a dedicated block: this is very good.
2) You rely on annotation regions for cut-sites geometry calculation, while cut-sites may be out of annotation regions: this is incorrect and should be fixed.

Look these examples for cut-site drawing related to the wrong geometry:

1) Good: Annotation fits X region.
image

2) Bad: Annotation does not fit X region: direct cut-site is not drawn.
image

For the wrap-mode we have an issue with Y:
1) Good, no wrap inside cut-site region
image

2) Bad: cutsite has wrong Y position
image

A good API test can be created if you extract a function that does not draw anything, but only calculate a cut-site layout in sequence (annotation regions) coordinates.

The function should accept as parameters: 
1) One annotation region (enzyme region)
2) a cutsite offset info

and will return another annotation region of length 1: where to draw the cut-site.

Next you can use 'annotation-region' logic to map the cutsite into screen and decide of to draw it (if it is visible) or not (if it out of the screen)

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: The comment does not answer the question why a circular sequence with a junction point needs to call drawCutSiteLambda twice, separately for each strand.

Look: the `merged` is a concatenation of 2 regions of a circular sequence: (P1...LEN) & (0...P2)
These both regions are located on the same strand.

Why do you draw cut-site for (P1...LEN) on the direct strand only and a cut-site for (0...P2) on the complement?

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: We can use annotation regions-only based math to draw annotation frames, but to draw cut-sites we need also use cut-site locations.

Resolving: lets continue the discussion in the comment above

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: Because we have two regions - (P1...LEN) and (0...P2). We need to calculate `annotationRect` and call `SequenceViewAnnotatedRenderer::drawCutSite` for each region separately.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: These parameters have default values, I thought their using pretty obvious. I update documentation about it.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK, I got the point, just have some question, to be certain:

1. What do you mean by "a cutsite offset info". Now, calculation of cut regions required very many different parameters (`canvasSize`, `visibleRange`, etc). Do you mean here "anything you need, but try at least to minimize the number of arguments"?
2. "will return another annotation region of length 1: where to draw the cut-site." - this is a pretty simple function, are you sure, that you want to extract to the separate function this exact part? In this case, we will need to extract [this part](https://github.com/ugeneunipro/ugene/pull/1338/files#diff-f378eee993cc84932724155c8f4dace9314d18d12c76c43d909129c04b233accR334-R359) of code to separate function - it would be enough to know sequence region of cut. But, even when we know this region, we still do need to process some calculations, to find out screen positions where we need to draw these cites (if we scroll sequence cut region remains the same, but screen position become different). Maybe, it would be better to calculate [CutSiteDrawData](https://github.com/ugeneunipro/ugene/blob/815c4f31ad3b2b853eec59bc7a01ee2024b6d9a9/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.h#L46), which have full information about "where this cut site should be drawn"?

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,56 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+            auto merged = U1AnnotationUtils::mergeAnnotatedRegionsAroundJunctionPoint(location, seqLength);
+            SAFE_POINT(merged.size() == 1, "Unexpected merged enzyme region", );
+
+            auto regionPair = merged.front();
+            // draw cuts separately,
+            // because they are located one on the 3' end of the direct sequence,
+            // and the other one on the 3' end of reverse-complementary sequence
+            drawCutSiteLambda(regionPair.first, true, false);
+            drawCutSiteLambda(regionPair.second, false, true);
+        } else if (location.size() == 1) {
```
DmitriiSukhomlinov, line 219: Here I found a problem (thanks to GUI tests) I need to discuss - it could be some possible bug or something. Previously, this line was just `else` and line above contained `SAFE_POINT`:  
`SAFE_POINT(location.size() == 1, ...);`  
But, I faced a problem here, **scenario**:

1.  Open "data\samples\Genbank\PBR322.gb" (it is circular).
2. Open the "Restriction sites" dialog and choose the only enzyme "EcoRI".
Expected: enzyme on the junction point has been found.
3. Uncheck "Mark as circular".
The above mentioned `SAFE_POINT` has been triggered.

Logically, it shouldn't happen - if sequence was not circular, than restriction enzyme just wouldn't be found. The problem is here - when we uncheck "Mark as circular", the `Auto-annotations update` task being triggered to update all annotations. This task doesn't work instantly and when this task is still in process, old restriction enzymes are trying to be shown, and the site, which was located on junction point just being split to two parts and trying to be shown as "joined". As far as we anticipate, that restriction enzymes annotations are **always** inseparable in case of non-circular view, `SAFE_POINT` is being triggered. 

You thing it could be a bug and we need to prevent somehow splitting restriction enzymes to two joined parts or not?

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: Why the first region is predefined to be `bool drawDirect = true` and the second `bool drawComplement = true` ?

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: The problem still persist:

1. You call `drawCutSiteLambda(direct=false, complement=true)`

2. drawCutSiteLambda calls `drawCutSite(p, aData, reg, QRect(), as->color, canvasSize, visibleRange);` with default values which are `(direct=true complement=true)`

Right?

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: > What do you mean by "a cutsite offset info". 

I suggest to make this function screen-layout independent: return in-sequence position of the cut-site (like an annotation region). 
This way you can test this function using API tests without any screen geometry.

>"will return another annotation region of length 1: where to draw the cut-site." - this is a pretty simple function, are you sure, that you want to extract to the separate function this exact part?

Yes, simple functions with a small amount of parameters are always good.  They are readable and testable.

What we have in the current solution is a lambda that accepts a lot of parameters and depends on the current context and you said you can't even test it.

As the result there may be these functions created:
1) Given an annotation calculate cut-site coordinates in sequence coordinates
2) Given a cut-site  sequence coordinates calculate on-screen coordinates

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
yalgaer: Here is what I think

1) We should show enzymes correctly for circular sequences.
2) We have no other way to draw annotations on circular sequences that cross junction point but only to split them into 2
3) The above makes enzymes split into 2 - we have no option

The code that deals with enzymes should be ready for this situation: allow 2 regions of kind (P1...LEN) & (0...P2) and handle them correctly for circular sequences

This is what you already do in this patch and this is good.


Next for non-circular sequences: 
1) There is no such concept like junction point anymore
2) All enzymes have only 1 region
3) The code is safe to skip 'wrong' enzymes: SAFE_POINT is not needed, we can ignore an enzyme with a "bad layout" and skip it silently.
4) When user switches Circular<->Non circular all auto-annotation tasks clean their results and re-run.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: It does not have some hidden meaning. As far as I remember, first `drawCutSiteLambda` had more calls than now (now it's called just three times, but in first version it was 5 times or something like this) and the most part of these calls had `..., true, true)` in the end, so I added these default values. Now it does not make ant sense, so it would be better just to remove default values.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: Yes, this is a typo, I forgot about these arguments. [CHECK](https://github.com/ugeneunipro/ugene/blob/a9113416d439fcd4e25eccf97d33960453e27fc9/src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp#L384) here prevented drawing extra cuts

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK, I tried to do something, just check

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -231,6 +275,52 @@ void SequenceViewAnnotatedRenderer::drawBoundedText(QPainter& p, const QRect& r,
...
 }
 
+int SequenceViewAnnotatedRenderer::getCutPosition(const U2Region& r, const QString& cutQual, bool isAnnotationDirect, bool isCutDirect) const {
+    bool hasDirect = false;
+    bool hasComplement = false;
+    int cutDirect = 0;
+    int cutComplement = 0;
+    if (!cutQual.isEmpty()) {
+        int complSplit = cutQual.indexOf('/');
+        if (-1 != complSplit) {
```
yalgaer: nit: do not use yoda style

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -231,6 +275,52 @@ void SequenceViewAnnotatedRenderer::drawBoundedText(QPainter& p, const QRect& r,
...
+            cutComplement = cutQual.mid(qMin(cutQual.length(), complSplit + 1)).toInt(&hasComplement);
+        } else {
+            cutDirect = cutQual.toInt(&hasDirect);
+            cutComplement = cutQual.toInt(&hasComplement);
+        }
+    }
+
+    int cutPosition = INCORRECT_CUT_POSITION;
+    if (isCutDirect && hasDirect) {
+        cutPosition = correctCutPos(isAnnotationDirect ? r.startPos + cutDirect : r.startPos + cutComplement);
```
yalgaer: correctCutPos is only used from your function and has wrong logic inside: it returns a valid value for invalid cuts: 
```
result = qBound((qint64)0, result, sequenceLength);
```
If we do not know how to render the cut correctly we should not render it at all.

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -231,6 +275,52 @@ void SequenceViewAnnotatedRenderer::drawBoundedText(QPainter& p, const QRect& r,
     p.drawText(r, Qt::TextSingleLine | Qt::AlignCenter, text.left(prefixLen));
 }
 
+int SequenceViewAnnotatedRenderer::getCutPosition(const U2Region& r, const QString& cutQual, bool isAnnotationDirect, bool isCutDirect) const {
+    bool hasDirect = false;
+    bool hasComplement = false;
+    int cutDirect = 0;
```
yalgaer: nit: add a readable suffix: `directCutIndex`, `complSplitIndex`

Current variable names look more like function names: `cutDirect`

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -231,6 +275,52 @@ void SequenceViewAnnotatedRenderer::drawBoundedText(QPainter& p, const QRect& r,
...
+    int cutComplement = 0;
+    if (!cutQual.isEmpty()) {
+        int complSplit = cutQual.indexOf('/');
+        if (-1 != complSplit) {
+            cutDirect = cutQual.left(complSplit).toInt(&hasDirect);
+            cutComplement = cutQual.mid(qMin(cutQual.length(), complSplit + 1)).toInt(&hasComplement);
```
yalgaer: Why do you need this qMin()? `complSplit` is always within the `cutQual` right?

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,61 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+                if (cutPosComplement != INCORRECT_CUT_POSITION) {
+                    auto annRect = getAnnotationRect(regionPair.second, canvasSize, visibleRange, selected, a, as);
+                    drawCutSite(p, aData, cutPosComplement, annRect, as->color, canvasSize, visibleRange, false);
+                }
+            }
+        } else if (location.size() == 1) {
+            const auto& reg = location.first();
+            const auto& qual = aData->findFirstQualifierValue(GBFeatureUtils::QUALIFIER_CUT);
+            QRect annRect;
+            bool isDirectStrand = aData->getStrand().isDirect();\
```
yalgaer: drop `\` character

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,61 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+            {
```
yalgaer: nit: this nesting is not needed - it does not affect visibility scope of any variables or resolves any conflicts
If you think this is a more readable version - use comments to mark blocks instead

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,61 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
-        } else if (isRestrictionSite && simple && displaySettings.displayCutSites) {
-            p.setPen(borderPen);
-            drawCutSite(p, aData, r, QRect(), as->color, canvasSize, visibleRange);
+        }
+    }
+    // draw restriction sites
+    if (isRestrictionSite && displaySettings.displayCutSites) {
+        p.setPen(borderPen);
+        const bool isCircular = ctx->getSequenceObject()->isCircular();
+        auto seqLength = ctx->getSequenceLength();
```
yalgaer: nit: here is example when auto makes code less readable. For a short type names int `int`/`int64` it is better to use the type, but not auto, like you do for `bool` in the same function

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,61 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+        if (isCircular && U1AnnotationUtils::isAnnotationContainsJunctionPoint(a, seqLength)) {
+            auto merged = U1AnnotationUtils::mergeAnnotatedRegionsAroundJunctionPoint(location, seqLength);
+            SAFE_POINT(merged.size() == 1, "Unexpected merged enzyme region", );
```
yalgaer: This should not stop other annotations rendering. Just ignore it

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
```diff
@@ -188,16 +187,61 @@ void SequenceViewAnnotatedRenderer::drawAnnotation(QPainter& p, const QSize& can
...
+
+            auto regionPair = merged.front();
+            // draw cuts separately,
+            // because they are located one on the 3' end of the direct sequence,
+            // and the other one on the 3' end of reverse-complementary sequence
+            const auto& qual =  aData->findFirstQualifierValue(GBFeatureUtils::QUALIFIER_CUT);
+            bool isDirectStrand = aData->getStrand().isDirect();
```
yalgaer: can be moved outside of if and reused in both merged/non-merged if cases.

yalgaer (commented):
The algorithm looks much cleaner and more readable now.
But some rendering problems still here.

One of the cut is not rendered here if I shift this enzyme:
image

like this:
image

The direct cut got wrong Y pos

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: I do not know, maybe some legacy, removed

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2View/src/ov_sequence/view_rendering/SequenceViewAnnotatedRenderer.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
> The direct cut got wrong Y pos

Fixed. This is some ancient bug, by the way.

yalgaer (commented):
Looks good. Will approve after tests.

BTW could you please do not overwrite the branch history in git: it does not allow to `git pull` locally without a full branch reset:

image

you see: all commits have incorrect time.

If this problem is caused by merging with master, please always use `git merge origin/master`

(We will anyway squash all internal commits on merging back to master into a single commit).

yalgaer (approved):

PR #1340, opened 2023-07-13
UGENE-7903 No breakpad dump on macOS

DmitriiSukhomlinov:


src/libs_3rdparty/breakpad/src/client/mac/handler/minidump_generator.cc
```diff
@@ -928,7 +928,12 @@ bool MinidumpGenerator::GetThreadState(thread_act_t target_thread,
         default:
             return false;
     }
-    return thread_get_state(target_thread, flavor, state, count) == KERN_SUCCESS;
+    // For some unknown reason,
```
DmitriiSukhomlinov, line 931: I found this solution [here](https://github.com/firebase/firebase-ios-sdk/blob/e97bc51566b3689b0e1f1249d9bfbb9a65580102/Crashlytics/Crashlytics/Components/FIRCLSProcess.c#L218). Probably, this behavior is just some Rosetta bug

src/ugenem/src/SendReportDialog.cpp
```diff
@@ -129,19 +129,11 @@ void ReportSender::parse(const QString& htmlReport, const QString& dumpUrl) {
         }
     }
 
-    QFile fp("/tmp/UGENEstacktrace.txt");
+    QFile fp(QDir::tempPath() + "/ugene_crashes/UGENEstacktrace.txt");
```
DmitriiSukhomlinov, line 132: Not related to the bug itself, but I found this problem too. This is the stacktrace, which was used before Google breakpad (it wasn't removed "just in case" and sometimes it really could show some extra info). Previously UGENE put it to `/tmp`, but this behavior was changed (see [here](https://github.com/ugeneunipro/ugene/blob/5f3edceb3e0c3dfe74d70b8a74d57f522508efce/src/corelibs/U2Private/src/crash_handler/CrashHandlerPrivateMac.cpp#L61)).

src/ugenem/src/SendReportDialog.cpp
```diff
@@ -129,19 +129,11 @@ void ReportSender::parse(const QString& htmlReport, const QString& dumpUrl) {
...
     if (fp.open(QIODevice::ReadOnly)) {
         QByteArray stacktrace = fp.readAll();
         report += "Stack trace:\n";
         report += stacktrace.data();
-        if (stacktrace.isEmpty()) {
```
DmitriiSukhomlinov, line 137: I think, this is some legacy - UGENE does not have file `UGENEerror` anywhere but here.

src/libs_3rdparty/breakpad/src/client/mac/handler/minidump_generator.cc
yalgaer: ok

src/ugenem/src/SendReportDialog.cpp
yalgaer: ok

src/ugenem/src/SendReportDialog.cpp
yalgaer: ok

yalgaer (approved):

PR #1342, opened 2023-07-14
Tune layout of enzymes dialog and fix trivial bugs

yalgaer:
Before:
image

After:
image

This is only minimal initial tuning. We can do better before the release

What was changed

1) No more buttons with the same name on the same dialog (like Select all)

2) Bigger spacing between logical blocks and smaller final dialog size.

3) Combo-box wiith checkboxes title shows only name of one  supplier or "All" or "N items". We anyway can't fit more than 2 but have 10+ in the list.

4) Fixed a regression with the initial state: UGENE with default settings didn't show "Not defined" enzymes at all (only 600 enzymes out of 16k were shown by default)

5) "None" is equal to "All" for the "Suppliers" filter: If this filter is not used (empty) it does not affect the table of enzymes (nothing is filtered). This way the table can never be empty (a useless state).

6) Fixed a bug when the dialog state was not saved at all when user pressed "OK" with no checked enzymes.

7) Cleaned up text of the labels to avoid logical overlaps (like "Selected enzymes" and "Selected enzymes info"), be more precise what is filtered ("Show/hide enzymes" (where?) -> "Enzyme table filter")


DmitriiSukhomlinov:
> Fixed a regression with the initial state: UGENE with default settings didn't show "Not defined" enzymes at all (only 600 enzymes out of 16k were shown by default)

Wait a bit - it wasn't a regression, I did it on purpose (see [UGENE-7888](https://ugene.dev/tracker/browse/UGENE-7888)). Users are interested in enzymes with defined suppliers first. SnapGene, for example, does not have enzymes with commercial sources. Geneious too, by the way.

> "None" is equal to "All" for the "Suppliers" filter: If this filter is not used (empty) it does not affect the table of enzymes (nothing is filtered). This way the table can never be empty (a useless state).

This could be quite unclear for user, I'd like to take the previous state back.

> Cleaned up text of the labels to avoid logical overlaps (like "Selected enzymes" and "Selected enzymes info"), be more precise what is filtered ("Show/hide enzymes" (where?) -> "Enzyme table filter")

Do not forget to update translations

yalgaer:
> > Fixed a regression with the initial state: UGENE with default settings didn't show "Not defined" enzymes at all (only 600 enzymes out of 16k were shown by default)
> 
> Wait a bit - it wasn't a regression, I did it on purpose (see [UGENE-7888](https://ugene.dev/tracker/browse/UGENE-7888)). Users are interested in enzymes with defined suppliers first. SnapGene, for example, does not have enzymes with commercial sources. Geneious too, by the way.

I see. I restored this behavior. In this case some tests should be fixed: test_3328 does not test what it is expected to test - the test is too fast with 600 enzymes (it needs 10k)


> > "None" is equal to "All" for the "Suppliers" filter: If this filter is not used (empty) it does not affect the table of enzymes (nothing is filtered). This way the table can never be empty (a useless state).
> 
> This could be quite unclear for user, I'd like to take the previous state back.

Have a second thought - agree. Returned the empty checkbox list behavior back.
 
> > Cleaned up text of the labels to avoid logical overlaps (like "Selected enzymes" and "Selected enzymes info"), be more precise what is filtered ("Show/hide enzymes" (where?) -> "Enzyme table filter")
> 
> Do not forget to update translations

Will do before merge after the tests are green

PR #1343, opened 2023-07-14
UGENE-7904 CMD crash handler version could be shown instead of GUI one

DmitriiSukhomlinov:
Read issue for details. Tested on ugene-cuda, ugene-macbook-1, ugene-macbook-2 and my Windows station

### DIFF
--- a/src/ugenem/src/main.cpp
+++ b/src/ugenem/src/main.cpp
@@ -42,11 +42,8 @@ QString loadReport(int argc, char* argv[]) {
 int main(int argc, char* argv[]) {
     bool useGui = true;
 #if defined(Q_OS_UNIX)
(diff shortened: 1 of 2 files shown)


src/ugenem/src/main.cpp
```diff
@@ -42,11 +42,8 @@ QString loadReport(int argc, char* argv[]) {
 int main(int argc, char* argv[]) {
     bool useGui = true;
 #if defined(Q_OS_UNIX)
-    useGui = (getenv("DISPLAY") != 0);
-    if (!useGui && argc == 1) {
```
DmitriiSukhomlinov, line 46: 20 lines below we have the same check

yalgaer (approved):

PR #1358, opened 2023-07-19
UGENE-7907 Enzyme is found if trailing/leading N's do not have symbols

DmitriiSukhomlinov:


src/plugins/enzymes/src/FindEnzymesAlgorithm.h
```diff
@@ -51,29 +51,54 @@ class FindEnzymesAlgorithm {
...
+        }
+        if (enzyme->cutComplement != ENZYME_CUT_UNKNOWN) {
+            if (enzyme->cutComplement < 0) {
+                trailingNsNumber = qMax(trailingNsNumber, qAbs(enzyme->cutComplement));
+            } else if (enzyme->cutComplement > enzymeSequence.size()) {
+                leadingNsNumber = qMax(leadingNsNumber, enzyme->cutComplement - enzymeSequence.size());
+            }
+        }
+        if (leadingNsNumber != 0) {
+            enzymeSequence.insert(0, QString(leadingNsNumber, EnzymeData::UNDEFINED_BASE));
```
yalgaer: nit: use QByteArray, not QString here.

src/plugins/enzymes/src/FindEnzymesAlgorithm.h
```diff
@@ -51,29 +51,54 @@ class FindEnzymesAlgorithm {
...
-    void run(const DNASequence& sequence, const U2Region& region, const SEnzymeData& enzyme, const char* pattern, U2Strand stand, FindEnzymesAlgListener* resultListener, TaskStateInfo& ti, int resultPosShift = 0) {
+    void run(const DNASequence& sequence, const U2Region& region, const SEnzymeData& enzyme, const QByteArray& byteArrayPattern, U2Strand stand, FindEnzymesAlgListener* resultListener, TaskStateInfo& ti, int resultPosShift = 0) {
+        const char* pattern = byteArrayPattern.constData();
         CompareFN fn(sequence.alphabet, enzyme->alphabet);
         const char* seq = sequence.constData();
         char unknownChar = sequence.alphabet->getDefaultSymbol();
-        int plen = enzyme->seq.length();
+        int plen = byteArrayPattern.length();
         for (int pos = region.startPos, endPos = region.endPos() - plen + 1; pos < endPos; pos++) {
             bool match = matchSite(seq + pos, pattern, plen, unknownChar, fn);
```
yalgaer, line 103: Looks like once you appended extra Ns this code won't work for circular sequences.
The simplest solution would be adding Ns only for non-circular sequences.

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_7001_8000.cpp
```diff
@@ -4832,5 +4832,21 @@ GUI_TEST_CLASS_DEFINITION(test_7896) {
...
+
+    FindEnzymesDialogFillerSettings settings;
+    settings.enzymes = QStringList{ "AbaSI" };
+    GTUtilsDialog::waitForDialog(new FindEnzymesDialogFiller(settings));
+    GTWidget::click(GTWidget::findWidget("Find restriction sites_widget"));
+    GTUtilsTaskTreeView::waitTaskFinished();
+
+    GTGlobals::FindOptions options;
+    options.failIfNotFound = false;
+    auto a = GTUtilsAnnotationsTreeView::findItem("AbaSI", nullptr, options);
```
yalgaer: This should be an XML test: it checks the algorithm only and XML tests run 100x faster.

Could you please also add a test when an enzyme Ns cross a circular sequence junction point (I this the current solution is not safe for circulars)

yalgaer (changes requested):

src/plugins/enzymes/src/FindEnzymesAlgorithm.h
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/FindEnzymesAlgorithm.h
DmitriiSukhomlinov: We have some problems with enzyme searching on circular molecules (see [UGENE-7908](https://ugene.dev/tracker/browse/UGENE-7908), reproducible on v. 47), so it would be better to fix it, and then figure out this scenario for circular molecules.

DmitriiSukhomlinov: It works for circular, I just tested it and everything is OK. I will add an XML test.

yalgaer: I see, there is a special handling of circular sequences below

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
```diff
@@ -1010,7 +1010,7 @@ GUI_TEST_CLASS_DEFINITION(test_6232_4) {
     // Expected state: the first fragment begins from 3 base, there are "AC" bases before it
     QList<U2Region> regions = GTUtilsAnnotationsTreeView::getAnnotatedRegionsOfGroup("STEP1_pFUS2_a2a_5 Fragment 1  (0, 1)");
     CHECK_SET_ERR(regions.size() == 1, QString("Unexpected number of fragments, expected: 1, current: %1").arg(regions.size()));
-    CHECK_SET_ERR(regions.first().startPos == 2, QString("Unexpected fragment startPos, expected: 2; current: %1").arg(regions.first().startPos));
+    CHECK_SET_ERR(regions.first().startPos == 0, QString("Unexpected fragment startPos, expected: 2; current: %1").arg(regions.first().startPos));
```
yalgaer: Please update the comment in the test

yalgaer (commented):
Looks good, will approve after tests

src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_6001_7000.cpp
DmitriiSukhomlinov: OK

DmitriiSukhomlinov:
@yalgaer please check again. Some unstable tests failed

yalgaer (approved):

PR #1372, opened 2023-07-26
UGENE-7894 Calculate sites number when file opens

DmitriiSukhomlinov:
Each enzyme's quantity calculates when you click on this enzyme in the dialog


yalgaer (approved):

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
```diff
@@ -76,6 +76,10 @@ class U2ALGORITHM_EXPORT EnzymeSettings {
 
 class U2ALGORITHM_EXPORT EnzymeData : public QSharedData {
 public:
+    bool operator==(const EnzymeData& other) const {
+        return id == other.id && seq == other.seq;
```
yalgaer: There should be no 2 enzymes with the same id registered in the system at the same moment

src/plugins/GUITestBase/src/tests/common_scenarios/sequence_view/GTTestsSequenceView.cpp
```diff
@@ -2572,6 +2572,39 @@ GUI_TEST_CLASS_DEFINITION(test_0084) {
...
+    public:
+        void run() override {
+            QWidget* dialog = GTWidget::getActiveModalWidget();
+            auto enzymesTree = GTWidget::findTreeWidget("tree", dialog);
+
+            QTreeWidgetItem* item = GTTreeWidget::findItem(enzymesTree, "AbaSI");
+            GTTreeWidget::click(item);
+            GTUtilsTaskTreeView::waitTaskFinished();
+            auto text = GTWidget::findTextBrowser("teSelectedEnzymeInfo", dialog)->toHtml();
+            CHECK_SET_ERR(text.contains("77655"), QString("Enzyme info shoul contain 77655, current text: %1").arg(text));
```
yalgaer: typo

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -611,7 +652,7 @@ FindEnzymesDialog::FindEnzymesDialog(ADVSequenceObjectContext* advSequenceContex
     initSettings();
 
     auto vl = new QVBoxLayout();
-    enzSel = new EnzymesSelectorWidget();
+    enzSel = new EnzymesSelectorWidget(advSequenceContext);
```
yalgaer, line 662: `advSequenceContext` may be destroyed before the widget. Use a shared pointer and do not show any counts if the object is destroyed (or switch to another object in the view)

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -273,6 +288,15 @@ void EnzymesSelectorWidget::setEnzymesList(const QList<SEnzymeData>& enzymes) {
         EnzymeGroupTreeItem* gi = dynamic_cast<EnzymeGroupTreeItem*>(item);
         if (ei != nullptr) {
             teSelectedEnzymeInfo->setHtml(ei->getEnzymeInfo());
+            if (!ei->hasNumberCalculationTask && advSequenceContext != nullptr) {
+                auto seqObj = advSequenceContext->getSequenceObject();
+                const auto& er = seqObj->getEntityRef();
+                U2Region reg(0, seqObj->getSequenceLength());
+                auto t = new FindSingleEnzymeTask(er, reg, ei->enzyme, nullptr, seqObj->isCircular());
```
yalgaer: Do not run the task for very large sequences. Info like: "the sequence has 782783723 hits" is a garbage info.
Limit the run to sequences ~10k

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -336,6 +360,23 @@ void EnzymesSelectorWidget::sl_filterTextChanged(const QString& filterText) {
     }
 }
 
+void EnzymesSelectorWidget::sl_findSingleEnzymeTaskStateChanged() {
+    auto t = qobject_cast<FindSingleEnzymeTask*>(sender());
+    SAFE_POINT(t != nullptr, L10N::nullPointerError("FindSingleEnzymeTask"), );
+    CHECK(t->getState() == Task::State_Finished, );
```
yalgaer, line 370: Check if the task failed too

yalgaer (changes requested):
Two issues in the patch:
1) We should not process large sequences and show meaningless info (spam info)
2) Sequence view widget should not be passed unprotected to other widgets: it may be destroyed in-flight.

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
DmitriiSukhomlinov: It could be and it legally happened - see "BarI". This enzyme has two possible cut pairs and we show them as separate enzymes.

src/plugins/GUITestBase/src/tests/common_scenarios/sequence_view/GTTestsSequenceView.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: If I use `QSharedPointer` here, `advSequenceContext` will be destroyed with `EnzymesSelectorWidget` - but it, of course, shouldn't be. We need to rewrite all `advSequenceContext` usage in UGENE - quite a large issue, don't you think?

Also, I can't get the scenario where `advSequenceContext` do not exist when you open a dialog, because the only possible way to open this dialog is from Sequence View.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: The dependency of restriction sites quantity from sequence length is quite implicit, I would even say, that there is almost no such dependency at all. For example, on human_T1 some enzymes occurs thousand times and some other - just several times or even zero, so this limitation could be very rough and it means, that we just forbid to find restriction enzymes in sequences larger than 10'000, which is bad (I didn't see such limitations is some other bioinformatic tools). 

It you want to add some limitation (but, as I think, this have no real sense, because all calculations passes very fast and sequences of any size) I can limit calculation of each enzyme, for example, 10'000 enzymes max. If there is more enzymes that 10'000 I will write "more than 10'000 sites". But I see it quite a bit redundant - we can calculate number, it is not complicated to us and let user decide if this number is meaningful for him or not

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
yalgaer: Lets do not inherit other system errors: ID is a unique identifier in UGENE - it's alone should be enough for the equality comparison.

I would suggest just drop the duplicate: AFAIU BarI duplicate is a plain rev-compl of the original one.

Are there any other enzymes like this? If yes and they are different we should reconsider our enzyme model: keep ids unique and introduce a new dimension for the variation like this.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: I do not ask to fix other bad places, only not to increase the tech-debt. 
The new code added must be written correctly.

Sequence view may have multiple sequences attached: each will have a separate sequence context.
Some of the sequences may be dynamically removed. In this case the context of the removed sequence is destroyed: so you can't cache it unprotected outside.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: I'm OK with either of the following solutions
1. Do not run the task for sequences > 10k
or
2. Run the task for any sequences but limit the time of the execution to a reasonable interval (like 1-2 seconds) and results memory usage to a reasonable value (<10mb).

You current solution has big flows: consider an nGb sequence and some trivial enzyme that has a single C letter: it may result to billions of results. This will be slow and finally crash UGENE with an out of memory error.

DmitriiSukhomlinov: I added a maximum results limit - 10'000. `FindSingleEnzymeTask` just stops if there are more sites than that number. Calculation of 10'000 sites never require more than 1-2 seconds

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
DmitriiSukhomlinov: Even if this is the only enzyme duplicated, we need to handle it, can't just drop it. I can split the only `id` field to `id` and `name`. Should I do it now? We've been living with it for a decade, maybe, at lease in separate issue?

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: OK

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
yalgaer: The problem I mentioned is in the new public method you added: if you do not introduce the wrong function I do not ask to fix the old issue as a part of this PR: we can make it a separate PR.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: I generated a random sequence of size 100Mb and already have a lot of long running tasks (each runs ~10 seconds on my computer):

image

What will be with a larger sequence size? The tasks have high pressure to memory, CPU, IO and are not canceled: so we have a multiplied pressure.

DmitriiSukhomlinov: I limited sequence length 200'000 bases max, I should be good, because on human"T1 everything works perfectly

src/corelibs/U2Algorithm/src/misc/EnzymeModel.h
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesPlugin.cpp
```diff
@@ -190,12 +190,14 @@ void EnzymesADVContext::initViewContext(GObjectViewController* view) {
 
 void EnzymesADVContext::sl_search() {
     auto action = qobject_cast<GObjectViewAction*>(sender());
-    assert(action != nullptr);
+    SAFE_POINT(action != nullptr, L10N::nullPointerError("GObjectViewAction"), );
```
yalgaer, line 193: nit: there is no need to localize safe points

src/plugins/enzymes/src/EnzymesPlugin.cpp
```diff
@@ -190,12 +190,14 @@ void EnzymesADVContext::initViewContext(GObjectViewController* view) {
...
+
     auto av = qobject_cast<AnnotatedDNAView*>(action->getObjectView());
-    assert(av != nullptr);
+    SAFE_POINT(av != nullptr, L10N::nullPointerError("AnnotatedDNAView"), );
+
+    auto seqCtx = QSharedPointer<ADVSequenceObjectContext>(av->getActiveSequenceContext(), [](ADVSequenceObjectContext*) {});
```
yalgaer: use a regular `QPointer seqCtx(av->get...)`
You do not really need QSharedPointer here

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -273,6 +288,18 @@ void EnzymesSelectorWidget::setEnzymesList(const QList<SEnzymeData>& enzymes) {
         EnzymeGroupTreeItem* gi = dynamic_cast<EnzymeGroupTreeItem*>(item);
         if (ei != nullptr) {
             teSelectedEnzymeInfo->setHtml(ei->getEnzymeInfo());
+            static constexpr int MAXIMU_CALCULATE_NUMBER_LENGTH = 200'000;
```
yalgaer: nit: typo
NUMBER_LENGTH also sounds bad. Rename to something like MAX_CHECKED_SEQUENCE_LENGTH

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -336,6 +363,27 @@ void EnzymesSelectorWidget::sl_filterTextChanged(const QString& filterText) {
     }
 }
 
+void EnzymesSelectorWidget::sl_findSingleEnzymeTaskStateChanged() {
+    auto t = qobject_cast<FindSingleEnzymeTask*>(sender());
+    SAFE_POINT(t != nullptr, L10N::nullPointerError("FindSingleEnzymeTask"), );
```
yalgaer, line 367: No need to localize safe points: their errors are not visible to users.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -336,6 +363,27 @@ void EnzymesSelectorWidget::sl_filterTextChanged(const QString& filterText) {
...
+
+    CHECK(!t->hasError(), );
+    CHECK(t->getState() == Task::State_Finished, );
+
+    auto taskEnzyme = t->getEnzyme();
+    auto taskItem = getEnzymeTreeItemByEnzymeData(taskEnzyme);
+    CHECK(taskItem != nullptr, );
+
+    int size = t->getResults().size();
+    bool maxResultsFound = t->isCanceled() && size == EnzymeTreeItem::MAXIMUM_ENZYMES_NUMBER;
```
yalgaer: If the task is canceled you can't trust the result: I can cancel it via Task manager.
The task should use max-enzymes limit and finish with no errors & not canceled state.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -273,6 +288,18 @@ void EnzymesSelectorWidget::setEnzymesList(const QList<SEnzymeData>& enzymes) {
...
+            if (!ei->hasNumberCalculationTask &&
+                !advSequenceContext.isNull() &&
+                advSequenceContext->getSequenceLength() < MAXIMU_CALCULATE_NUMBER_LENGTH) {
+                auto seqObj = advSequenceContext->getSequenceObject();
+                const auto& er = seqObj->getEntityRef();
+                U2Region reg(0, seqObj->getSequenceLength());
+                auto t = new FindSingleEnzymeTask(er, reg, ei->enzyme, nullptr, seqObj->isCircular(), EnzymeTreeItem::MAXIMUM_ENZYMES_NUMBER, false);
+                AppContext::getTaskScheduler()->registerTopLevelTask(t);
+                connect(t, &FindSingleEnzymeTask::si_stateChanged, this, &EnzymesSelectorWidget::sl_findSingleEnzymeTaskStateChanged);
+                ei->hasNumberCalculationTask = true;
```
yalgaer, line 300: If I cancel the task via Task manager it will never recalculate the result again with the current implementation.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
```diff
@@ -898,6 +953,13 @@ QString EnzymeTreeItem::getEnzymeInfo() const {
     QString result;
     result += QString("<a href=\"http://rebase.neb.com/rebase/enz/%1.html\">%1</a>")
                   .arg(text(Column::Id));
+    if (enzymesNumber != INCORRECT_ENZYMES_NUMBER) {
+        if (enzymesNumber > MAXIMUM_ENZYMES_NUMBER) {
+            result += tr(" (>10000 sites)");
```
yalgaer: I do not see this text in russian.ts. Looks like tr() does not work here as expected.

yalgaer: The constant should be a part of the text: we can change it later.

yalgaer (changes requested):

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: You cannot cancel task via task manager because you have a modal "Restriction enzymes" dialog opened.

src/plugins/enzymes/src/EnzymesPlugin.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/EnzymesPlugin.cpp
DmitriiSukhomlinov: Don't get the point. I just replaced `assert` with `SAFE_POINT`, no localizations done

src/plugins/enzymes/src/FindEnzymesDialog.cpp

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: But I do not do any localization, do I?

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: You cannot cancel task because of the modal dialog and it could be not easy to rewrite `FindSingleEnzymeTask` as you want. Not sure if it's worth it

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: Just forgot to update translation file, fixed

src/plugins/enzymes/src/EnzymesPlugin.cpp
yalgaer: > Don't get the point. I just replaced `assert` with `SAFE_POINT`, no localizations done

What does L10N mean in this case?

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: L10N means L[ocalizatio]N

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: The important part here is that you can't trust result of a canceled task. The cancellation may happen by different reasons.

src/plugins/enzymes/src/FindEnzymesDialog.cpp
yalgaer: There are many reasons why cancellation may happen and the dialog should not guess any of them: it should work in its own scope correctly.

src/plugins/enzymes/src/EnzymesPlugin.cpp
DmitriiSukhomlinov: I didn't even think about it, I just use `L10N::nullPointerError` all the times when I have `SAFE_POINT`'s like this, you have had no questions about it in my previous reviews

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: Same

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: Same

src/plugins/enzymes/src/FindEnzymesDialog.cpp
DmitriiSukhomlinov: OK

src/plugins/enzymes/src/FindEnzymesDialog.cpp

yalgaer:
Looks good will approve after tests

DmitriiSukhomlinov:
@yalgaer all test passed, Linux XML and Windows GUI have some global problems (on all branches)

yalgaer (approved):

PR #1420, opened 2023-09-05
UGENE-7908 Enzyme is found, but shouldn't be

DmitriiSukhomlinov:
`SequenceDbiWalkerTask` appends a beginning of sequence as a postfix to the whole sequence if `SequenceDbiWalkerConfig::walkCircular` is set to true - this simulates that our sequence is circular. From the over hand, we do the same thing [here](https://github.com/ugeneunipro/ugene/blob/d91b87af14a470b27c5af906f6000d3b8c803933/src/plugins/enzymes/src/FindEnzymesAlgorithm.h#L126) when we are already calculating enzymes on junction point. I decided to remove this handling from `SequenceDbiWalkerTask` because when you are looking to `FindEnzymesAlgorithm::run(...)` it is not clear where this postfix already come from, you need to investigate it for a while.

### DIFF
--- a/src/plugins/enzymes/src/FindEnzymesTask.cpp
+++ b/src/plugins/enzymes/src/FindEnzymesTask.cpp
@@ -257,9 +257,10 @@ void FindSingleEnzymeTask::prepare() {
     SequenceDbiWalkerConfig sequenceWalkerConfig;
     sequenceWalkerConfig.seqRef = sequenceObjectRef;
     sequenceWalkerConfig.range = region;
-    sequenceWalkerConfig.chunkSize = qMax(enzyme->seq.size(), chunkSize);
+    int enzymeFullLength = enzyme->getFullLength();
+    sequenceWalkerConfig.chunkSize = qMax(enzymeFullLength, chunkSize);
     sequenceWalkerConfig.lastChunkExtraLen = sequenceWalkerConfig.chunkSize / 2;
-    sequenceWalkerConfig.overlapSize = enzyme->seq.size() - 1;
+    sequenceWalkerConfig.overlapSize = enzymeFullLength - 1;
     sequenceWalkerConfig.walkCircular = isCircular;
     sequenceWalkerConfig.walkCircularDistance = sequenceWalkerConfig.overlapSize;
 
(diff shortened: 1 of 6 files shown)


DmitriiSukhomlinov:
`primers-pairs-checker/test_0001.xml` fails because you changed `_common_data/cmdline/primers/patterns/check_primers.html`

yalgaer:
FindEnzymesTask works with SequenceWalker which is desiged to provide a common optimized and tested way to deal with circular and large sequences.

What is the reason for FindAlgorithmTask to repeat this functionality (and to it not an optimal way by cloning a sequence in memory)?

The current solution may deliver correct result, but it may be wrong by design if there are no real reasons what prevent us using SequenceWalker here as expected. Otherwise the solution that removes circular "if ()" block from the FindAlgorithm would be the correct one.

DmitriiSukhomlinov:
> FindEnzymesTask works with SequenceWalker which is desiged to provide a common optimized and tested way to deal with circular and large sequences.
> 
> What is the reason for FindAlgorithmTask to repeat this functionality (and to it not an optimal way by cloning a sequence in memory)?
> 
> The current solution may deliver correct result, but it may be wrong by design if there are no real reasons what prevent us using SequenceWalker here as expected. Otherwise the solution that removes circular "if ()" block from the FindAlgorithm would be the correct one.

OK, I took everything in SequenceWalker bach and fixed the algorithm itselt. Actually, I just removed `if (circular)` part, because, in this case, it is reallu redundant.

src/plugins/enzymes/src/FindEnzymesTask.cpp
```diff
@@ -257,9 +257,10 @@ void FindSingleEnzymeTask::prepare() {
     SequenceDbiWalkerConfig sequenceWalkerConfig;
     sequenceWalkerConfig.seqRef = sequenceObjectRef;
     sequenceWalkerConfig.range = region;
-    sequenceWalkerConfig.chunkSize = qMax(enzyme->seq.size(), chunkSize);
+    int enzymeFullLength = enzyme->getFullLength();
```
DmitriiSukhomlinov, line 260: Enzyme length is not the enzyme only, but leading and trailing N bases. We had no problems with it previously, because we had separate handling of the enzyme in `FindEnzymesAlgorithm.h`, but now we need to add a propper number of bases to the end of sequence.

yalgaer:
Looks good. Will approve after tests

yalgaer (approved):

PR #1432, opened 2023-10-04
UGENE-7924 Incorrect timeout for Query NCBI BLAST database...

DmitriiSukhomlinov:
Also, I've got a suspicion, that test **remote-request/test_0002.xml** started to fail because of this line. But I'm not sure, it could be something else

### DIFF
--- a/src/plugins/remote_blast/src/HttpRequestBLAST.cpp
+++ b/src/plugins/remote_blast/src/HttpRequestBLAST.cpp
@@ -111,7 +111,6 @@ void HttpRequestBLAST::sendRequest(const QString& params, const QString& query) 
     SAFE_POINT(rTask != nullptr, "Not a RemoteBlastHttpRequestTask", );
     int checkTimeSeconds = qMax(5, estimatedRunTimeSeconds / 10);
     rTask->resetProgress();
-    rTask->setTimeOut(10 * estimatedRunTimeSeconds);
     int iteration = 0;
     int startTime = QDateTime::currentSecsSinceEpoch();
     do {


yalgaer:
> Also, I've got a suspicion, that test **remote-request/test_0002.xml** started to fail because of this line. But I'm not sure, it could be something else

What will happen if no explicit timeout is provided? Won't we have a situation when the task will run forever?

DmitriiSukhomlinov:
> > Also, I've got a suspicion, that test **remote-request/test_0002.xml** started to fail because of this line. But I'm not sure, it could be something else
> 
> What will happen if no explicit timeout is provided? Won't we have a situation when the task will run forever?

Timeout has always been provided for this task. It uses not `TaskScheduler` features, but some inner code:

1. We set the timeout value in the dialog (in minuits):
image
2. Then we set timer, which triggers, when the time is out (see [this code](https://github.com/ugeneunipro/ugene/blob/05e1e09a5c45fb638716721a9b2d3fb2a09411af/src/plugins/remote_blast/src/RemoteBLASTTask.cpp#L172C6-L172C6) and 4 lines below)
3. When time is out, the `timeout` flag switches, `rTask->isTimeOut()` returns `false` and the execution runs out of the `do {} wile` loop (see [here](https://github.com/ugeneunipro/ugene/blob/05e1e09a5c45fb638716721a9b2d3fb2a09411af/src/plugins/remote_blast/src/HttpRequestBLAST.cpp#L134))

yalgaer:
Looks good. Approve after GUI tests

DmitriiSukhomlinov:
All tests passed except `GUITest_common_scenarios_primer3_test_0025` which fails because I changed test data for https://github.com/ugeneunipro/ugene/pull/1433

yalgaer (approved):

PR #1433, opened 2023-10-04
UGENE-7940 Fix Primer3 Sequence Quality and add some tests

DmitriiSukhomlinov:
Also, several bugs has been fixed:

- Quality hasn't been saved in the Primer3 dialog, because `U2SavableWidget` wasn't able to work with `QPlainTextEdit`
- Some fields of the dialog had incorrect min/max ranges, so some good values couldn't beed typed (e.g. it was no possibility to type negative annealing temp)


src/plugins/GUITestBase/src/tests/regression_scenarios/GTTestsRegressionScenarios_1001_2000.cpp
```diff
@@ -5970,9 +5970,6 @@ GUI_TEST_CLASS_DEFINITION(test_1654) {
     GTUtilsOptionPanelSequenceView::closeTab(GTUtilsOptionPanelSequenceView::Search);
 
     GTUtilsOptionPanelSequenceView::openTab(GTUtilsOptionPanelSequenceView::Search);
-    GTWidget::checkEnabled("getAnnotationsPushButton", false);
```
DmitriiSukhomlinov, line 5973: Раньше на вкладке "Поиск паттерна" паттерн сохранялся при закрытии-открытии вкладки. В версии 46 произошла регрессия и паттерн начал сбрасываться. Сейчас я восстановил это поведение, добавив `QPlainTextEdit` в `U2SavableWidget`. Данный тест был переписан относительно недавно ([тут](https://github.com/ugeneunipro/ugene/pull/1417/files)) таким образом, что он начал поддерживать некорректное (появившееся в версии 46) поведение. Сейчас я исправляю его, он становится аналогичным сценарию, написанному в оригинальной задаче 1654

data/primer3/presets/Default.txt
```diff
@@ -43,7 +43,7 @@ SEQUENCE_FORCE_LEFT_START=-1000000
 SEQUENCE_FORCE_RIGHT_END=-1000000
 SEQUENCE_FORCE_RIGHT_START=-1000000
 SEQUENCE_START_CODON_POSITION=-2000000
-PRIMER_ANNEALING_TEMP=0
+PRIMER_ANNEALING_TEMP=-10
```
yalgaer, line 46: Why do we have this as the default value? What does -10 means?

src/plugins_3rdparty/primer3/src/Primer3Dialog.cpp
```diff
@@ -657,9 +658,12 @@ bool Primer3Dialog::doDataExchange() {
             }
             qualityList.append(value);
         }
-        if (!qualityList.isEmpty() && (qualityList.size() != (rs->getRegion().length))) { // todo add check on wrong region
+        int qualityLength = qualityList.size();
+        int sequenceLength = context->getSequenceLength();
```
yalgaer, line 662: nit: this should be qint64 or the whole dialog should be protected to never be called for sequences with length > MAX_INT

data/primer3/presets/Default.txt
DmitriiSukhomlinov: This is the default value used by the original Primer3 tool (see [here](https://github.com/primer3-org/primer3/blob/7f9f17d6012f404e83bbf0f931a4d06eb4af465b/src/libprimer3.cc#L702)). We had 0 previusly, because `QDoubleSpinBox`, responsible for this value, had 0 as a minimal limit. When I created this ".txt" file I just saved parameters from the Primer3 dialog and didn't notice this mistake. In this commit I fixed limits for some spinboxed (this one was included) and changed the default value as it set in Primer3.

If you are asking about the biological sence of this variable, I'm not sure that I'm the one who may give you a good answer, I understand it very approximatelly. But, as far as Primer3 developers set this parameter to -10 as default, I suppose, it makes sense.

src/plugins_3rdparty/primer3/src/Primer3Dialog.cpp
DmitriiSukhomlinov: And it is [already protected](https://github.com/ugeneunipro/ugene/blob/05e1e09a5c45fb638716721a9b2d3fb2a09411af/src/plugins_3rdparty/primer3/src/Primer3Dialog.cpp#L740C61-L740C61)

yalgaer:
Замечаний нет, approve после тестов

DmitriiSukhomlinov:
Тесты прошли кроме каких-то нестабильных

yalgaer (approved):

PR #1678, opened 2024-09-17
UGENE-8141 Workflow "Annotate with UQL Element" does not save annotaion as expected

rasputinkirill:


yalgaer (commented):
Замечаний нет, апрув после тестов

yalgaer (approved):

PR #1746, opened 2025-05-28
UGENE-8097

rasputinkirill:
Comma-separated name filter in the Restriction Enzymes dialog


yalgaer (commented):
Замечаний нет, ждем тестов

yalgaer (approved):

